PR #82, opened 2021-12-06
Update supported Python versions

thisac:
**Context:**
Only Python 3.8 and 3.9 are supported.

* Python 3.7 is removed from supported Python versions in `setup.py`.
* The Python badge in the README is updated to show `python 3.8 | 3.9` instead of `python 3.8+`.

**Benefits:**
The correct Python versions are documented.

**Possible Drawbacks:**
None

**Related GitHub Issues:**
None


josh146:
Out of curiosity, how come 3.7 is not supported?

PR #98, opened 2021-12-16
Pin `requirements-dev`

sduquemesa:
**Context:**
Both Tensorflow and pylint require different versions of `typing-extensions` leading to conflict when installing from unpinned `requirements-dev.txt`.

This PR pins package versions on `requirements-dev.txt` and solves the dependency conflict.

**Benefits:**
Smoot installation of packages for development.

**Possible Drawbacks:**
Will use `pylint==2.10.0` which is not the latest release.

### DIFF
--- a/requirements-dev.txt
+++ b/requirements-dev.txt
@@ -1,4 +1,4 @@
-pytest
-hypothesis
-pytest-cov
-pylint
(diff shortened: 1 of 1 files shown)


ziofil (approved):
good idea

PR #101, opened 2022-01-06
Improve pylint score of `math` module

sduquemesa:
**Context:**
Mr Mustard uses `pylint` for code quality checking.

- This PR refactors the module `math` so that `pylint mrmustard/math` code rating goes from `1.45/10` to `9.09/10`. 
- A couple of new docstrings are included.

**Benefits:**
Pylint score of `math` module is improved.

### DIFF
--- a/mrmustard/math/math_interface.py
+++ b/mrmustard/math/math_interface.py
@@ -12,15 +12,18 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-from abc import ABC, abstractmethod, abstractproperty
-from mrmustard.types import *
-import numpy as np
+"""This module contains the :class:`Math` interface that every backend has to implement."""
+
+from abc import ABC, abstractmethod
 from functools import lru_cache
+from itertools import product
+import numpy as np
 from scipy.special import binom
 from scipy.stats import unitary_group
-from itertools import product
 
+from mrmustard.types import List, Tensor, Matrix, Scalar, Vector, Sequence, Tuple, Optional, Dict, Trainable, Callable, Any
 
+# pylint: disable=too-many-public-methods
 class MathInterface(ABC):
     r"""The interface that all backends must implement."""
     _euclidean_opt: type = None  # NOTE this is an object that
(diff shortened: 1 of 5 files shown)


mrmustard/math/tensorflow.py
```diff
@@ -336,23 +349,23 @@ def grad(dLdpoly):
 
         return poly, grad
 
-    def eigvals(self, tensor: tf.Tensor) -> Tensor:
+    def eigvals(tensor: tf.Tensor) -> Tensor:
```
heltluke, line 352: Not sure you can just remove `self` given that this is a method of the `TFMath` class. Maybe what you want to do is decorate it with `@staticmethod`?

mrmustard/math/math_interface.py
```diff
@@ -840,14 +857,17 @@ def zeros_like(self, array: Tensor) -> Tensor:
 
     @property
     def euclidean_opt(self):
+        r"""Returns the configured Euclidiand optimizer."""
```
ziofil, line 860:
```suggestion
        r"""Returns the configured Euclidean optimizer."""
```

mrmustard/math/tensorflow.py
```diff
@@ -135,8 +148,8 @@ def gather(self, array: tf.Tensor, indices: tf.Tensor, axis: int = None) -> tf.T
     def hash_tensor(self, tensor: tf.Tensor) -> int:
         try:
             REF = tensor.ref()
-        except AttributeError:
-            raise TypeError(f"Cannot hash tensor")
+        except AttributeError as e:
+            raise TypeError("Cannot hash tensor") from e
```
ziofil, line 152: super

mrmustard/math/tensorflow.py
ziofil: hmm I think Zhi wrote these. Looks like he forgot the self argument in a few of them?

ziofil (approved):
perhaps we need those `self` arguments?

mrmustard/math/tensorflow.py
heltluke: Not sure it was "forgotten" but rather "removed" (perhaps on the advice of `pylint` because the method doesn't actually use it?). Still, I believe the tests are currently failing because of this change, and would recommend either putting it back or making these `@staticmethod`s.

sduquemesa: Thanks @heltluke! I indeed removed `self` following pylint suggestions.

PR #103, opened 2022-01-07
Make `TFMath` adhere to signatures on `MathInterface`

sduquemesa:
**Context:**
All backends must implement and adhere to signatures on `MathInterface`

Modifies the signature of methods on `TFmath` and some of `MathInterface` abstract methods so that both the tensorflow and torch backend adhere to `MathInterface`.

**Benefits:**
Consistent implementation of backends

### DIFF
--- a/mrmustard/math/math_interface.py
+++ b/mrmustard/math/math_interface.py
@@ -83,13 +83,14 @@ class MathInterface(ABC):
         ...
 
     @abstractmethod
-    def arange(self, start: int, limit: int = None, delta: int = 1) -> Tensor:
+    def arange(self, start: int, limit: int = None, delta: int = 1, dtype: Any = None) -> Tensor:
         r"""Returns an array of evenly spaced values within a given interval.
 
         Args:
             start (int): start of the interval
             limit (int): end of the interval
             delta (int): step size
+            dtype (type): dtype of the returned array
 
         Returns:
             array: array of evenly spaced values
@@ -225,8 +226,10 @@ class MathInterface(ABC):
         self,
         array: Tensor,
         filters: Tensor,
-        padding="VALID",
-        data_format="NWC",
+        strides: Optional[List[int]] = None,
+        padding = "VALID",
+        data_format =" NWC",
+        dilations: Optional[List[int]] = None,
     ) -> Tensor:  # TODO: remove strides and data_format?
         r"""Performs a convolution on array with filters.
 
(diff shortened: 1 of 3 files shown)


ziofil (changes requested):
until there is a reason to include extra arguments in our interface, probably better to remove them (I'm referring to the convolution). thank you for the rest of the changes!

mrmustard/math/math_interface.py
```diff
@@ -225,8 +226,10 @@ def convolution(
         self,
         array: Tensor,
         filters: Tensor,
+        strides: Optional[List[int]] = None,
```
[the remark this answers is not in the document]
sduquemesa: I realized both of the current backends (TF and Torch) use such arguments so I though it'd be better to add them to the interface. However, I agree with you: if those arguments are of no use then better no to expose them.

mrmustard/math/tensorflow.py
```diff
@@ -292,9 +294,8 @@ def zeros_like(self, array: tf.Tensor) -> tf.Tensor:
     # Special functions
     # ~~~~~~~~~~~~~~~~~
 
-    def DefaultEuclideanOptimizer(
-        self,
-    ) -> tf.keras.optimizers.Optimizer:  # TODO: a wrapper class is better?
+    @staticmethod
+    def DefaultEuclideanOptimizer() -> tf.keras.optimizers.Optimizer:  # TODO: a wrapper class is better?
```
thisac, line 298: (very minor) just to avoid >100 lines. 
```suggestion
    # TODO: is a wrapper class better?
    @staticmethod
    def DefaultEuclideanOptimizer() -> tf.keras.optimizers.Optimizer:
```

mrmustard/math/tensorflow.py
```diff
@@ -374,6 +375,7 @@ def eigh(tensor: tf.Tensor) -> Tensor:
         """Returns the eigenvalues and eigenvectors of a matrix."""
         return tf.linalg.eigh(tensor)
 
+    # pylint: disable=arguments-differ
```
thisac, line 378: (minor) would it perhaps be better to add the `rtol` and `atol` arguments to `MathInterface.sqrtm` (and also `TFMath.sqrtm`)?

thisac (approved):
Looks good to me! 👍

mrmustard/math/tensorflow.py
sduquemesa: I guess that, as Filippo mentioned before, no need to clutter the interface with extra args that not all backends will implement. For example, torch backend does not use the `rtol` and `atol` args.

sduquemesa:
If implementation and interface arguments differ, I wonder if having the docstrings in the interface is a good idea since every backend will have different arguments on the implementation -- and they should be properly documented.

thisac:
> If implementation and interface arguments differ, I wonder if having the docstrings in the interface is a good idea since every backend will have different arguments on the implementation -- and they should be properly documented.

Good point. If they differ significantly, I'd argue that they should be inherited from the same abstract base class in the first place. But I agree, that it's important to document them thoroughly.

PR #105, opened 2022-01-12
Stop optimizer execution on `KeyboardInterrupt`

### LINKED ISSUE #104: Trying to stop a looped optimization does not work as expected

### Expected behavior

When running a for loop that is iterating over the Optimizer (e.g. to sweep through a parameter and find an optimal circuit), if I want to interrupt the program (I tried in Spyder or in Jupyter notebook), the whole program should stop. That is, the loop should stop, not just the current iteration of the loop.

### Actual behavior

Only the optimization in the current iteration of the loop stops, and the next iteration of the loop begins.

### Reproduces how often

100%

### System information

```shell
Mr Mustard: a differentiable bridge between phase space and Fock space.
Copyright 2018-2021 Xanadu Quantum Technologies Inc.

Python version:            3.8.12
Platform info:             Windows-10-10.0.19042-SP0
Installation path:         C:\Users\Xanadu\.conda\envs\mmenv\lib\site-packages\mrmustard
Mr Mustard version:        0.1.1
Numpy version:             1.21.4
Numba version:             0.53.1
Scipy version:             1.7.3
The Walrus version:        0.17.0
TensorFlow version:        2.7.0
Torch version:             None
```

### Source code

```shell
from mrmustard.lab import Dgate, Attenuator, Vacuum, Coherent
from mrmustard.utils.training import Optimizer

D = Dgate(x = 0.1, y = -0.5, x_trainable=True, y_trainable=True)
L = Attenuator(transmissivity=0.5)

def cost_fn_eucl():
    state_out = Vacuum(1) >> D >> L
    return 1 - fidelity(state_out, Coherent(0.1, 0.5))

# The for loop is the important part. The exact circuit is irrelevant.
for _ in range(10):
    opt = Optimizer(symplectic_lr=0.1, euclidean_lr=0.01)
    opt.minimize(cost_fn_eucl, by_optimizing=[D])
```

### Tracebacks

_No response_

### Additional information

```shell
You need to try to stop the code while the for loop is running.
```

sduquemesa:
**Context:**
When running a `for` loop that is iterating over the `Optimizer` sending a `KeyboardInterrupt` signal (e. g., `ctrl/cmd + c`) does not stop the program execution but only the current iteration of the loop.

The execution will gracefully stop, meaning no traceback will be printed to the user. To do this a helper class `StopExecution` inheriting from `Exception` is implemented overriding the `_render_traceback_` method in order not to print a traceback.

**Benefits:**
`ctrl/cmd + c` while on a optimization loop stops the execution of the program.

**Possible Drawbacks:**
This will not restore the parameters to the previous values, i. e. the values before any optimization step was performed. 

**Related GitHub Issues:** 
Resolves #104

### DIFF
--- a/mrmustard/utils/training.py
+++ b/mrmustard/utils/training.py
@@ -12,10 +12,12 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+from logging import Logger
 import numpy as np
 from scipy.linalg import expm
 from mrmustard.types import *
 from mrmustard.utils import graphics
+from mrmustard.logger import create_logger
 from mrmustard.math import Math
 
 math = Math()
@@ -38,6 +40,7 @@ class Optimizer:
         self.orthogonal_lr: float = orthogonal_lr
         self.euclidean_lr: float = euclidean_lr
         self.opt_history: List[float] = [0]
+        self.log = create_logger(__name__)
 
     def minimize(
         self, cost_fn: Callable, by_optimizing: Sequence[Trainable], max_steps: int = 1000
@@ -73,9 +76,10 @@ class Optimizer:
                     update_euclidean(params["euclidean"], grads["euclidean"], self.euclidean_lr)
                     self.opt_history.append(cost)
                     bar.step(math.asnumpy(cost))
-        except KeyboardInterrupt as e:  # graceful exit
-            print("KeyboardInterrupt: Stopping execution...")
-            raise StopExecution from e
+        except KeyboardInterrupt:  # graceful exit
+            raise self.OptimizerInterruptedError(
+                "Stopping optimizer execution...", self.log
+            ) from None
 
     def should_stop(self, max_steps: int) -> bool:
         r"""Returns ``True`` if the optimization should stop (either because the loss is stable or because the maximum number of steps is reached)."""
@@ -86,16 +90,19 @@ class Optimizer:
                 sum(abs(self.opt_history[-i - 1] - self.opt_history[-i]) for i in range(1, 20))
                 < 1e-6
             ):
-                print("Loss looks stable, stopping here.")
+                self.log.info("Loss looks stable, stopping here.")
                 return True
         return False
 
+    class OptimizerInterruptedError(Exception):
+        """A helper class to quietly stop execution without printing a traceback."""
 
-class StopExecution(Exception):
-    """A helper class to quietly stop execution without printing a traceback."""
+        def __init__(self, message: str, logger: Logger) -> None:
+            super().__init__()
+            logger.info(message)
 
-    def _render_traceback_(self):
-        pass
+        def _render_traceback_(self):
+            pass
 
 
 # ~~~~~~~~~~~~~~~~~
--- a/mrmustard/utils/training.py
+++ b/mrmustard/utils/training.py
@@ -77,9 +77,8 @@ class Optimizer:
                     self.opt_history.append(cost)
                     bar.step(math.asnumpy(cost))
         except KeyboardInterrupt:  # graceful exit
-            raise self.OptimizerInterruptedError(
-                "Stopping optimizer execution...", self.log
-            ) from None
+            logger.info("Optimizer execution halted due to keyboard interruption.")
+            raise self.OptimizerInterruptedError() from None
 
     def should_stop(self, max_steps: int) -> bool:
         r"""Returns ``True`` if the optimization should stop (either because the loss is stable or because the maximum number of steps is reached)."""
@@ -97,9 +96,8 @@ class Optimizer:
     class OptimizerInterruptedError(Exception):
         """A helper class to quietly stop execution without printing a traceback."""
 
-        def __init__(self, message: str, logger: Logger) -> None:
+        def __init__(self) -> None:
             super().__init__()
-            logger.info(message)
 
         def _render_traceback_(self):
             pass
--- a/mrmustard/utils/training.py
+++ b/mrmustard/utils/training.py
@@ -73,8 +73,9 @@ class Optimizer:
                     update_euclidean(params["euclidean"], grads["euclidean"], self.euclidean_lr)
                     self.opt_history.append(cost)
                     bar.step(math.asnumpy(cost))
-        except KeyboardInterrupt:  # graceful exit
-            return
+        except KeyboardInterrupt as e:  # graceful exit
+            print("KeyboardInterrupt: Stopping execution...")
+            raise StopExecution from e
 
     def should_stop(self, max_steps: int) -> bool:
         r"""Returns ``True`` if the optimization should stop (either because the loss is stable or because the maximum number of steps is reached)."""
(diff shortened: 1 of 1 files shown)


mrmustard/utils/training.py
```diff
@@ -73,8 +73,9 @@ def minimize(
                     update_euclidean(params["euclidean"], grads["euclidean"], self.euclidean_lr)
                     self.opt_history.append(cost)
                     bar.step(math.asnumpy(cost))
-        except KeyboardInterrupt:  # graceful exit
-            return
+        except KeyboardInterrupt as e:  # graceful exit
+            print("KeyboardInterrupt: Stopping execution...")
+            raise StopExecution from e
```
thisac, line 78: I'm not sure if this is the best way to do this. It's not that "graceful" of an exit, since another error is simply raised from the `KeyboradInterrupt` error. I get a:
```
~/miniforge3/envs/mm/lib/python3.9/site-packages/IPython/core/ultratb.py in stb2text(self, stb)
   1281     def stb2text(self, stb):
   1282         """Convert a structured traceback (a list) to a string."""
-> 1283         return self.tb_join_char.join(stb)
   1284 
   1285 

TypeError: can only join an iterable
```
with the
```
~/Repos/mrmustard/mrmustard/utils/training.py in minimize(self, cost_fn, by_optimizing, max_steps)
     77             print("KeyboardInterrupt: Stopping execution...")
---> 78             raise StopExecution from e
     79 

StopExecution: 

During handling of the above exception, another exception occurred:
```
much higher up, and the `"KeyboardInterrupt: Stopping execution..."` basically out-of-sight.

thisac: Some other possible options:
* Simply remove the try-except block and thus raise the first `KeyboardInterrupt` error.
* Use `raise SystemExit("KeyboardInterrupt: Stopping execution...")` (or `sys.exit("KeyboardInterrupt: Stopping execution...")`), although I'm not sure about this one, since it'll basically work like the current solution.
* Have `minimize` return a status (e.g., a "finished" bool, or a status message, e.g., "success", "failed", "break") which then needs to be used in the outer for-loop:
```python
# The for loop is the important part. The exact circuit is irrelevant.
for _ in range(10):
    opt = Optimizer(symplectic_lr=0.1, euclidean_lr=0.01)
    status = opt.minimize(cost_fn_eucl, by_optimizing=[D])
    if status == "break":
        break
```
I find the latter perhaps the nicest solution, although it wouldn't solve the posted issue directly, but rather change the accepted syntax.

elib20: Are there any advantages to having the "graceful exit"? I'm fine with an error being printed when I use KeyboardInterrupt.

thisac: The usual reasons for having a graceful exit is to be able to keep running code after the exit has been made, e.g., the example you had where the loop kept running even though a `KeyboardInterrupt` was raised (if that would have been advantageous). I reckon an example where this could be used is when you wish to abort an optimization but still keep the the data/minimized value(s) up to that point, to be able to save or plot it. Potentially useful if it's a long optimization and you don't want to discard the results just because it hasn't finished yet.

josh146: @thisac could this be solved by having 

```python
except KeyboardInterrupt:
    raise OptimizerInterruptedError("some message") from None
```

from what I recall, the `from None` suppresses the previous traceback.

On another note, it is always best practice not to include hardcoded `print` statements in libraries, since they are not configurable by users. I suggest either changing it to an error message, or using a `logger`.

thisac: Thanks @josh146. I think that might be the best solution here.

sduquemesa: Thanks all for your comments!

sduquemesa:
@thisac, I have implemented a logger for Mr Mustard as suggested by @josh146. This implied adding tests for the logger module, the same we use in strawberry fields. Can you please review them?

The test added here solve the issue of failing assertions when running `pytest` directly (as also experienced on strawberry fields). After review we can add this very same tests to SF to solve the issue there as well.

sduquemesa:
@thisac, I have implemented a logger for Mr Mustard on PR #107. For the changes applied here to work it is required to merge the logger implementation first.

sduquemesa:
I believe this is ready for a last round of review before merging 🚀

mrmustard/utils/training.py
```diff
@@ -74,7 +77,9 @@ def minimize(
                     self.opt_history.append(cost)
                     bar.step(math.asnumpy(cost))
         except KeyboardInterrupt:  # graceful exit
-            return
+            raise self.OptimizerInterruptedError(
+                "Stopping optimizer execution...", self.log
```
thisac, line 81: Perhaps rewording this slightly since it's now an exception message instead of a printed statement. For example, "Optimizer execution halted due to keyboard interruption", or something similar.

mrmustard/utils/training.py
```diff
@@ -85,10 +90,20 @@ def should_stop(self, max_steps: int) -> bool:
...
 
+    class OptimizerInterruptedError(Exception):
+        """A helper class to quietly stop execution without printing a traceback."""
+
+        def __init__(self, message: str, logger: Logger) -> None:
+            super().__init__()
+            logger.info(message)
+
+        def _render_traceback_(self):
+            pass
```
thisac, line 105: Is this actually needed when raising exception from `None`? 🤔

mrmustard/utils/training.py
```diff
@@ -85,10 +90,20 @@ def should_stop(self, max_steps: int) -> bool:
...
+                self.log.info("Loss looks stable, stopping here.")
                 return True
         return False
 
+    class OptimizerInterruptedError(Exception):
+        """A helper class to quietly stop execution without printing a traceback."""
+
+        def __init__(self, message: str, logger: Logger) -> None:
+            super().__init__()
+            logger.info(message)
```
thisac, line 102: What's the reason to put the logging inside of the exception initialization? Would it perhaps make more sense to have the logging where the exception is raised instead?

mrmustard/utils/training.py
sduquemesa: Raising from `None` still raises a traceback which is quite uninformative since the `KeyboardInterrup` exception is swallowed by raising `None`. The empty [`_render_traceback_`](https://ipython.readthedocs.io/en/stable/config/integrating.html#custom-exception-tracebacks) avoids this traceback to be printed out on ipy-interpreters, hence giving a less clumped message when stopping execution.

mrmustard/utils/training.py
sduquemesa: Agree!

mrmustard/utils/training.py
sduquemesa: No good reason indeed. It does make more sense to have the logging where you suggest 🤓

thisac (approved):
Looks good! Missing a changelog entry (and I realized it's missing for the logger implementation as well). 🙂 

I wonder why the tests don't seem to be finishing. Should it take 2+ hours, really? 🤔

PR #107, opened 2022-01-17
Implement `logger`

sduquemesa:
**Context:**
Modules on Mr Mustard use  `print` statements instead of a `logger` that is configurable by the user.

This PR implements a logger to be used on Mr Mustard

**Benefits:**
`print` statements are not configurable by users whereas `logger` can be configured by the application using Mr Mustard.

### DIFF
--- a/tests/test_logger.py
+++ b/tests/test_logger.py
@@ -140,11 +140,9 @@ class TestLoggerIntegration:
         there were not other configurations made."""
 
         logger = create_logger(module.__name__)
-        stream_name = re.search(r"[0-9]", logger.handlers[0].stream.name).group()
-
         assert len(logger.handlers) == 1
         # checks if stream is stderr (stream name for stderr is 8 or 9, whereas stdout 6 or 7)
-        assert stream_name in ["8", "9"]
+        assert logger.handlers[0].stream.fileno() in [8, 9]
 
     @pytest.mark.parametrize("module", modules_contain_logging)
     def test_custom_logger_before_mm_logger_with_higher_level(self, module):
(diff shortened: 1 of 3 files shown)


tests/test_logger.py
```diff
@@ -0,0 +1,170 @@
...
+    def test_default_mm_logger(self, module):
+        """Tests that stderr is set for the MM logger by default as stream if
+        there were not other configurations made."""
+
+        logger = create_logger(module.__name__)
+        stream_name = re.search(r"[0-9]", logger.handlers[0].stream.name).group()
+
+        assert len(logger.handlers) == 1
+        # checks if stream is stderr (stream name for stderr is 8 or 9, whereas stdout 6 or 7)
+        assert stream_name in ["8", "9"]
```
thisac, line 147: Seems like the same can be achieved by using the `fileno()` method, instead of applying a regex to find the stream name. Couldn't find any details regarding that method, though, so I'm not exactly sure exactly what it's supposed to do.

```suggestion
        assert len(logger.handlers) == 1
        # checks if stream is stderr (stream name for stderr is 8 or 9, whereas stdout 6 or 7)
        assert logger.handlers[0].stream.fileno() in [8, 9]
```

thisac (changes requested):
Nice work @sduquemesa! We should add this fix to SF, and also make sure that the tests are run when running `make test` (in SF). It seems like that might not be the case at the moment.

tests/test_logger.py
sduquemesa: Thanks @thisac, this indeed seem to be a cleaner approach. I couldn't find any documentation or information about the method either, the values here set were found by trial and error on different machines and running using `make test` and `pytest`.

thisac (approved):
Looks great @sduquemesa! 💯

PR #111, opened 2022-01-26
Improve pylint score of Mr Mustard

sduquemesa:
**Context:**
Mr Mustard uses `pylint` for code quality checking.

This PR refactors code pieces that were not touched by other `pylint-*` PRs, leaving Mr Mustard with an overall high score when running `pylint`.

**Benefits:**
Mr Mustard code quality score is above 9 points on `pylint`.

### DIFF
--- a/mrmustard/types.py
+++ b/mrmustard/types.py
@@ -12,7 +12,9 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-# pylint: disable=unused-wildcard-import
+"""A module containing all base type annotations."""
+
+# pylint: disable=unused-wildcard-import,wildcard-import
 
 from typing import *
 
--- a/mrmustard/types.py
+++ b/mrmustard/types.py
@@ -12,6 +12,8 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+# pylint: disable=unused-wildcard-import
+
 from typing import *
 
 # NOTE: when type-annotating with typevars, objects with the same typevars must have the same type
--- a/mrmustard/utils/graphics.py
+++ b/mrmustard/utils/graphics.py
@@ -12,17 +12,17 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+"""A module containing utility classes and functions for graphical display."""
+
+from copy import copy
+from typing import Tuple
 from rich.progress import Progress, TextColumn, BarColumn, TimeRemainingColumn
 import matplotlib.pyplot as plt
-from matplotlib.colors import Normalize
 from matplotlib import cm
 import numpy as np
-from mrmustard.types import *
 from mrmustard import settings
-from numba import njit
-from copy import copy
 
-
+# pylint: disable=disallowed-name
 class Progressbar:
     "A spiffy loading bar to display the progress during an optimization."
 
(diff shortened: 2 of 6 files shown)


mrmustard/types.py
```diff
@@ -12,6 +12,10 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+"""A module containing all base type annotations."""
+
+# pylint: disable=unused-wildcard-import,wildcard-import
+
 from typing import *
```
thisac, line 19: Possibly better to import the needed imports here explicitly instead? 🤔

mrmustard/utils/graphics.py
```diff
@@ -12,17 +12,17 @@
...
-from matplotlib.colors import Normalize
 from matplotlib import cm
 import numpy as np
-from mrmustard.types import *
 from mrmustard import settings
-from numba import njit
-from copy import copy
-
 
+# pylint: disable=disallowed-name
```
thisac, line 25: What name is disallowed? 👀

mrmustard/utils/training.py
```diff
@@ -12,17 +12,18 @@
...
+
+from mrmustard.types import List, Callable, Sequence, Trainable, Tensor
 from mrmustard.utils import graphics
 from mrmustard.logger import create_logger
 from mrmustard.math import Math
 
 math = Math()
 
-
+# pylint: disable=disallowed-name
```
thisac, line 26: 👀

mrmustard/utils/xptensor.py
```diff
@@ -12,11 +12,13 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+# pylint: disable=too-many-return-statements,too-many-branches,missing-function-docstring
```
thisac, line 15: Would be better to actually add missing docstrings here rather than just disabling the Pylint message. The other two might require more refactoring though.

thisac (commented):
We should be careful with disabling Pylint messages that would be better to address at some point (specifically, e.g., adding docstrings and removing wildcard imports). I'd say it's better to keep the lower score and fix them thoroughly later, rather than disabling them just to get a better score.

mrmustard/utils/graphics.py
sduquemesa: `bar`, because is used too often in examples together with `foo` and `baz`

mrmustard/utils/training.py
sduquemesa: A python programmer walks into a `bar` and asks for something imported 😅

mrmustard/types.py
sduquemesa: I think the idea with the `types` module is for it to be a proxy for `typing`/type annotations. I also thought of removing this wildcard and explicitly importing the needed types; however, when on a different module every time one is required to import a new type then it has to be added to the `types` module import — quite obscure for someone that might not be familiar with Mr Mustard.

I don't know how bad is to bloat the namespace with all the types though. Anyhow, when importing from a module other than `types`, one will import only whatever is required. 

What do you think?

thisac: Oh, I see. I think this is actually the way we've extended NumPy in PennyLane. I don't have any other suggestions, so I'd say we can keep it like this. 🙂

mrmustard/utils/graphics.py
thisac: Ah, maybe better rename it then to e.g., `pbar` or `progress_bar`?

thisac (approved):
Thanks @sduquemesa! 🎸

mrmustard/utils/graphics.py
sduquemesa: No need I think, the name makes sense in the context. Also, the warning is only disabled for the `Progressbar` class.

PR #112, opened 2022-01-27
fix logger test side effects

sduquemesa:
**Context:**
A logger was implemented on Mr Mustard on PR #107 following strawberry fields logger implementation. As discovered on strawberry fields, logger tests had side-effects on the logger configuration hence making tests that use the logger fail if they are executed after `test_logger`.

In specific, `test_logging_handler_defined` was setting `logger.propagate = False` meaning that messages are not passed to the handlers of ancestor loggers (see logger [docs](https://docs.python.org/3/library/logging.html#logging.Logger)). This in turn implied that, if no handlers were set for the specific logger, no message would be emitted hence making tests that check for output messages fail.

This PR incorporates the same changes applied on PR _fix failing logger tests [#675](https://github.com/XanaduAI/strawberryfields/pull/675)_.

**Benefits:**
Hopefully no failing logger tests for obscure reasons.

**Possible Drawbacks:**
None

**Related GitHub Issues:**
None

### DIFF
--- a/tests/test_logger.py
+++ b/tests/test_logger.py
@@ -63,25 +63,29 @@ def reset_logging(pytestconfig):
     logging.root.handlers = []
     root_level = logging.root.level
 
+    # Reset the logging specific configurations before test
+    for module in modules_contain_logging:
+        logger = logging.getLogger(module.__name__)
+        logger.handlers = []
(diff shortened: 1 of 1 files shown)


thisac (approved):
Looks good to me! Thanks @sduquemesa!

PR #114, opened 2022-02-02
set `State` modes with `kwargs` and `__getitem__`

### LINKED ISSUE #110: Homodyne and heterodyne detectors not accepting modes in kwargs

### Expected behavior

`Homodyne(modes=[m],...)` should apply a homodyne detector to mode m. Similarly for Heterodyne.

### Actual behavior

It applies it to mode 0.

### Reproduces how often

`Homodyne(modes=[m],...)` always applies it to mode 0.
Note that `Homodyne(...)[m]` works correctly, applying it to mode m.
Similarly for Heterodyne.

### System information

```shell
Python version:            3.8.0
Platform info:             Windows-10-10.0.19041-SP0
Installation path:         C:\Users\Rache\AppData\Local\Programs\Python\Python38\lib\site-packages\mrmustard
Mr Mustard version:        0.1.1
Numpy version:             1.19.2
Numba version:             0.53.1
Scipy version:             1.4.1
The Walrus version:        0.17.0
TensorFlow version:        2.7.0
Torch version:             None
```

### Source code

```shell
from mrmustard.lab import *
import numpy as np
from mrmustard.utils.graphics import mikkel_plot
import matplotlib.pyplot as plt

# Initial state is two modes with "diagonal" (angle=pi/2) squeezed state in mode 0
# and "vertical" (angle=0) squeezed state in mode 1
S1 = Sgate(modes=[0], r=1, phi=np.pi/2)
S2 = Sgate(modes=[1], r=1, phi=0)
initial_state = Vacuum(2) >> S1 >> S2

# Because the modes are separable, measuring in one mode should leave
# the state as simply the state in the unmeasured mode unchanged

# A homodyne measurement on the second mode should leave
# the diagonal state in the first mode but doesn't
final_state = initial_state << Homodyne(modes=[1],quadrature_angle=0,result=[0.3])
mikkel_plot(np.array(final_state.dm()))

# But coded like this does work
final_state = initial_state << Homodyne(quadrature_angle=0,result=[0.3])[1]
mikkel_plot(np.array(final_state.dm()))

plt.show()
```

### Tracebacks

_No response_

### Additional information

```shell
Proposed fix from @ilan-tz:

`Homodyne` initializes `DisplacedSqueezed` which initializes `State`.
Line 370 in states.py currently is 
`State.__init__(self, cov=cov, means=means, cutoffs=cutoffs)` 
but should be
`State.__init__(self, cov=cov, means=means, cutoffs=cutoffs, modes=modes)`

Similarly for `Heterodyne`, line 112 in `Coherent` should have the modes kwarg added.

In fact, in general the modes of the states do not seem to depend on the kwarg `modes` but just the length of the list of parameters given.
```

sduquemesa:
**Context:**
`Homodyne(modes=[m],...)` and `Heterodyne(modes=[m],...)` should apply a detector to modes `m`.

- `modes` kwarg is added to `State.__init__` in `Homodyne`
- adds `modes` kwarg to `Coherent` on `Heterodyne`
- test detectors are applied to correct modes
- test using `modes` kwarg or `__getitem__` give consistent results 

**Benefits:**
Setting state/detector modes using `modes` kwarg or `__getitem__` give consistent results.

**Possible Drawbacks:**
None

**Related GitHub Issues:**
Closes #110

### DIFF
--- a/tests/test_lab/test_detectors.py
+++ b/tests/test_lab/test_detectors.py
@@ -14,19 +14,32 @@
 
 from hypothesis import settings, given, strategies as st
 from hypothesis.extra.numpy import arrays
-from mrmustard.math import Math
 
-math = Math()
 import numpy as np
 import tensorflow as tf
 from scipy.stats import poisson
 
-from mrmustard.lab import *
+from mrmustard.math import Math
+from mrmustard.lab import (
+    PNRDetector,
+    Coherent,
+    Sgate,
+    Vacuum,
+    S2gate,
+    BSgate,
+    Attenuator,
+    Homodyne,
+    Heterodyne,
+    TMSV,
+    Dgate,
+    Fock,
+)
 from mrmustard.utils.training import Optimizer
 from mrmustard.physics import gaussian
 from mrmustard import physics
 from mrmustard import settings
 
+math = Math()
 np.random.seed(137)
 
 
--- a/tests/test_lab/test_detectors.py
+++ b/tests/test_lab/test_detectors.py
@@ -57,7 +57,7 @@ def test_detector_squeezed_state(r, phi, eta, dc):
     mean = np.arange(len(ps)) @ ps.numpy()
     expected_mean = eta * np.sinh(r) ** 2 + dc
     assert np.allclose(mean, expected_mean)
-    variance = np.arange(len(ps)) ** 2 @ ps.numpy() - mean ** 2
+    variance = np.arange(len(ps)) ** 2 @ ps.numpy() - mean**2
     expected_variance = eta * np.sinh(r) ** 2 * (1 + eta * (1 + 2 * np.sinh(r) ** 2)) + dc
     assert np.allclose(variance, expected_variance)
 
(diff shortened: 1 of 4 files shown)


tests/test_lab/test_states.py
```diff
@@ -175,3 +177,17 @@ def test_random_state_is_entangled():
     assert N1 > 0
     assert N2 > 0
     assert np.allclose(N1, N2)
+
+
+@given(modes=st.lists(st.integers(), min_size=2, max_size=5, unique=True))
```
thisac, line 182: This is really neat! I really want to look into hypothesis a bit more. 🥳

tests/test_lab/test_detectors.py
```diff
@@ -262,3 +271,27 @@ def test_norm_2mode_normalized():
 def test_norm_2mode_gaussian_normalized():
     leftover = Coherent(x=[2.0, 2.0]) << Coherent(x=1.0, normalize=True)[0]
     assert np.isclose(1.0, physics.norm(leftover), atol=1e-5)
+
+
+def test_homodyne_mode_kwargs():
+
+    S1 = Sgate(modes=[0], r=1, phi=np.pi / 2)
```
thisac, line 278: Would be nice to start adding docstrings for tests as well. At least a single line just describing what it's testing. Makes it easier both to review and when adding more tests.

thisac (approved):
Looks good to me! Don't forget to update the changelog!

tests/test_lab/test_detectors.py
sduquemesa: Agree!

PR #115, opened 2022-02-03
implement fidelity for mixed states

### LINKED ISSUE #109: Fidelity between two mixed states
### Before posting a feature request

- [X] I have searched exisisting GitHub issues to make sure the feature request does not already exist.

### Feature details

Compute the fidelity between two mixed states.

### Implementation

I tried using 
`trace( sqrtm( matmul( matmul( sqrtm(rho1), rho2), sqrtm(rho1) ) ) )**2`
but was hoping for something with a faster implementation.

### How important would you say this feature is?

2: Somewhat important. Needed this quarter.

### Additional information

_No response_

sduquemesa:
**Context:**
Fidelity for mixed states is not implemented

This PR implements a fidelity measure for mixed states on Mr Mustard

**Benefits:**
Mr Mustards has a fidelity measure for general states

**Possible Drawbacks:**
None

**Related GitHub Issues:**
Resolves #109

### DIFF
--- a/mrmustard/physics/fock.py
+++ b/mrmustard/physics/fock.py
@@ -218,16 +218,16 @@ def ABC(cov, means, full: bool, choi_r: float = None) -> Tuple[Matrix, Vector, S
 def fidelity(state_a, state_b, a_ket: bool, b_ket: bool) -> Scalar:
     r"""Computes the fidelity between two states in Fock representation."""
     if a_ket and b_ket:
-        min_cutoffs = (slice(min(a, b)) for a, b in zip(state_a.shape, state_b.shape))
+        min_cutoffs = [slice(min(a, b)) for a, b in zip(state_a.shape, state_b.shape)]
         state_a = state_a[min_cutoffs]
         state_b = state_b[min_cutoffs]
         return math.abs(math.sum(math.conj(state_a) * state_b)) ** 2
 
     if a_ket:
-        min_cutoffs = (
+        min_cutoffs = [
             slice(min(a, b))
             for a, b in zip(state_a.shape, state_b.shape[: len(state_b.shape) // 2])
-        )
+        ]
         state_a = state_a[min_cutoffs]
         state_b = state_b[min_cutoffs * 2]
         a = math.reshape(state_a, -1)
@@ -236,10 +236,10 @@ def fidelity(state_a, state_b, a_ket: bool, b_ket: bool) -> Scalar:
         )
 
     if b_ket:
-        min_cutoffs = (
+        min_cutoffs = [
             slice(min(a, b))
             for a, b in zip(state_a.shape[: len(state_a.shape) // 2], state_b.shape)
-        )
+        ]
         state_a = state_a[min_cutoffs * 2]
         state_b = state_b[min_cutoffs]
         b = math.reshape(state_b, -1)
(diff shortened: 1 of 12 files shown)


mrmustard/physics/fock.py
```diff
@@ -247,7 +247,13 @@ def fidelity(state_a, state_b, a_ket: bool, b_ket: bool) -> Scalar:
...
         )
 
-    raise NotImplementedError("Fidelity between mixed states is not implemented yet.")
+    # mixed state
+    return (
+        math.trace(
+            math.sqrtm(math.matmul(math.matmul(math.sqrtm(state_a), state_b), math.sqrtm(state_a)))
+        )
+        ** 2
+    )
```
thisac, line 256: Looks good, but it seems like @elib20 was "hoping for something with a faster implementation". Maybe this is exactly what he meant, but I think it's worth checking. 🙂 

Still nice to have support for mixed state calculations though.

tests/test_fidelity.py
```diff
@@ -5,85 +5,125 @@
...
-    )
-    assert np.allclose(expected, gp.fidelity(means1, cov1, means2, cov2, hbar=hbar))
+from mrmustard.physics import gaussian as gp, fock as fp
+
+from mrmustard.math import Math
+
+math = Math()
+
+
+class TestGaussianStates:
```
thisac, line 15: Are these tests simply organized in a class now, or are they changed in any way?

tests/test_fidelity.py
```diff
@@ -5,85 +5,125 @@
...
+    def test_fidelity_with_self(self):
+        """Test that the fidelity of two identical quantum states is 1"""
+        assert np.allclose(fp.fidelity(self.state1, self.state1, False, False), 1, atol=1e-4)
+
+    def test_fidelity_is_symmetric(self):
+        """Test that the fidelity is symmetric and between 0 and 1"""
+        f12 = fp.fidelity(self.state1, self.state2, False, False)
+        f21 = fp.fidelity(self.state2, self.state1, False, False)
+        assert np.allclose(f12, f21)
+        assert 0 <= np.real_if_close(f12) < 1.0
```
thisac, line 110: (minor) I would probably put these in two separate tests

tests/test_fidelity.py
```diff
@@ -5,85 +5,125 @@
...
+
+    def test_fidelity_formula(self):
+        """Test fidelity of mixed states agrees with definition.
+
+        Richard Jozsa (1994) Fidelity for Mixed Quantum States,
+        Journal of Modern Optics, 41:12, 2315-2323,
+        DOI: 10.1080/09500349414552171
```
thisac, line 117: Perhaps move this to the implementation instead, since this is the definition used there.

thisac: Not sure if this test adds much, though. Would be better to have a test with hard-coded values for the fidelity (that we know are correct).

thisac (approved):
Thanks Sebas! I'm happy to approve this. I have a few minor comments, but nothing blocking this being merged.

mrmustard/physics/fock.py
sduquemesa: Yes, this is just the same approach @elib20 was following.

I've been thinking on ways to improve performance:one could be implementing the calculation with numba. However, because the calculation is executed by the backend (TF or torch in this case) I don't see much gain on going to numba.

What do you think? do you have any guess on how that can be improved.

tests/test_fidelity.py
sduquemesa: Good point! I'll do that

sduquemesa:
Thanks @thisac for you review! 😎

PR #121, opened 2022-03-15
Add flexibility to apply the Interferometer

sylviemonet:
**Context:**
Add one argument for Interferometer in mrmustard/lab/gates.py

Add the argument 'modes' which indicates where the Interferometer is applied. And add the possible ValueError, when we have both arguments: num_modes and modes to guarantee they are consistent.

**Benefits:**
More flexibility for the application.


ziofil (changes requested):
great! (sorry it took so long). Please update the changelog 👍

sylviemonet:
> Great! (sorry it took so long). Could you update the changelog? : )
> 
> EDIT: I think the 'request changes' option was not the right one to request changes 🤔

I haven't found the changelog...

sduquemesa:
Hi @sylviemonet! The `CHANGELOG` file is in the `.github` folder. I can add a new entry for you if that's ok. 

Also, I'm pushing some commits to make your PR work with the recent version of MM. Hope that's alright 🙃

sylviemonet:
> Hi @sylviemonet! The `CHANGELOG` file is in the `.github` folder. I can add a new entry for you if that's ok.
> 
> Also, I'm pushing some commits to make your PR work with the recent version of MM. Hope that's alright 🙃

Thanks @sduquemesa! I hope it works. I found the changelog now.

sduquemesa:
@sylviemonet sure! that'd be very helpful. I think this is only missing the case in which the error is raised (when `modes` and `num_modes` are defined at the same time)

ziofil (approved):
👍

sylviemonet:
Are there some special black params for MM? I just blacked it but it doesn't work.

sduquemesa:
I usually do `black -l 100 mrmustard/ tests/` for formatting

sylviemonet:
> I usually do `black -l 100 mrmustard/ tests/` for formatting

Thanks, @sduquemesa , could you please tell me if the test is right (My first time writing the test about the raise error case) and in the right place (I'm not sure about this)?

tests/random.py
```diff
@@ -149,6 +149,18 @@ def random_Interferometer(draw, num_modes, trainable=False):
...
+@st.composite
+def random_Interferometer_param(draw, trainable=False):
+    n_modes = [0,1,2]
+    modes = [2,5,6]
+    try:
+        Interferometer(num_modes=num_modes, orthogonal_trainable=trainable, modes=modes)
+    except:
+        pass
+    else:
+        raise AssertionError("No error here!")
```
sduquemesa: Thanks @sylviemonet! I think this should be placed in `tests/test_lab/test_gates_fock.py`. Also, pytest has a functionality to test error scenarios — you can use it as follows:
```suggestion
def random_Interferometer_param():
    n_modes = [0,1,2]
    modes = [2,5,6]
    with pytest.raises(ValueError):
        Interferometer(num_modes=num_modes, modes=modes)
```
here `pytest.raises` will succeed if `ValueError` is raised and fail otherwise.

tests/test_lab/test_gates_fock.py
```diff
@@ -125,3 +125,10 @@ def test_fock_representation_mzgate(phi_a, phi_b):
     MZ = MZgate(phi_a=phi_a, phi_b=phi_b, internal=False)
     expected = mzgate(theta=phi_b, phi=phi_a, cutoff=20)
     assert np.allclose(expected, MZ.U(cutoffs=[20, 20]), atol=1e-5)
+
+
+def random_Interferometer_param():
+    n_modes = [0,1,2]
+    modes = [2,5,6]
+    with pytest.raises(ValueError):
+        Interferometer(num_modes=num_modes, modes=modes)
```
sduquemesa, line 134: Excellent 🚀

tests/random.py
sylviemonet: Thank you so much for this kind help. (I thought MM uses the Hypothesis for testing without Pytest...)

sduquemesa:
@sylviemonet let me check...

sduquemesa:
I think I found the issue: all test functions should start with `test_`, if not they won't run. I've changed the name.

sylviemonet:
> I think I found the issue: all test functions should start with `test_`, if not they won't run. I've changed the name.

Thanks for this tip, but I'm wondering why the tests in the random.py do not obey this rule.

sduquemesa:
Oh, `random.py` is not a test file. The functions there are used to generate random inputs to the tests, for example

```python
@st.composite
def modes(draw, num_modes):
```

produces a list of integers of length `num_modes` that are used to randomly define modes.

sduquemesa:
This is ready to merge @sylviemonet! 🚀 On you to press the green button!

sylviemonet:
> This is ready to merge @sylviemonet! 🚀 On you to press the green button!

Thanks a lot!

PR #127, opened 2022-04-19
update requirements.txt

sduquemesa:
**Context:**
A Buffer Overflow vulnerability exists in NumPy 1.9.x in the PyArray_NewFromDescr_int function of ctors.c when specifying arrays of large dimensions (over 32) from Python code, which could let a malicious user cause a Denial of Service.

This PR updates Mr Mustard's dependencies to the latest possible versions that support `python >= 3.8`.

**Benefits:**
Latest version of dependencies. No vulnerabilities from Numpy.

**Possible Drawbacks:**
None

**Related GitHub Issues:**
None

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -8,12 +8,16 @@ on:
 jobs:
   build:
     runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+      matrix:
+        python-version: ['3.8', '3.9', '3.10']
     steps:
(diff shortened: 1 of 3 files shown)


ziofil:
on conda you can install tensorflow only up to 2.7, not 2.8...

sduquemesa:
@thisac and @ziofil, I just updated the tests workflow to check all currently supported python versions (>= 3.8). I guess that'd be better than checking each environment by hand. What do you think?

PR #131, opened 2022-05-05
Fix param passthrough

ziofil:
**Context:**
Circuits don't allow for correlated parameters

Objects accept parameters other than simply an initial value. This allows for circuits like e.g. `Sgate(r=BS.theta)`.

**Benefits:**
We can make circuits that live on submanifolds and we can correlate gates with each other.

**Possible Drawbacks:**
We should explain somewhere that if a parameter is a function of trainable variables, the function has to be computed inside the cost function (e.g. `BSgate(theta = f(vars))` has to run inside the cost function, otherwise `f(vars)` is not traced for the gradients).

**Related GitHub Issues:**
None

### DIFF
--- a/tests/test_utils/test_opt.py
+++ b/tests/test_utils/test_opt.py
@@ -23,9 +23,13 @@ from thewalrus.symplectic import two_mode_squeezing
 from mrmustard.lab.gates import Sgate, BSgate, S2gate, Ggate, Interferometer, Ggate
 from mrmustard.lab.circuit import Circuit
 from mrmustard.utils.training import Optimizer
+from mrmustard.utils.parametrized import Parametrized
 from mrmustard.lab.states import Vacuum
 from mrmustard.physics.gaussian import trace, von_neumann_entropy
 from mrmustard import settings
+
+from mrmustard.math import Math
+math = Math()
 
 
 @given(n=st.integers(0, 3))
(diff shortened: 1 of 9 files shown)


tests/test_utils/test_opt.py
```diff
@@ -230,6 +235,29 @@ def cost_fn():
     assert np.allclose(np.sinh(circ.trainable_parameters["euclidean"][2]) ** 2, 1, atol=1e-2)
 
 
+def test_parameter_passthrough():
+    """Same as the test above, but with param passthrough"""
+    tf.random.set_seed(137)
+    r = np.arcsinh(1.0)
+    par = Parametrized(
+        r=math.new_variable(r, (0.0, None), "r"),
+        phi=math.new_variable(np.random.normal(), (None, None), "phi"),
```
sduquemesa, line 244: Should this variable be parametrized with hypothesis instead of defined here?

sduquemesa (approved):
Great @ziofil! I just suggest to extend a bit the changelog to better describe the change.

On a more general point, I wonder if parameters should be better represented by classes themselves, in that way they can self-contain all the relevant information and avoid having to include conditional logic when dealing with them — a parameter object would know if it is trainable or constant, if already defined as a tensor in the backend or not, etc. 🤔

tests/test_utils/test_opt.py
ziofil: > Great @ziofil! I just suggest to extend a bit the changelog to better describe the change.
> 
> On a more general point, I wonder if parameters should be better represented by classes themselves, in that way they can self-contain all the relevant information and avoid having to include conditional logic when dealing with them — a parameter object would know if it is trainable or constant, if already defined as a tensor in the backend or not, etc. 🤔

Yeah that's a good idea. I've thought about implementing a Parameter class when I wanted to track the history of each parameter and plot it, but haven't found time to work on it yet : /

ziofil: Thanks for the change in the changelog, it looks great!

PR #133, opened 2022-05-20
Refactor parametrized class

sduquemesa:
**Context:**

Currently parametrized objects in Mr Mustard use the `Parametrized` class to dynamically assigns attributes of classes. This is done on the `__init__` method of the Parametrized class and it

- transforms the incoming value into a `Parameter` instance
- each value with `{name}` is assigned as a Trainable or Constant given that the `{name}_trainable` is present in the kwargs. Constants are not to be modified during any optimization procedure.

Trainable parameters are of three types: Euclidian, Orthogonal or Symplectic. This mentioned types define the update rules of each trainable parameter during the optimization procedure.

This PR

- Removes assignment of static private fields from the `Parametrized` class. That is, only values that are related to the training procedure are given to the `Parametrized` class initializer method. All other values should be assigned statically on the child class.
- Creates a new class `Parameter` with subclasses `Constant` and `Trainable`. Objects of this class will always contain instances of the backend objects (a `Tensor`, `Variable` or `Constant`) within the `Parameter.value` field - meaning now the `Parametrized` class only deals with backend objects.
- Childs of the `Trainable` class (`Symplectic`, `Orthogonal`, `Euclidian`) have their own update rules definef on the `parameter_update` module.
- The `Parametrized` class will only assign `Parameter`s (and no other type of objects) dynamically as fields of its child classes.
- Refactors the codebase to use the new API.
- Creates a new package called `training` containing all objects and logic related to the training procedure of Mr Mustard circuits/objects.

**Benefits:**

- Separation of concerns: the `Parametrized` class only deals with trainable parameters
- Encapsulation: different types of trainable objects are aware of their own optimization procedure
- Abstraction: this abstracts the complexity of handling different types of optimization procedures. 
- Open for extension: new types of trainable variables can be easily added now by creating a new child of the `Trainable` class and registering it within the factory function of the `parameter` module. Its optimization procedures should be created on the 
`parameter_update` module and registered in the `parameter_update.param_update_method` dictionary.

**Possible Drawbacks:**

- This PR will change the Mr Mustrad's API, specifically of trainable variables belonging to the new `Parameter` class. To access the values of trainable one has to use the `value` property of the parameter.

A better approach to this refactoring is to use experimental Tensorflow's [Extension Types](https://www.tensorflow.org/guide/extension_type) to create trainable types that work seamlessly with TensorFlow's APIs and to register custom optimizers for Orthogonal and Symplectic trainables. However, this would completely remove portability to backends other than Tensorflow.

**Related GitHub Issues:**
None


mrmustard/utils/parameter.py
```diff
@@ -0,0 +1,164 @@
...
+    def __init__(self, value, bounds, name, owner=None) -> None:
+        pass
+
+    @abstractmethod
+    def update(self, grad, learning_rate) -> Tuple[Tensor, Tensor]:
+        pass
+
+
+class Symplectic(Trainable):
+    def __init__(self, value, bounds, name, owner=None) -> None:
```
ziofil, line 74: Symplectic and Orthogonal technically shouldn't support `bounds`.
Can we omit it from `__init__` and use `(None,None)` when calling `math.new_variable`?

mrmustard/utils/parametrized.py
```diff
@@ -31,73 +32,28 @@ class Parametrized:
...
-            self._param_names += [] if name.startswith("_") else [name]
+
+            # filter `{name}_trainable` or `{name}_bounds`` to become fields
+            # of the class as those kwargs are used to define the variables
+            if "_trainable" in name or "_bounds" in name:
+                continue
+
+            # convert into parameter class
+            is_trainable = kwargs.get(f"{name}_trainable", False) or math.is_trainable(value)
+            bounds = kwargs.get(f"{name}_bounds", None)
```
ziofil, line 47: we may need `(None, None)` as the default, but we can revisit this

mrmustard/utils/parametrized.py
```diff
@@ -31,73 +32,28 @@ class Parametrized:
...
+            param_value = create_parameter(value, name, is_trainable, bounds, owner)
+
+            # dynamically assign variable as attribute of the class
+            self.__dict__[f"{name}"] = param_value
```
ziofil, line 51: why not just `[name]` if it's already a string?

mrmustard/utils/parametrized.py
```diff
@@ -31,73 +32,28 @@ class Parametrized:
...
-                    [p for op in self._ops for p in op.trainable_parameters["orthogonal"]]
-                ),
-                "euclidean": math.unique_tensors(
-                    [p for op in self._ops for p in op.trainable_parameters["euclidean"]]
-                ),
-            }
-
-        return self._trainable_parameters  # default
+    def trainable_parameters(self) -> Sequence[Trainable]:
+        return [value for value in self.__dict__.values() if isinstance(value, Trainable)]
```
ziofil, line 55: in the current version this works also for circuits, which may be made out of other circuits and/or gates, so it needs to be able to recurse (in the current version this is done by the `for op in self._ops` part, though I had to hard code the name `_ops` into the Parameter class which is awkward because _ops is an attribute of the `Circuit` class)

mrmustard/utils/training.py
```diff
@@ -58,38 +63,24 @@ def minimize(
...
+                        for item in by_optimizing
+                        if isinstance(item, Parametrized)
+                    ]
+                )
+            )
+
+            while not self.should_stop(max_steps):
+                cost, grads = loss_and_gradients(cost_fn, trainable_params)
+                self.opt_history.append(cost)
+                for param, grad in zip(trainable_params, grads):
```
ziofil, line 80: love it!

ziofil:
Very nice Sebas! I love where this is going. Everything looks much more solid than before already.

mrmustard/utils/parametrized.py
sduquemesa: I think I have already addressed that by changing the signature of the [`new_variable` backend method](https://github.com/XanaduAI/MrMustard/blob/23f4f021909ff3c735d8f64e9b506417d2948d66/mrmustard/math/tensorflow.py#L195-L204) to accept bounds as `None` and replace the value to `(None, None)` if needed.

mrmustard/utils/parametrized.py
sduquemesa: Sure! before it was something like `self.__dict__[f"_{name}"]` and then forgot to change it 🙈

mrmustard/lab/abstract/transformation.py
```diff
@@ -393,15 +394,15 @@ def __repr__(self):
         table.add_column("Shape")
         table.add_column("Trainable")
         with np.printoptions(precision=6, suppress=True):
-            for name in self._param_names:
-                par = self.__dict__[name]
+            parameters = {k: v for k, v in self.__dict__.items() if isinstance(v, Parameter)}
```
ziofil, line 397: +1 for using `self.__dict__`!

mrmustard/utils/graphics.py
```diff
@@ -49,6 +49,7 @@ def __init__(self, max_steps: int):
             total=max_steps,
             loss=1.0,
             refresh=True,
+            visible=settings.PROGRESSBAR,
```
ziofil, line 52: nice!

PR #139, opened 2022-06-03
disable pytest deprecation warnings

sduquemesa:
**Context:**
Pytest ends with a long list of not-useful internal deprecation warnings.

Add file `pytest.ini` to the root directory of the project instructing pytest to suppress internal deprecation warnings.

**Benefits:**
Removes noise from pytest output.

**Possible Drawbacks:**
Either live dangerously with the warnings or check the code to eliminate the warnings this is fine fire

Warnings are coming from `flatbuffers` (which is not a direct dependency of Mr Mustard) and `keras_preprocessing` image module. Those are likely to be addressed by maintainers and removed in future releases.

**Related GitHub Issues:**
None


ziofil (approved):
💯

PR #140, opened 2022-06-04
Iterable states

ziofil:
**Context:**
Adding a new feature to iterate over the marginals of a multimode state.

With this PR we support the syntax (e.g.)
```python
for mode in state:
    print(mode.purity)
```
**Benefits:**
cleaner syntax.

**Possible Drawbacks:**
maybe one could expect that `for xxx in state` yields something other than the single-mode marginals.

**Related GitHub Issues:**
None


sduquemesa (approved):

ziofil:
Oh I still need to update the changelog

PR #141, opened 2022-06-22
Improve `__repr__` and `__str__` of Transformation and Circuit classes 🖨️

sduquemesa:
**Context:**
When printing a circuit a very rich output is obtained (a table with parameter properties and values) but no string is actually generated.

A new printable representation on `__repr__` and `__str__` methods is implemented for the `Circuit` and `Transformation` class. The previous output is generated only on notebooks, this is achieved by implementing the `_ipython_display_()` method on the `Transformation` class.

_Example_: 

```python
>>> from mrmustard.lab import Dgate, Ggate, Attenuator, BSgate

>>> G = Ggate(modes=2)
>>> D = Dgate(x = 0.1, y = -0.5, x_trainable=True, y_trainable=True)
>>> L = Attenuator(transmissivity=0.5)
>>> BS = BSgate(theta = 0.5, phi = 0, modes = [2,3])

>>> circ = G >> D >> L >> BS
>>> print(circ)

< Circuit | 4 ops | compiled = False >

>>> circ.__repr__()

Interferometer(symplectic = array([[-0.40319021,  0.80472941,  0.17596138, -0.44307475],       [ 0.16820243, -0.08208862,  0.70276445,  0.43407917],       [-0.33721183,  0.13506795, -0.69644504,  0.7456155 ],       [-0.97487464, -0.81378182, -0.15007701,  0.15970126]]))[0, 1] >> Dgate(x=array(0.1), y=array(-0.5), modes = [0]) >> Attenuator(nbar=array(0.), transmissivity=array(0.5), modes = [0]) >> BSgate(phi=array(0.), theta=array(0.5), modes = [2, 3])
```

**Benefits:**
A terminal-friendly printable representation of `Circuit` and `Transformation` objects.

**Possible Drawbacks:**
None

### DIFF
--- a/.github/CHANGELOG.md
+++ b/.github/CHANGELOG.md
@@ -45,6 +45,14 @@
 * The Parametrized and Training classes have been refactored. The new training module has been added
   and with it the new Parameter class: now trainable tensors are wrapped in an instance of Parameter.
   [(#133)](https://github.com/XanaduAI/MrMustard/pull/133)
+  
+* The string representations of the `Circuit` and `Transformation` objects have been improved:
+  the `Circuit.__repr__` method now produces a string that can be used to generate a circuit in
+  a identical state (same gates and parameters), the `Transformation.__str__` and objects
+  inheriting from it now prints the name, memory location of the object as well as the modes 
+  of the circuit in which the transformation is acting on. The `_markdown_repr_` has been implemented
+  and on a jupyter notebook produces a table with valuable information of the Transformation objects.
+  [(#141)](https://github.com/XanaduAI/MrMustard/pull/141)
 
 ### Bug fixes
 * Fixed a bug in the `State.ket()` method. An attribute was called with a typo in its name.


mrmustard/lab/abstract/transformation.py
```diff
@@ -386,6 +386,26 @@ def __eq__(self, other):
         return True
 
     def __repr__(self):
+
+        class_name = self.__class__.__name__
+        modes = self._modes
```
ziofil, line 391: isn't it better to use the property `self.modes`?

sduquemesa: Yes!

mrmustard/lab/abstract/transformation.py
```diff
@@ -386,6 +386,26 @@ def __eq__(self, other):
...
+
+        return f"{class_name}({params_str}, modes = {modes})".replace("\n", "")
+
+    def __str__(self):
+
+        class_name = self.__class__.__name__
+        modes = self.modes
+        return f"<{class_name} object at {hex(id(self))} acting on modes {modes}>"
+
+    def _ipython_display_(self):
```
elib20, line 408: For some reason, now when I create a circuit like

```
circuit = mm.Sgate(1,modes=[0,1])
circuit >>= mm.BSgate(np.pi/4,modes=[0,1])
```

and then have a cell in a jupyter notebook where I call `circuit`, it just prints an empty table.

elib20 (commented):
The behaviour of `__repr__` and `__str__` is as expected 💯 

However, the `_ipython_display_` no longer seems to work as expected.

mrmustard/lab/gates.py
```diff
@@ -539,6 +549,11 @@ def _validate_modes(self, modes):
                 f"Invalid number of modes: {len(modes)} (should be {self.symplectic.value.shape[1] // 2})"
             )
 
+    def __repr__(self):
+        modes = self.modes
+        symplectic = repr(math.asnumpy(self.symplectic.value))
+        return f"Interferometer(symplectic = {symplectic}){modes}".replace("\n", "")
```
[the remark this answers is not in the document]
sduquemesa: woops... good catch!

mrmustard/lab/abstract/transformation.py
sduquemesa: This method should be name `_repr_markdown`. This is how it looks now on a notebook for me

image

Is it looking good? Can you try it out?

elib20: It works for me now! Awesome!

PR #143, opened 2022-06-30
Sampling from measurements

sduquemesa:
**Context:**
Currently applying homodyne or heterodyne detection returns the probability of the outcome. It is of interest as well to be able to output samples from measurements.

- Adds `tensorflow-probability==0.17.0` to Mr Mustard's dependencies. [TensorFlow Probability](https://www.tensorflow.org/probability/overview) provides integration of probabilistic methods with automatic differentiation and hardware acceleration (GPUs). Pytorch also provides similar functionality through the [`torch.distributions`](https://pytorch.org/docs/stable/distributions.html) module.
- For gaussian states, samples are drawn from the gaussian PDF. The PDF has been implemented using TensorFlow's multivariate normal distribution.
- For Fock states...

**Benefits:**
...

**Possible Drawbacks:**
Adds an extra dependency  `tensorflow-probability` to Mr Mustard increasing the bundle size of the library.

### DIFF
--- a/mrmustard/lab/abstract/state.py
+++ b/mrmustard/lab/abstract/state.py
@@ -338,14 +338,28 @@ class State:
             remaining_modes = [m for m in other.modes if m not in self.modes]
 
             if self.is_gaussian and other.is_gaussian:
-                prob, cov, means = gaussian.general_dyne(
-                    other.cov,
-                    other.means,
-                    self.cov,
-                    self.means,
-                    other.indices(self.modes),
-                    settings.HBAR,
-                )
+
+                if getattr(self, "sample", False):
+                    quadrature_angle = getattr(self, "quadrature_angle")
+                    result, prob, cov, means = gaussian.general_dyne_sampling(
+                        other.cov,
+                        other.means,
+                        self.cov,
+                        quadrature_angle,
+                        other.indices(self.modes),
+                        settings.HBAR,
+                    )
+                else:
+                    prob, cov, means = gaussian.general_dyne(
+                        other.cov,
+                        other.means,
+                        self.cov,
+                        self.means,
+                        other.indices(self.modes),
+                        settings.HBAR,
+                    )
+                    result = self.means
+
                 if len(remaining_modes) > 0:
                     return State(
                         means=means,
@@ -353,7 +367,7 @@ class State:
                         modes=remaining_modes,
                         _norm=prob if not getattr(self, "_normalize", False) else 1.0,
                     )
-                return prob
+                return result
 
             # either self or other is not gaussian
             other_cutoffs = [
--- a/mrmustard/lab/detectors.py
+++ b/mrmustard/lab/detectors.py
@@ -16,7 +16,7 @@
 This module implements the set of detector classes that perform measurements on quantum circuits.
 """
 
-from typing import List, Tuple, Union, Optional
+from typing import List, Tuple, Union, Optional, Sequence
 from mrmustard.types import Matrix
 from mrmustard.training import Parametrized
 from mrmustard.lab.abstract import FockMeasurement
@@ -246,19 +246,26 @@ class Homodyne(DisplacedSqueezed):
     def __init__(
         self,
         quadrature_angle: Union[float, List[float]],
-        result: Union[float, List[float]] = 0.0,
-        modes: List[int] = None,
+        result: Optional[Union[float, List[float]]] = None,
+        modes: Optional[List[int]] = None,
         r: Union[float, List[float]] = settings.HOMODYNE_SQUEEZING,
     ):
-        quadrature_angle = math.astensor(quadrature_angle, dtype="float64")
-        result = math.astensor(result, dtype="float64")
-        x = result * math.cos(quadrature_angle)
-        y = result * math.sin(quadrature_angle)
-        r = math.astensor(r, dtype="float64")
-        super().__init__(
-            r=r,
-            phi=2 * quadrature_angle,
-            x=x,
-            y=y,
-            modes=modes,
-        )
+        if result is None:
+            self.sample = True
+            self.quadrature_angle = quadrature_angle
+            super().__init__(
+                r=r,
+                modes=modes,
+            )
+        else:
+            quadrature_angle = math.astensor(quadrature_angle, dtype="float64")
+            result = math.astensor(result, dtype="float64")
+            x = result * math.cos(quadrature_angle)
+            y = result * math.sin(quadrature_angle)
+            super().__init__(
+                r=r,
+                phi=2 * quadrature_angle,
+                x=x,
+                y=y,
+                modes=modes,
+            )
--- a/mrmustard/physics/gaussian.py
+++ b/mrmustard/physics/gaussian.py
@@ -16,6 +16,8 @@
 This module contains functions for performing calculations on Gaussian states.
 """
 
+import tensorflow as tf
+import tensorflow_probability as tfp
 from typing import Tuple, Union, Sequence, Any
 from numpy import pi
 from thewalrus.quantum import is_pure_cov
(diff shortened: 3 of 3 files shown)


mrmustard/lab/detectors.py
```diff
@@ -246,19 +246,26 @@ class Homodyne(DisplacedSqueezed):
     def __init__(
         self,
         quadrature_angle: Union[float, List[float]],
-        result: Union[float, List[float]] = 0.0,
-        modes: List[int] = None,
```
ziofil, line 250: thanks

mrmustard/physics/gaussian.py
```diff
@@ -16,6 +16,8 @@
 This module contains functions for performing calculations on Gaussian states.
 """
 
+import tensorflow as tf
```
ziofil, line 19: This should not be imported here. Any new math functionality that you need should go through the interface:
1. add the required new methods to the [math interface](https://github.com/XanaduAI/MrMustard/blob/main/mrmustard/math/math_interface.py)
2. implement them in the [tf backend](https://github.com/XanaduAI/MrMustard/blob/main/mrmustard/math/tensorflow.py)
3. use them here via `math.your_method`

ziofil (commented):
Looks great, except let's not start bypassing the math interface.

mrmustard/physics/gaussian.py
sduquemesa: Thanks @ziofil! This will be implemented as you describe. I've done it this way to test that it works as expected... polishing details to come 🧹

elib20 (commented):
I cannot say that I understand every single detail (especially the tensorflow aspects), but the logic of the methods are sound, and testing yields samples in line with the expected marginal distributions. 

I am curious as to what the speed bottleneck of the code is, if you know?

mrmustard/lab/detectors.py
```diff
@@ -246,15 +253,21 @@ class Homodyne(DisplacedSqueezed):
...
         quadrature_angle = math.astensor(quadrature_angle, dtype="float64")
-        result = math.astensor(result, dtype="float64")
-        x = result * math.cos(quadrature_angle)
-        y = result * math.sin(quadrature_angle)
-        r = math.astensor(r, dtype="float64")
+
+        if result is None:
+            self.sample = True
+            x = 0
+            y = 0
```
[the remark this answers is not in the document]
sduquemesa: This values are used to initialize the `DisplacedSqueezed` state and cannot be set to `None` because that class doesn't allow it. They are updated accordingly later during the computation.

sduquemesa:
Thanks @elib20 for your review! The speed bottleneck is the computation of the full pdf `p(x)=Tr[ ⍴ |x><x| ]` from which we sample. I've added a cache to not recalculate the `|x><x|` wavefunctions every time, gaining a decent speed up. 

A rough benchmark: running the test `test_sampling_mean_and_std` (500 shots) and assuming pytest overheads are constant, the test takes 74.60s without the cache and 7.88s with it.

mrmustard/lab/abstract/state.py
```diff
@@ -394,6 +345,120 @@ def primal(self, other: Union[State, Transformation]) -> State:
...
+        the states' Fock representation."""
+
+        # if both states are gaussian
+        if self.is_gaussian and other.is_gaussian:
+            return self._project_onto_gaussian(other)
+
+        # either self or other is not gaussian
+        return self._project_onto_fock(other)
+
+    def _project_onto_fock(self, other: State) -> Union[State, float]:
```
ziofil, line 359: I feel this method is in the wrong place. It should be in the physics module, because otherwise you have to import from lab (this messes with dependency inversion).

mrmustard/utils/homodyne.py
```diff
@@ -0,0 +1,227 @@
...
+
+    # build `\psi_n(x) \psi_m(x)` terms
+    omega_over_hbar = 1 / settings.HBAR
+    q_tensor = math.new_constant(estimate_quadrature_axis(cutoff), "q_tensor")
+    x = np.sqrt(omega_over_hbar) * q_tensor
+    hermite_polys = math.expand_dims(physicist_hermite_polys(x, cutoff), axis=-1)
+    hermite_matrix = math.matmul(hermite_polys, hermite_polys, transpose_b=True)
+
+    prefactor = np.empty_like(reduced_state.dm())
+    for idx, _ in np.ndenumerate(prefactor):
```
ziofil, line 199: if you just need the index you can use `np.ndindex`

mrmustard/lab/abstract/state.py
sduquemesa: `Homodyne` is here just for typechecking. I have removed it in favour of not imprting anything from lab, however the attribute `sample` will come only from `Homodyne` instances. The latter implies that there is a hidden dependency anyhow.

I left a comment in the code so it is easy to track in the future.

elib20:
If I measure homodyne on a subset of all modes, it returns a State. Is there an easy way to also access the outcome(s) sampled from the other mode(s)?

sduquemesa:
> If I measure homodyne on a subset of all modes, it returns a State. Is there an easy way to also access the outcome(s) sampled from the other mode(s)?

@elib20, currently no but it can be implemented with min— `Homodyne` could return a tuple containing the outcome and the conditional state.

I also noticed the use of plural: is multiple shots something of interest? I think in that case `Homodyne` could return a list of tuples (pairs of outcome and conditional state).

elib20:
I was using plural to refer to if you measured Homodyne on more than one mode, rather than multiple shots, but multiple shots could be good too!

mrmustard/lab/abstract/state.py
```diff
@@ -394,6 +345,117 @@ def primal(self, other: Union[State, Transformation]) -> State:
...
+        ]
+
+        projector_state = self  # state_m is the state used to project onto
+        outcome = None
+        # Only instances of `Homodyne` define the attribute "sampling". If this is
+        # True then homodyne sampling is performed.
+        if getattr(self, "sample", False):
+            # build pdf and sample homodyne outcome
+            outcome, projector_state = homodyne.sample_homodyne_fock(
+                state=other, quadrature_angle=self.phi.value / 2, mode=self.modes
```
elib20, line 383: Why is phi divided by 2 here?

sduquemesa: It is multiplied by 2 when defining the `DisplacedSqueezed` state [here](https://github.com/XanaduAI/MrMustard/blob/3ea6f4b9c7d6399726c4d9013ac9396a006bbe84/mrmustard/lab/detectors.py#L273).

mrmustard/utils/homodyne.py
```diff
@@ -0,0 +1,280 @@
...
+
+    # calculate prefactors of the PDF
+    omega_over_hbar = 1 / settings.HBAR
+    q_tensor = math.new_constant(estimate_quadrature_axis(cutoff), "q_tensor")
+    x = np.sqrt(settings.HBAR) * q_tensor
+    hermite_polys = physicist_hermite_polys(q_tensor, cutoff)
+
+    # taking only the real part because the imaginary part cancels out during summation,
+    # this saves some type casting of real matrices into complex ones.
+    reduced_dm = math.real(state_dm)
```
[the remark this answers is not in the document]
sduquemesa: fixed!

elib20 (approved):
We went through the math/units in detail and it passed the sanity checks.

mrmustard/math/math_interface.py
```diff
@@ -377,14 +377,18 @@ def hash_tensor(self, tensor: Tensor) -> int:
         """
 
     @abstractmethod
-    def hermite_renormalized(self, A: Matrix, B: Vector, C: Scalar, shape: Sequence[int]) -> Tensor:
+    def hermite_renormalized(
+        self, A: Matrix, B: Vector, C: Scalar, shape: Sequence[int], modified: bool
+    ) -> Tensor:
```
ziofil, line 382: why do we need the unmodified Hermite polynomials?

ziofil (approved):
seems good, just I don't see what the kw for (un)modified hermite polys is for.

mrmustard/math/math_interface.py
sduquemesa: removed

sduquemesa:
To make the arguments consistent with the walrus so it's easy to debug

mrmustard/math/math_interface.py
sduquemesa: For homodyne sampling I use the unmodified Hermite polynomials. I'll make it more explicit in the code.

sduquemesa:
> seems good, just I don't see what the kw for (un)modified hermite polys is for.

For homodyne sampling I use the unmodified Hermite polynomials. I'll make it more explicit in the code.

PR #144, opened 2022-07-04
Check for `Sequence` instead of `Iterable` on parameter extraction

sduquemesa:
**Context:**
`State` instances contain tensors as variables (for example, the symplectic or covariance matrix of the state). States being also instances of `Parametrized` are recursed to obtain the trainable and constant parameters.

This PR substitutes the recurrence check from `Iterable` to `Sequence`. This specific logic branch is targeted for objects that contain lists (hence a `Sequence`) of Parametrized instances, like a circuit. Since tensors are also instances of `Iterable`, every object containing a tensor as a variable will be recursed but will result in an error.

**Benefits:**
Resolves the bug
```pycon
>>> SqueezedVacuum(r=1.0).constant_parameters 
TypeError: Cannot iterate over a scalar tensor.

>>> SqueezedVacuum(r=1.0).trainable_parameters 
TypeError: Cannot iterate over a scalar tensor.
```
for all instances of `State`.

**Possible Drawbacks:**
None

**Related GitHub Issues:**
None

### DIFF
--- a/.github/CHANGELOG.md
+++ b/.github/CHANGELOG.md
@@ -44,12 +44,13 @@
 
 * The Parametrized and Training classes have been refactored. The new training module has been added
   and with it the new Parameter class: now trainable tensors are wrapped in an instance of Parameter.
-  [(#133)](https://github.com/XanaduAI/MrMustard/pull/133)
-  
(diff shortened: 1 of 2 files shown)


ziofil (approved):
good catch :)

ziofil:
don't forget to update the changelog

PR #155, opened 2022-09-13
Rotation gate unitary

sduquemesa:
**Context:**
The rotation gate unitary is currently being calculated using the Choi isomorphisms which generates `nan` for angles of value zero where the identity matrix should be returned.

The rotation gate being diagonal in the Fock basis is easy to implement with better speed performance. This PR implements the `U` method of the rotation gate overriding the default way of calculating the unitaries of Gaussian transforms.

**Benefits:**
Fast computation of the rotation gate Fock representation which avoids invalid numerical outcomes.

**Possible Drawbacks:**
None

### DIFF
--- a/mrmustard/lab/gates.py
+++ b/mrmustard/lab/gates.py
@@ -22,11 +22,10 @@ from typing import Union, Optional, List, Tuple, Sequence
 import numpy as np
 from mrmustard.types import Tensor
 from mrmustard import settings
-from mrmustard.lab.abstract import Transformation
 from mrmustard.training import Parametrized
 from mrmustard.physics import gaussian
-
 from mrmustard.math import Math
+from .abstract import Transformation
 
 math = Math()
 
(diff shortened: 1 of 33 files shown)


mrmustard/lab/gates.py
```diff
@@ -177,6 +178,47 @@ def __init__(
...
+        num_args = (
+            args.shape[0] if len(args.shape.as_list()) > 0 else 1
+        )  # number or arguments given to the gate
+        angles = np.zeros((num_modes,))
+        if num_args == 1:
+            # one arg for all modes
+            angles[modes] = args
+        elif num_args == len(modes):
+            # an arg for each mode
+            angles[modes] = args
```
elib20, line 215: Should this be angles = args? Otherwise there is no difference from the outcome of the preceding if statement

elib20 (commented):
Nice! A couple small questions

mrmustard/lab/gates.py
```diff
@@ -177,6 +178,47 @@ def __init__(
...
+        elif num_args == len(modes):
```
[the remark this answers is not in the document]
sduquemesa: yes!

mrmustard/lab/gates.py
sduquemesa: yes! Thanks for spotting this

elib20 (approved):

ziofil (approved):
I made a couple of changes, hopefully tests are all ok

PR #156, opened 2022-09-13
bugfix - hermitian conjugate 🐛

sduquemesa:
**Context:**
The transposition operation from the math module reverses all indexes by default.
```python
math.transpose(M[[1,2,3,4]]) = M[[4,3,2,1]]
```
This is not the desired behaviour when calculating the hermitian conjugate of an operator — because the function `math.dagger` uses `math.transpose` with default args under the hood meaning the result is incorrect. Currently,
```python
math.dagger(M[[1,2,3,4]]) = math.conjugate(math.transpose(M[[1,2,3,4]])) = math.conjugate(M[[4,3,2,1]])
```

To fix the issue, the dagger operation now swaps the 1st and 2nd half of indices of a tensor as is expected for a transposition on the dagger operation. Now,
```python
math.dagger(M[[1,2,3,4]]) = math.conjugate(M[[3,4,1,2]])
```

**Benefits:**
Appropriate dagger operation, fixing a bug for transformations acting on density matrices. 

**Possible Drawbacks:**
None

### DIFF
--- a/.github/CHANGELOG.md
+++ b/.github/CHANGELOG.md
@@ -59,6 +59,11 @@
 * Fixed a bug in the `State.ket()` method. An attribute was called with a typo in its name.
   [(#135)](https://github.com/XanaduAI/MrMustard/pull/135)
 
+* The `math.dagger` function applying the hermitian conjugate to an operator was incorrectly
+transposing the indices of the input tensor. Now `math.dagger` appropiately calculates the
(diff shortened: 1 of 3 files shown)


ziofil (approved):
looks good!

sduquemesa:
Thanks @ziofil. Just added a couple of tests catching the issue and hopefully helping to avoid it in the future 🤞🏻

PR #158, opened 2022-09-14
Include backend variables in optimization

sduquemesa:
**Context:**
Mr Mustard optimizer can optimize gate parameters but not values directly defined in the backend.

This PR allows the optimizer to keep track of variables directly defined in the backend, for example

```python
# %%
from mrmustard.lab import Dgate, Sgate, Attenuator, Vacuum, Coherent, SqueezedVacuum, Rgate
from mrmustard.training import Optimizer
from mrmustard.math import Math; math = Math()
import numpy as np

# rotated displaced squeezed state
rotation_angle = np.pi/2
target_state = SqueezedVacuum(r=1.0, phi=rotation_angle)

# angle of rotation gate
r_angle = math.new_variable(0, bounds = (0, np.pi), name='r_angle')
# trainable squeezing
S = Sgate(r = 0.1, phi=0, r_trainable = True, phi_trainable=False)

def cost_fn_sympl():
    state_out = Vacuum(1) >> S >> Rgate(angle = r_angle)
    return 1 - fidelity(state_out, target_state)

opt = Optimizer(symplectic_lr=0.1, euclidean_lr=0.1)
opt.minimize(cost_fn_sympl, by_optimizing=[S, r_angle])  

print(np.allclose(r_angle.numpy(), rotation_angle/2, atol=1e-3))
```
here `r_angle` is defined outside of a gate.

**Benefits:**
Allows for writing and then optimizing complicated functions that defines a parameter.

```python
def my_param(x,y,z):
    # some very long and complex function

x = math.new_variable(...)
y = math.new_variable(...)
z = math.new_variable(...)

def cost_fn():
    state = Vacuum(...) >> SomeGate(param=my_param(x,y,z), param_trainable=True)
    # process

opt.Optimize(cost_fn, by_optimizing=[x,y,z])
```

**Possible Drawbacks:**
None

**Related GitHub Issues:**
None

### DIFF
--- a/mrmustard/training/optimizer.py
+++ b/mrmustard/training/optimizer.py
@@ -21,7 +21,7 @@ from typing import List, Callable, Sequence
 from mrmustard.utils import graphics
 from mrmustard.logger import create_logger
 from mrmustard.math import Math
-from .parameter import Parameter, Trainable
+from .parameter import Parameter, Trainable, create_parameter
 from .parametrized import Parametrized
 from .parameter_update import param_update_method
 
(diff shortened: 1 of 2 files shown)


ziofil (approved):
super 👍

PR #167, opened 2022-10-14
Add generic workflow for un-triggered actions

sduquemesa:
**Context:**
PR #166 refactored CI checks leading to some stale check that prevent some PRs from merging.

Due to path filtering, a pull request that points to the develop branch of the repo will not trigger the build workflow on `builds.yml` and then workflows will stay stale and prevent merging. This PR adds a generic workflow, with the same name, will return true in any case `the build workflow in `builds.yml` is not triggered.

See github actions docs [here](https://docs.github.com/en/repositories/configuring-branches-and-merges-in-your-repository/defining-the-mergeability-of-pull-requests/troubleshooting-required-status-checks#handling-skipped-but-required-checks).

**Benefits:**
Avoid blocking merges for no reason.

**Possible Drawbacks:**
None


sduquemesa:
Closing this as I found a cleaner way to do it via github config

PR #168, opened 2022-10-17
Set default r_bounds to (None,None)

elib20:
Allows for optimization over all real-valued squeezing without needing to make phi trainable

### Before submitting

      test directory!

      `pip install black`.

**Context:**

**Benefits:**

**Possible Drawbacks:**

**Related GitHub Issues:**


ziofil:
Sorry, I don't think this is necessary. A user can set the bounds, and (None,None) is a perfectly valid choice, without us changing the defaults. The defaults (0, None) make sense as r is technically never negative, and phi = pi is equivalent to -r.

PR #171, opened 2022-10-21
Refactor Wigner function into module

sduquemesa:
**Context:**
Calculation of the Wigner function is already included in Mr Mustard but hidden in the graphics module

This PR moves the Wigner function calculation to its own module and numbifies it.

**Benefits:**
Now you can calculate the Wigner function using
```python
from mrmustard.utils.wigner import wigner_discretized

wigner_discretized(dm, q, p) # dm is a density matrix
```

It should be faster as well because it uses numba jit.

**Possible Drawbacks:**
None

### DIFF
--- a/mrmustard/utils/wigner.py
+++ b/mrmustard/utils/wigner.py
@@ -29,20 +29,20 @@ def wigner_discretized(rho, qvec, pvec, hbar=settings.HBAR):
         rho (complex array): the density matrix of the state in Fock representation
         xvec (array): array of discretized :math:`x` quadrature values
         pvec (array): array of discretized :math:`p` quadrature values
-        hbar (float): the value of ``\hbar``
+        hbar (optional float): the value of `\hbar`, defaults to ``settings.HBAR``.
 
     Retunrs:
-        tuple(array, array, array): array containing the discretized Wigner function, and the P and
-            Q coordinates in meshgrid form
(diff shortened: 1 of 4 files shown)


ziofil:
Nice! We can easily write the P and Q as well right?

sduquemesa:
Correct! Though I think all this functions will have a better place to be in soon with the representations. Right now it feels a bit missplaced.

mrmustard/utils/wigner.py
```diff
@@ -0,0 +1,67 @@
...
+        array: array containing the discretized Wigner function
+    """
+
+    Q = np.outer(pvec, np.ones_like(qvec))
+    P = np.outer(np.ones_like(pvec), qvec)
+    cutoff = rho.shape[-1]
+    A = (Q + P * 1.0j) / (2 * np.sqrt(hbar / 2))
+
+    cutoff = rho.shape[-1]
+    A = (Q + P * 1.0j) / (2 * np.sqrt(hbar / 2))
```
elib20, line 44: Repeated lines?

elib20 (approved):
Looks good! Why not replace the plot_wigner function in graphics with the new function?

mrmustard/utils/wigner.py
sduquemesa: Indeed. Good catch!
```suggestion
```

PR #174, opened 2022-11-08
settings repr

ziofil:
**Context:**
settings are not easy do discover

the settings object has now a nice repr

**Benefits:**
users can see al the settings at once

**Possible Drawbacks:**
None I can think of

**Related GitHub Issues:**
None

Screen Shot 2022-11-08 at 9 55 08 AM


sduquemesa (approved):
Great addition!

PR #175, opened 2022-11-09
optional callback at each opt step

ziofil:
**Context:**
Optimizer is opaque

in the minimize function now we can pass a callback that will be executed at the end of each step (with `trainable_parameters` as argument) and the return is stored in `self.callback_history`.

**Benefits:**
Can do lots of things, e.g.
test

**Possible Drawbacks:**
None

**Related GitHub Issues:**
None


zeyueN:
Thanks for the PR! This is the number 1 item on my wish list 😃

I think this already satisfies my need right now. But here's a few nice to haves:
- instead of a single callback, it could be more convenient to pass a dict/list of callbacks.
- callback function taking as argument the current step number of the optimizer, and the cost of the current step.
- callback function taking as argument the current state (e.g. lr) of the optimizer, or even better, the ability to update optimizer state (think LR scheduling).

ziofil:
Oh nice suggestions! We can do all of them in principle. For the dict/list of callbacks: no problem. For the other two, we need to make a choice no?

zeyueN:
Yes I think the first one is easy to add. The second one should be simple too, I would say at least step number can be added. I'm taking inspirations from [keras' callbacks](https://keras.io/guides/writing_your_own_callbacks/) mostly, but I don't think, at this point, we need dedicated "kinds" of callbacks like their callback class api.

sduquemesa (approved):
This looks great 🤙 Any chances to add some tests and check it integrates as expected?

ziofil:
Would you do it? 😇

sduquemesa:
Sure, I will

sduquemesa:
I am merging now and adding a ticket to add test to this PR.

PR #176, opened 2022-11-10
Bugfix dgate gradient

ziofil:
**Context:**
There was a bug in the gradient computation of the Gate

Bug is fixed (there was a numpy array at some point that was breaking the chain rule)

**Benefits:**
It works

**Possible Drawbacks:**
None

**Related GitHub Issues:**
None


sduquemesa (approved):
Great!

PR #177, opened 2022-11-15
Refactor `utils.homodyne` into `physics.fock`

sduquemesa:
**Context:**
PR #143 introduced sampling for homodyne measurements and with it a bunch of functionality in the `utils/homodyne.py` module. Most of this functions are related to the sampling in fock representation and are not only useful for homodyne sampling — take for example the calculation of the quadrature distribution.

This PR refactors the homodyne module into:

- `physics.fock` — functions related to the fock representation were moved into this module. Some of them are even split and written as separated functions such that they are available for use in other contexts (for example `quadrature_distribution` and `oscillator_eigenstate`).
- `math.caching` — the cache decorator used for the Hermite polys is refactored into this new module, this decorator is not only applicable to the Hermite polys but also to any function taking  a 1D tensor + int parameters. Now it can be used generically by any function with this same signature. The idea of this module is to contain caching functions of this kind.

Also

- Hermite polynomials have the modified flag removed (as per [this](https://github.com/XanaduAI/MrMustard/pull/143#pullrequestreview-1142728731) comment) and now only the regular polynomials are used. 

_Note:_ This PR only refactors code, meaning where the code is placed and _not_ what it does and _nor_ how it is done; there is no change to the logic whatsoever.

**Benefits:**
Pieces of the code that were there already are now reusable. The code for the sampling logic is more readable now.

**Possible Drawbacks:**
None

**Related GitHub Issues:**
None


mrmustard/math/caching.py
```diff
@@ -0,0 +1,42 @@
...
+    def cached_wrapper(hashable_array, cutoff):
+        array = np.array(hashable_array, dtype=np.float64)
+        return fn(array, cutoff)
+
+    @wraps(fn)
+    def wrapper(tensor, cutoff):
+        return cached_wrapper(tuple(tensor.numpy()), cutoff)
+
+    # copy lru_cache attributes over too
+    wrapper.cache_info = cached_wrapper.cache_info
```
ziofil, line 39: nice!

mrmustard/physics/fock.py
```diff
@@ -467,3 +469,204 @@ def trace(dm, keep: List[int]):
     # make it square on those indices
     dm = math.reshape(dm, dm.shape[: 2 * len(keep)] + (d, d))
     return math.trace(dm)
+
+
+@tensor_int_cache
+def oscillator_eigenstate(q: Vector, cutoff: int) -> Tensor:
```
ziofil, line 475: This is going to be useful for a lot of things! Thanks

mrmustard/math/caching.py
sduquemesa: thanks

mrmustard/physics/fock.py
```diff
@@ -467,3 +469,204 @@ def trace(dm, keep: List[int]):
...
+    Returns
+        (float): maximum quadrature value
+    """
+    if cutoff == 0:
+        xmax_estimate = 3
+    else:
+        # maximum q for a classical particle with energy n=cutoff
+        classical_endpoint = np.sqrt(2 * cutoff)
+        # approximate probability of finding particle outside classical region
+        excess_probability = 1 / (7.464 * cutoff ** (1 / 3))
```
ziofil, line 566: what is the 7.464 constant? 🤔

ziofil (approved):
looks all good!

mrmustard/physics/fock.py
sduquemesa: The probability of finding the particle outside the classically allowed region is approximately

image

i.e., one gets there by taking the probability density integrated between the classical boundaries and use the asymptotics of the Hermite polys [[wikipedia](https://en.wikipedia.org/wiki/Quantum_harmonic_oscillator#Highly_excited_states)].

PR #179, opened 2022-11-18
Use marginal distribution when drawing 🎨

sduquemesa:
**Context:**
Currently marginals are calculated form the Wigner function. In cases in which not all features of the state are captured on the Wigner function, the marginals contain negativities hence not representing a true probability density.

This PR makes `mikkel_plot` 
- calculate marginals independently from the Wigner function thus ensuring that the marginals are physical even though the Wigner function might not contain all the features of the state within the defined window,
- expose the `ticks`, `tick_labels` and `grid` arguments to customize the visualization — useful for example when checking that your state has peaks in those dreaded multiples of √(π)
- return the figure and axes for further processing if needed

**Benefits:**

- Marginals in the visualization will always represent physical states

_before_: note how this lion is not fully displayed in the Wigner function leading to negativities in the marginal distributions

image

_after_: although the big kittie is not displayed in its full glory its marginals show the true probability distribution

image

- More configurability of the visualization

```python
from matplotlib import cm
ticks = [-2*np.sqrt(2),2*np.sqrt(2)]
labels = [r"$-2\sqrt{\hbar}$", r"$2\sqrt{\hbar}$"]
graphics.mikkel_plot(dm, xticks=ticks, yticks=ticks, xtick_labels=labels, ytick_labels=[r"$-2\sqrt{\hbar}$", r"$2\sqrt{\hbar}$"], grid=True, cmap=cm.PuOr)
```

image

- Users can take the figure and axes for post-processing or storing

**Possible Drawbacks:**
This will _marginally_ increase the computation time for the visualization but shouldn't be too impactful.

**Related GitHub Issues:**
None

### DIFF
--- a/mrmustard/utils/graphics.py
+++ b/mrmustard/utils/graphics.py
@@ -69,7 +69,7 @@ def mikkel_plot(
     rho: np.ndarray,
     xbounds: Tuple[int] = (-6, 6),
     ybounds: Tuple[int] = (-6, 6),
-    ticks=[-5, 0, 5],
+    ticks=(-5, 0, 5),
     tick_labels=None,
     grid=False,
 ):
--- a/mrmustard/utils/graphics.py
+++ b/mrmustard/utils/graphics.py
@@ -20,6 +20,7 @@ import matplotlib.pyplot as plt
 from matplotlib import cm
 import numpy as np
 from mrmustard import settings
+from mrmustard.physics.fock import quadrature_distribution
 from .wigner import wigner_discretized
 
 # pylint: disable=disallowed-name
@@ -64,7 +65,7 @@ class Progressbar:
         return self.bar.__exit__(exc_type, exc_val, exc_tb)
 
 
-def mikkel_plot(rho: np.ndarray, xbounds: Tuple[int] = (-6, 6), ybounds: Tuple[int] = (-6, 6)):
+def mikkel_plot(rho: np.ndarray, xbounds: Tuple[int] = (-6, 6), ybounds: Tuple[int] = (-6, 6), ticks = [-5, 0, 5], tick_labels = None, grid = False):
     """Plots the Wigner function of a state given its density matrix.
 
     Args:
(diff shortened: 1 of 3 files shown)


mrmustard/utils/graphics.py
```diff
@@ -64,44 +65,55 @@ def __exit__(self, exc_type, exc_val, exc_tb):
         return self.bar.__exit__(exc_type, exc_val, exc_tb)
 
 
-def mikkel_plot(rho: np.ndarray, xbounds: Tuple[int] = (-6, 6), ybounds: Tuple[int] = (-6, 6)):
+def mikkel_plot(
+    rho: np.ndarray,
+    xbounds: Tuple[int] = (-6, 6),
+    ybounds: Tuple[int] = (-6, 6),
+    ticks=(-5, 0, 5),
```
ziofil, line 72: should we have xticks and yticks?

ziofil (approved):
nice!

mrmustard/utils/graphics.py
```diff
@@ -64,45 +65,70 @@ def __exit__(self, exc_type, exc_val, exc_tb):
...
     plt.subplots_adjust(wspace=0.05, hspace=0.05)
 
+    if xtick_labels is None:
+        xtick_labels = xticks
+    if ytick_labels is None:
+        ytick_labels = yticks
+
     # Wigner function
-    ax[1][0].contourf(X, P, W, 60, cmap=cm.RdBu, vmin=-abs(W).max(), vmax=abs(W).max())
+    ax[1][0].contourf(X, P, W, 120, cmap=cm.RdBu, vmin=-abs(W).max(), vmax=abs(W).max())
```
heltluke, line 116:
```suggestion
    ax[1][0].contourf(X, P, W, 120, cmap=cm.magma, vmin=-abs(W).max(), vmax=abs(W).max())
```
I think we should prefer a perceptually uniform sequential colormap https://matplotlib.org/stable/tutorials/colors/colormaps.html.

mrmustard/utils/graphics.py
```diff
@@ -64,45 +65,70 @@ def __exit__(self, exc_type, exc_val, exc_tb):
...
     ax[1][0].tick_params(direction="in")
     ax[1][0].set_xlim(xbounds)
     ax[1][0].set_ylim(ybounds)
     ax[1][0].grid(grid)
 
     # X quadrature probability distribution
-    ax[0][0].fill(xvec, ProbX, color=cm.RdBu(0.5))
-    ax[0][0].plot(xvec, ProbX)
-    ax[0][0].get_xaxis().set_ticks(ticks)
+    ax[0][0].fill(q, ProbX, color=cm.RdBu(0.5))
```
heltluke, line 129:
```suggestion
    ax[0][0].fill(q, ProbX, color=cm.magma(0.5))
```

mrmustard/utils/graphics.py
```diff
@@ -112,10 +138,10 @@ def mikkel_plot(rho: np.ndarray, xbounds: Tuple[int] = (-6, 6), ybounds: Tuple[i
     ax[0][0].grid(grid)
 
     # P quadrature probability distribution
-    ax[1][1].fill(ProbP, pvec, color=cm.RdBu(0.5))
-    ax[1][1].plot(ProbP, pvec)
+    ax[1][1].fill(ProbP, p, color=cm.RdBu(0.5))
```
heltluke, line 141:
```suggestion
    ax[1][1].fill(ProbP, p, color=cm.magma(0.5))
```

heltluke (commented):
Maybe we should also add a colorbar?

sduquemesa:
Thanks @heltluke for your comments! This is how it looks after applying your suggestions

image

What do you think about exposing the colormap arg so that the user can customize it to its liking? I believe the current colormap works very well on screen but agree is not the best for printing — in that case the user can choose one that better suits their needs.

PR #180, opened 2022-11-28
Bugfix rgate modes

ziofil:
**Context:**
The `Rgate` and the `Dgate` don't work correctly in Fock representation in some cases.

1. Fixed the `Rgate` and `Dgate` by removing the parser method and simplifying the code.
2. Added two functions in the fock module for explicitly applying an operator to a ket or to a dm (sandwich) which avoid constructing unitaries as large as the whole circuit. Now they are used in the `transform_fock` method of the `Transformation` class.
3. fixed a bug in the trace function
4. minor improvements

**Benefits:**

**Possible Drawbacks:**

**Related GitHub Issues:**


mrmustard/lab/gates.py
```diff
@@ -97,8 +96,9 @@ def U(self, cutoffs: Sequence[int]):
         """Returns the unitary representation of the Displacement gate using the Laguerre
         polynomials."""
 
-        N = len(cutoffs)
-        x, y = self._parse_modes_and_args(cutoffs)
+        N = self.num_modes
+        x = self.x.value * math.ones(N, dtype=self.x.value.dtype)
+        y = self.y.value * math.ones(N, dtype=self.y.value.dtype)
```
sduquemesa, line 101: Nice trick! Just leaving here a comment for future readers (this took me a while to understand): 

- this is taking scalars to vectors 
- if the arguments are vectors it will leave them untouched (well... multiplied by one)

hence this lines here take care of maintaining the API where one can use a single gate with 1) one parameter for multiple modes or 2) multiple parameters for multiple modes

mrmustard/lab/abstract/transformation.py
```diff
@@ -142,13 +146,10 @@ def transform_fock(self, state: State, dual: bool) -> State:
...
-            if d is not None:
-                self._modes = list(range(d.shape[-1] // 2))
-            elif X is not None:
-                self._modes = list(range(X.shape[-1] // 2))
-            elif Y is not None:
-                self._modes = list(range(Y.shape[-1] // 2))
+            for elem in self.XYd:
+                if elem is not None:
+                    self._modes = list(range(elem.shape[-1] // 2))
+                    break
```
sduquemesa, line 152: I think every instance of `Transformation` should assign the property `_modes` given the number of parameters or the user-given modes; In that sense, `_modes=None` would be forbidden. On the other hand, as we discussed, `Transformation`s shouldn't have to deal with modes at all and the circuit should take care of it.

For now I'd leave here a TODO comment:
```suggestion
            # TODO: transfomations should set `_modes` instead of leaving the property as `None`
            for elem in self.XYd:
                if elem is not None:
                    self._modes = list(range(elem.shape[-1] // 2))
                    break
```

mrmustard/lab/gates.py
```diff
@@ -221,7 +202,7 @@ def X_matrix(self):
 
     def U(self, cutoffs: Sequence[int]):
 
-        angles = self._parse_modes_and_args(cutoffs)
+        angles = self.angle.value * math.ones(self.num_modes, dtype=self.angle.value.dtype)
```
sduquemesa, line 205: same trick here

mrmustard/physics/fock.py
```diff
@@ -331,6 +331,56 @@ def purity(dm: Tensor) -> Scalar:
...
+        perm.insert(o, D - N + i)
+    op_dm = math.transpose(op_dm, perm)
+    op_dm_op = math.tensordot(
+        op_dm, math.conj(op), axes=[[o + D // 2 for o in op_modes], np.arange(N, 2 * N)]
+    )
+    # the N output indices of op are now at the end. We need to move them at op_modes + D//2
+    perm = list(range(D - N))
+    for i, o in enumerate(op_modes):
+        perm.insert(o + D // 2, D - N + i)
+    return math.transpose(op_dm_op, perm)
```
sduquemesa, line 381: wish we had [einops](https://github.com/arogozhnikov/einops) or something more verbose here... `einsum` could take care of all the operations and transposition in a one-liner but the issue is being able to construct the input string (which boils down again to the same issue of juggling indexes) 🤹

tests/test_lab/test_circuit.py
```diff
@@ -20,3 +20,9 @@
 from mrmustard.lab import *
 from mrmustard import settings
 from tests import random
+
+
+def test_circuit_placement():
+    assert Sgate(1.0)[1] >> Dgate(1.0)[0] == Dgate(1.0)[0] >> Sgate(1.0)[1]
+    assert Sgate(1.0)[1] >> Rgate(1.0)[0] == Rgate(1.0)[0] >> Sgate(1.0)[1]
+    assert Rgate(1.0)[1] >> Dgate(1.0)[0] == Dgate(1.0)[0] >> Rgate(1.0)[1]
```
sduquemesa, line 28: I think this test should be split for different gates, also it would be good to add test for the multiple parameter case and gates that act on a subset of modes

tests/test_physics/test_fock/test_fock.py
```diff
@@ -173,3 +173,14 @@ def test_dm_to_ket_error():
...
+
+def test_fock_trace():
+    state = Vacuum(2) >> Ggate(2)
+    from_gaussian = state.get_modes(0).dm([3])
+    from_fock = State(dm=state.dm([40])).get_modes(0).dm([3])
+    assert np.allclose(from_gaussian, from_fock, atol=1e-5)
+
+    from_gaussian = state.get_modes(1).dm([3])
+    from_fock = State(dm=state.dm([40])).get_modes(1).dm([3])
+    assert np.allclose(from_gaussian, from_fock, atol=1e-5)
```
sduquemesa, line 186: Here a test directly using the trace function should be added as well. This test should contain a known input and compare the expected output, in that way it directly checks the function and this current test is then checkin its integration with the rest of the API.

sduquemesa (changes requested):
Thanks for tackling this issue @ziofil! 

I believe that we should start testing more thoroughly to avoid coming back to the same issues and at the same start building a strong testing suite that will give us confidence that new code will work as expected and integrate well with existing code.

mrmustard/physics/fock.py
ziofil: I love the idea of having `einops`

tests/test_physics/test_fock/test_fock.py
ziofil: ok I'll improve this test

PR #181, opened 2022-12-13
allows for full cutoff specification

ziofil:
**Context:**
sometimes it's useful to compute the fock representation using different cutoffs for input-output indices of the same mode

transformations (gates, circuit) now accept also a list of double (or quadruple for choi) length which specifies the cutoffs per index (rather than per mode)

**Benefits:**
saves runtime if one needs e.g. to input fock states into a gaussian circuit

**Possible Drawbacks:**
none

**Related GitHub Issues:**

### DIFF
--- a/mrmustard/lab/abstract/transformation.py
+++ b/mrmustard/lab/abstract/transformation.py
@@ -234,7 +234,7 @@ class Transformation:
         return fock.fock_representation(
             choi_state.cov,
             choi_state.means,
-            shape=cutoffs * 2,
+            shape=cutoffs * 2 if len(cutoffs) == self.num_modes else cutoffs,
             return_unitary=True,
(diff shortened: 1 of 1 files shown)


sduquemesa:
Hi @ziofil! I really don't see what is going on here and how this is beneficial, maybe an example could help. Also, how come only the input arguments to the functions are changed but not the logic inside the function? (:scratching head:)

ziofil:
Hi @sduquemesa 
So, this PR allows for example to simulate low photon number inputs to a circuit without having to compute large cutoffs in those indices. The reason why it's so simple is that the cutoffs are passed as the shape to the hermite poly function.

sduquemesa (approved):
Thanks for the explanation 🤓

PR #182, opened 2022-12-13
changes norm to probability in repr of State

ziofil:
**Context:**
norm is confusing field in the repr of `State` (it's the sqrt of the probability if the state is pure and the probability if it's mixed).

Replace norm with probability 

**Benefits:**
Consistent meaning and also more practically useful

**Possible Drawbacks:**
some users may miss the good old norm?

**Related GitHub Issues:**


sduquemesa (approved):
Thanks @ziofil! The `norm` attribute remains to be the same, right? (sqrt of probability for pure states and probability for mixed states)

PR #187, opened 2023-01-04
MMTensor wrapper

ziofil:
**Context:**
In MM we work with tensors all the time and we need to contract indices that have physical meaning. 
Wouldn't it be great if we could assign names to the indices and have an easy time contracting them?
This can be a good starting point for the TN project.

** drumroll **  introducing `MMTensor`, a wrapper around tensors that allows for an `axis_label` keyword and that overrides `__matmul__` in order to implement the index contraction.
As a first use case I use it in `apply_op_to_dm` and `apply_op_to_ket`, which are used somewhere in the `Transformation` class.

**Benefits:**
Very easy to contract tensors now.

**Possible Drawbacks:**
One needs to remember to extract the wrapped array for some applications.

**Related GitHub Issues:**
None

### DIFF
--- a/mrmustard/lab/abstract/state.py
+++ b/mrmustard/lab/abstract/state.py
@@ -132,9 +132,8 @@ class State:
         if self._purity is None:
             if self.is_gaussian:
                 self._purity = gaussian.purity(self.cov, settings.HBAR)
-                # TODO: add symplectic representation
             else:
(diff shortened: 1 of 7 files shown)


ziofil:
I merged this branch into bargmann_methods. The two are too linked to be developed independently. I'll continue there.

PR #188, opened 2023-01-09
Choi application bugfix

ziofil:
**Context:**
The application of a choi operator to a density matrix was resulting in a transposed dm

Fixes the order of the indices in the application of a choi operator to dm and ket

**Benefits:**
Correct result

**Possible Drawbacks:**
None

**Related GitHub Issues:**
None


sduquemesa:
Thanks a lot for this fix!

sduquemesa (approved):

PR #189, opened 2023-01-10
Check if changelog entry is created on PRs

sduquemesa:
**Context:**
Always forgetting to add entries to the CHANGELOG file?

Now github will remind you to do so — this PR implements a new CI workflow checking if an entry has been added to the CHANGELOG file. This check is not mandatory meaning it won't block the ability to merge the PR, however checks will appear as failed. In case no changelog entry is needed one can use the `no changelog` label to disable the check.

This PR also adds the changelog entry for PR #188.

**Benefits:**
No more PRs without CHANGELOG entries

**Possible Drawbacks:**
None


ziofil (approved):
fantastic 💯

PR #191, opened 2023-01-17
Improving tests

ziofil:
**Context:**
Tests could improve in MM

Add several new tests and improves hypothesis strategies

**Benefits:**
Better test suite

**Possible Drawbacks:**
More tests to maintain? Nah

**Related GitHub Issues:**
None


sduquemesa:
Thanks @ziofil. This PR is moving test from pytest to hypothesis but not adding new tests/coverage? 

I see that test have been canceled by github actions a couple of times already because they take more than 6h to execute to completion. This may be due to the fact that hypothesis runs the same test 50 times in CI, contrary to 10 times when testing in local — Is this something we'd like to change? However, lowering this number reduces the chances of hypothesis to find edge cases.

ziofil:
Hey Sebas, thank you for the suggestions. I think 10 may be ok, considering that we run hypothesis plenty of times locally (at least I do) before pushing to github. What do you think?

sduquemesa (approved):

PR #194, opened 2023-01-23
Adds ray-based distributed trainer.

zeyueN:
**Context:**

Adds `map_trainer` as the interface for distributing optimization workflows using ray so that things can happen in parallel -- replacing the need for `for` loops.

_Code here has been personally looked at by Trudeau._

Demo notebook in recipes.

Documentation page for the `trainer` module is added with some examples, which is also in the docstring of `map_trainer` so that it can be conveniently read in Jupyter directly with shift+tab.

I've also had some weird problem with the unit tests (all passing locally) running on github action where some times it would just hang. I've tried removing my new tests one by one and adding them back again one by one, and it somehow worked at the end... I changed the testing workflow file with direct `pip install .[ray]` instead of building wheel first, which I don't think is the reason. Thanks Trudeau I guess? 

**Benefits:**
fast and simple experimentation -> more research done.

**Possible Drawbacks:**
more interface to introduce

**Related GitHub Issues:**

### DIFF
--- a/.github/CHANGELOG.md
+++ b/.github/CHANGELOG.md
@@ -2,6 +2,51 @@
 
 ### New features
 
+  * Ray-based distributed trainer is now added to `training.trainer`. It acts as a replacement for `for` loops and enables the parallelization of running many circuits as well as their optimizations. To install the extra dependencies: `pip install .[ray]`.
+  [(#194)](https://github.com/XanaduAI/MrMustard/pull/194)
+
+    ```python
+    from mrmustard.lab import Vacuum, Dgate, Ggate
+    from mrmustard.physics import fidelity
+    from mrmustard.training.trainer import map_trainer
+
+    def make_circ(x=0.):
+        return Ggate(num_modes=1, symplectic_trainable=True) >> Dgate(x=x, x_trainable=True, y_trainable=True)
+    
+    def cost_fn(circ=make_circ(0.1), y_targ=0.):
+        target = Gaussian(1) >> Dgate(-1.5, y_targ)
+        s = Vacuum(1) >> circ
+        return -fidelity(s, target)
+    
+    # Use case 0: Calculate the cost of a randomly initialized circuit 5 times without optimizing it.
+    results_0 = map_trainer(
+        cost_fn=cost_fn,
+        tasks=5,
+    )
+
+    # Use case 1: Run circuit optimization 5 times on randomly initialized circuits.
+    results_1 = map_trainer(
+        cost_fn=cost_fn,
+        device_factory=make_circ,
+        tasks=5,
+        max_steps=50,
+        symplectic_lr=0.05,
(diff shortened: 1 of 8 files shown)


.github/CHANGELOG.md
```diff
@@ -2,6 +2,51 @@
...
+    # Use case 1: Run circuit optimization 5 times on randomly initialized circuits.
+    results_1 = map_trainer(
+        cost_fn=cost_fn,
+        device_factory=make_circ,
+        tasks=5,
+        max_steps=50,
+        symplectic_lr=0.05,
+    )
+
+    # Use case 2: Run circuit optimization 5 times on randomly initialized circuits with custom parameters.
```
ziofil, line 36: is it still 5 times or 2 here?

mrmustard/training/trainer.py
```diff
@@ -0,0 +1,422 @@
...
+        ]
+    else:
+        raise ValueError(
+            f"`tasks` is expected to be of type int, list, or dict. got {type(tasks)}: {tasks}"
+        )
+
+    if not unblock:
+        # blocks and wait till all tasks complete to return the end results.
+        if pbar:
+            # results = [
```
ziofil, line 355: remove commented lines

mrmustard/training/trainer.py
```diff
@@ -0,0 +1,422 @@
...
+    Hacky: https://github.com/ray-project/ray/issues/5554
+    """
+    import ray  # pylint: disable=import-outside-toplevel
+
+    while futures:
+        done, futures = ray.wait(futures)
+        yield ray.get(done[0])
+
+
+def map_trainer(trainer=train_device, tasks=1, pbar=True, unblock=False, num_cpus=None, **kwargs):
```
ziofil, line 215: Can you add a short explanation that in practice the most common use case is to ignore the `trainer` keyword (as it defaults to `train_device`), pbar, unblock etc and just concentrate on `tasks` and `**kwargs`?

tests/test_training/test_trainer.py
```diff
@@ -0,0 +1,226 @@
...
+        len(set(r["cost"] for r in results))
+        >= (tasks if isinstance(tasks, int) else len(tasks)) - 1
+    )
+    assert all(isinstance(r["optimizer"], Optimizer) for r in results)
+    assert all((r["optimizer"].opt_history) for r in results)
+
+    # Check if optimization history is actually decreasing.
+    opt_history = np.array(results[0]["optimizer"].opt_history)
+    assert len(opt_history) == max_steps + 1
+    assert opt_history[0] - opt_history[-1] > 1e-6
```
ziofil, line 110: note that as a quirk of opt_history, the first value is fixed 0 (in case this affects this test)

ziofil (approved):
Fantastic :D

PR #198, opened 2023-02-04
modular Fock strategies

ziofil:
**Context:**
The Fock strategies that we have found (diagonal, factorial, wormhole..) are still not part of MM

This PR organizes that functionality modularly so that we can add the various parts together easily and mix/match them

**Benefits:**
Long awaited refactor

**Possible Drawbacks:**
Some methods will become deprecated

**Related GitHub Issues:**
None


ziofil:
Braking this PR up into two

PR #199, opened 2023-02-04
better type annotations

ziofil:
**Context:**
We never addressed type annotations in MrMustard

This PR begins making improvements all across the board on type annotations

**Benefits:**
Easier to spot errors and bugs, finally we can rely on a type checker as well

**Possible Drawbacks:**
One more thing to maintain, but it should be easy once set up

**Related GitHub Issues:**
None


.github/workflows/builds.yml
```diff
@@ -16,7 +16,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        python-version: ['3.8', '3.9', '3.10']
+        python-version: ['3.9', '3.10']
```
sduquemesa, line 19: why removing support for python 3.8?

.readthedocs.yml
```diff
@@ -1,15 +1,42 @@
...
 
-build:
-    image: latest
+# Optionally declare the Python requirements required to build your docs
+python:
+   install:
+   - requirements: doc/requirements.txt
+   - method: pip
+     path: .
+   system_packages: true
```
sduquemesa, line 42: missing new line at the end of file
```suggestion
   system_packages: true

```

.readthedocs.yml
```diff
@@ -1,15 +1,42 @@
+# version: 2
```
sduquemesa, line 1: If this is working correctly you can remove all the commented lines in this file

sduquemesa (changes requested):
Hi @ziofil! Are the new types not supported by python 3.8?

.github/workflows/builds.yml
ziofil: Because readthedocs doesn't know how to deal with one of the typing features (I am blanking on which one, but one of the things I introduced in the mm.typing module) unless python is >= 3.9. It was throwing an error and this was the only solution.

ziofil:
it's the tests: I can't use `issubclass(type(batch), Batch)` with python < 3.9

PR #200, opened 2023-02-16
Phase noise

ziofil:
**Context:**
We need a PhaseNoise gate in MM

Adds PhaseNoise *in Fock representation*

**Benefits:**
We have the PhaseNoise gate at least in Fock representation

**Possible Drawbacks:**
No Gaussian implementation possible means that when it's applied, all the rest gets converted to Fock representation

**Related GitHub Issues:**
None


fab1an-q:
Thanks a lot for the prompt delivery, @ziofil! Very impressive!

I just tested this and I ran into errors whenever I defined a circuit with the `Attenuator` gate occurring _after_ the new `PhaseNoise` gate.

For illustration, running this code:
```
circuit = (
    Sgate(0.456, modes=[0, 1])
    >> Attenuator(0.234, modes=[0])
    >> PhaseNoise(0.123, modes=[0])
    >> Attenuator(0.234, modes=[0])
)
state = Vacuum(2) >> circuit
```
will throw this error:

```
...
...
File ~/Code/MrMustard/mrmustard/math/tensorflow.py:387, in TFMath.hermite_renormalized(self, A, B, C, shape)
    384 if isinstance(shape, List) and len(shape) == 1:
    385     shape = shape[0]
--> 387 poly = hermite_multidimensional(
    388     self.asnumpy(A), shape, self.asnumpy(B), self.asnumpy(C), True, True, True
    389 )
    391 def grad(dLdpoly):
    392     dpoly_dC, dpoly_dA, dpoly_dB = tf.numpy_function(
    393         grad_hermite_multidimensional, [poly, A, B, C], [poly.dtype] * 3
    394     )

File ~/.conda/envs/xanadu/lib/python3.10/site-packages/thewalrus/_hermite_multidimensional.py:101, in hermite_multidimensional(R, cutoff, y, C, renorm, make_tensor, modified, rtol, atol)
     99 dtype = np.find_common_type([Rt.dtype.name, yt.dtype.name], [np.array(C).dtype.name])
    100 array = np.zeros(cutoffs, dtype=dtype)
--> 101 array[(0,) * num_indices] = C
    103 if renorm:
    104     values = np.array(_hermite_multidimensional_renorm(Rt, yt, array))

IndexError: too many indices for array: array is 2-dimensional, but 4 were indexed
```

The snippet runs fine if I comment out either the `PhaseNoise` gate or the _bottom_ `Attenuator` gate. I tested a few other constellations and so far it seems it's really only the `Attenuator` gate that doesn't like to occur after `PhaseNoise`. Also, I can make the above snippet work if I reduce it to a one-mode circuit.

So to sum it up, this error seems to occur when `PhaseNoise` is followed by `Attenuator` in a circuit that involves more than one mode.

ziofil:
Thank you Fabian! The issue was with the cutoffs passed to the `choi` method of `Transformation`. It is now fixed.

mrmustard/math/math_interface.py
```diff
@@ -668,6 +669,19 @@ def reshape(self, array: Tensor, shape: Sequence[int]) -> Tensor:
...
+
+        Args:
+            array (array): array to set the diagonal of
+            diag (array): diagonal to set
+            k (int): diagonal to set
+
+        Returns:
+            array: array with the diagonal set to ``diag``
+        """
+
```
ryk-wolf, line 684: Is it me or is there no code for that method set_diag? ^^'

tests/test_lab/test_gates_fock.py
```diff
@@ -210,6 +217,43 @@ def test_raise_interferometer_error():
         RealInterferometer(num_modes=num_modes, modes=modes)
 
 
+@given(phase_stdev=medium_float.filter(lambda x: x > 0))
+def test_phasenoise_creates_dm(phase_stdev):
```
ryk-wolf, line 221: Imho there's a discrepancy between the title of the test, the doc of the test and what the test actually evaluates. None of them is wrong but the fact that there's no match is a pity.
Testing that something isn't None doesn't tell you what it is, it tells it's not nothing. (what the test does)
Testing that something (a DM) is created doesn't entail it's correct, it merely guarantees it exists (name vs doc)

tests/test_lab/test_gates_fock.py
```diff
@@ -210,6 +217,43 @@ def test_raise_interferometer_error():
...
+    assert (Coherent(1.0) >> PhaseNoise(phase_stdev))._dm is not None
+    assert (Fock(10) >> PhaseNoise(phase_stdev))._dm is not None
+
+
+@given(phase_stdev=medium_float.filter(lambda x: x > 0))
+def test_phasenoise_symmetry(phase_stdev):
+    assert (Fock(1) >> PhaseNoise(phase_stdev)) == Fock(1)
+    settings.AUTOCUTOFF_MIN_CUTOFF = 100
+    assert (Thermal(1) >> PhaseNoise(phase_stdev)) == Thermal(1)
+    settings.AUTOCUTOFF_MIN_CUTOFF = 1
```
ryk-wolf, line 232: Would be nice to have this cleaning line as part of a fixture since it's repeated across tests.

mrmustard/lab/gates.py
```diff
@@ -797,3 +801,69 @@ def __init__(
...
+
+
+class PhaseNoise(Parametrized, Transformation):
+    r"""The phase noise channel.
+
+    The phase noise channel is a non-Gaussian transformation that is equivalent to
+    a random phase rotation.
+
+    Args:
+        phase_stdev (float or List[float]): the standard deviation of the (wrapped) normal
```
ryk-wolf, line 813: ```suggestion```
Args:
    phase_stdev: the standard deviation of the (wrapped) normal distribution for the angle of the rotation
    phase_stdev_trainable: whether the phase standard deviation should be trainable or not
    phase-stdev_bounds: the upper and lower bounds for the phase's standard deviation
    modes: the single mode this gate is applied to (default [0])

mrmustard/lab/gates.py
```diff
@@ -797,3 +801,69 @@ def __init__(
...
+            distribution for the angle of the rotation
+        modes (optional, list(int)): the single mode this gate is applied to (default [0])
```
ryk-wolf, line 815: `modes` (with an s) is a list of int but the description states **single**, a bit confusing

ryk-wolf (commented):
My limited knowledge of the maths/physics only lets me comment on the software perspective :'). 
Looks cool!
I'd recommend adding documentation to all functions (a number of them have nothing), even just a quick one-liner summing what's happening. Future contributors/reviewers/devs will be grateful ;).

PR #201, opened 2023-02-16
ray test fix

zeyueN:
**Context:**
Attempt to fix the (random) never ending tests on github actions.
I have so far rerun the test 5 times without the issue reoccurring. 🤞 

Forces ray to init with 1 cpu for testing.

**Benefits:**

**Possible Drawbacks:**

**Related GitHub Issues:**

### DIFF
--- a/.github/CHANGELOG.md
+++ b/.github/CHANGELOG.md
@@ -166,6 +166,9 @@ This is now fixed.
 a pure dm was converted to ket).
 [(#184)](https://github.com/XanaduAI/MrMustard/pull/184)
 
+* The ray related tests sometimes hangs in github action. This is an attempt to address it.
+[(#201)](https://github.com/XanaduAI/MrMustard/pull/201)
+
 ### Documentation
 
 ### Contributors 
--- a/.github/CHANGELOG.md
+++ b/.github/CHANGELOG.md
@@ -166,7 +166,7 @@ This is now fixed.
 a pure dm was converted to ket).
 [(#184)](https://github.com/XanaduAI/MrMustard/pull/184)
 
-* The ray related tests sometimes hangs in github action. This is an attempt to address it.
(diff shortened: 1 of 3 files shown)


.github/CHANGELOG.md
```diff
@@ -166,6 +166,9 @@ This is now fixed.
 a pure dm was converted to ket).
 [(#184)](https://github.com/XanaduAI/MrMustard/pull/184)
 
+* The ray related tests sometimes hangs in github action. This is an attempt to address it.
```
sduquemesa, line 169: I'd change the wording here to adhere more to the CHANGELOG style. I always assume people is reading this in the future so I write it in past tense.
```suggestion
* The ray related tests were hanging in github action causing test to halt and fail. Now ray is forced to init with 1 cpu when running tests preventing the issue.
```

tests/test_training/test_trainer.py
```diff
@@ -20,11 +20,16 @@
...
 import numpy as np
+import ray
 from mrmustard.lab import Vacuum, Dgate, Ggate, Gaussian
 from mrmustard.physics import fidelity
 from mrmustard.training import Optimizer
 from mrmustard.training.trainer import map_trainer, train_device, update_pop
 
+NUM_CPUS = 1
+ray.init(num_cpus=NUM_CPUS)
+print(f"RAY INITIATED: {ray.is_initialized()}")
```
sduquemesa, line 31: I think pytest supresses print statements, I believe using a logger is better — in that way pytest can capture and handle the logs accordingly.
```suggestion
logging.info(f"RAY INITIATED: {ray.is_initialized()}")
```
Don't forget to import the logging module.

sduquemesa (approved):
Thanks @zeyueN for the quick turn around on this! I left very minor comments that do not prevent this PR from being merged.

tests/test_training/test_trainer.py
zeyueN: Nice catch, this was used for debugging with `pytest -s`, no point for it being here anymore, will just delete.

PR #202, opened 2023-02-27
Bugfix misc

ziofil:
**Context:**
Fixing small bugs/typos before 0.4 release

- Threshold detector can now be correctly initialized
- changes in settings.HBAR are correctly reflected everywhere
- Interferometer and RealInterferometer can be placed on any set of modes
- number of decimals are respected in circuit drawer

**Benefits:**
Fewer bugs/typos

**Possible Drawbacks:**
None

**Related GitHub Issues:**
None


mrmustard/physics/fock.py
```diff
@@ -735,9 +735,7 @@ def estimate_quadrature_axis(cutoff, minimum=5, period_resolution=20):
     return xaxis
 
 
-def quadrature_distribution(
-    state: Tensor, quadrature_angle: float = 0.0, x: Vector = None, hbar: float = settings.HBAR
-):
+def quadrature_distribution(state: Tensor, quadrature_angle: float = 0.0, x: Vector = None):
```
sduquemesa, line 738: I would keep this as an optional argument. This allows to use the `fock` module as an independent utility module in scripts without the need of using the full config of MM. What do you think?

mrmustard/physics/fock.py
```diff
@@ -783,16 +781,13 @@ def quadrature_distribution(
     return x, math.cast(pdf, "float64")
 
 
-def sample_homodyne(
-    state: Tensor, quadrature_angle: float = 0.0, hbar: float = settings.HBAR
-) -> Tuple[float, float]:
+def sample_homodyne(state: Tensor, quadrature_angle: float = 0.0) -> Tuple[float, float]:
```
sduquemesa, line 784: same here, I'd keep the optional kwarg

mrmustard/lab/gates.py
```diff
@@ -499,8 +499,10 @@ def __init__(
         orthogonal_trainable: bool = False,
         modes: Optional[List[int]] = None,
     ):
-        if modes is not None and (num_modes != len(modes)):
-            raise ValueError(f"Invalid number of modes, got {len(modes)} but expected {num_modes}.")
+        if modes is not None and (
+            num_modes != len(modes) or any(mode >= num_modes for mode in modes)
+        ):
+            raise ValueError("Invalid number of modes and the mode list here!")
```
sduquemesa: I don't understand this error message. It would be useful to make it more verbose for the user.

sduquemesa (approved):
Left one comment but it does not prevent this from being merged

mrmustard/physics/fock.py
ziofil: We can do that too, but we can't initialize the arg using `settings.FOO` otherwise it won't change when we change the setting

mrmustard/physics/fock.py
ziofil: same here

mrmustard/lab/gates.py
ziofil: it shouldn't be this, I thought I had changed it. thanks

ziofil: that's odd, I don't have this line in my copy of the branch 🤔

PR #208, opened 2023-02-28
Add unitary update

sylviemonet:
Fix the bugs about the Riemannian optimization.

Add two optimization methods of the Interferometer (unitary update) and RealInterferometer (orthogonal update).


mrmustard/training/parameter.py
```diff
@@ -174,17 +178,27 @@ def __init__(self, value: Any, name: str, owner: Optional[str] = None) -> None:
...
     def __init__(self, value: Any, name: str, owner: Optional[str] = None) -> None:
-        self._value = (
-            value
-            if math.from_backend(value) and not math.is_trainable(value)
-            else math.new_constant(value, name)
-        )
+        if math.from_backend(value) and math.is_trainable(value):
+            self._value = value
+        elif type(value) in [list, int, float]:
+            self._value = math.new_variable(value, bounds=None, name=name)
```
ziofil, line 199: this is the `Constant` class, why are we making a variable?

ziofil (commented):
Nice! I only have that comment on the Constant class (which I didn't notice earlier)

mrmustard/training/parameter.py
sylviemonet: You are right. I forgot why I changed here. I just use the previous version of this part.

ziofil (approved):
looks good! :)

PR #215, opened 2023-03-03
Update requirements

sduquemesa:
**Context:**
Dependencies have not been revisited in a while.

Updates version of dependencies. In specific, tensorflow is pinned to version `2.10.1` since version `2.11.0` has changes in the functions `stack` and `concat` that break the integration with Mr Mustard.

To unpin it we will have to revisit the new tf api for those functions so that users installing MrMustard with `pip install mrmustard` won't run into unexpected errors.

Also, since `ray` is an optional dependency, tests related to ray are now skipped if the package is not installed.

**Benefits:**
Fixes a lot of security issues in TF as alerted by dependabot and also provides updated versions of the packages.

**Possible Drawbacks:**
TF is pinned — we should address this for the next release. Also, we should consider moving to a [PEP 621](https://peps.python.org/pep-0621/) compliant packaging system by either using a `pyproject.toml` file for dependency specification or a dependency manager like poetry.

**Related GitHub Issues:**
None


setup.py
```diff
@@ -21,8 +21,8 @@
     "scipy",
     "numba",
     "thewalrus>=0.17.0",
-    "tensorflow>=2.4.0",
-    "tensorflow-probability",
+    "tensorflow<=2.10.1",
```
ziofil, line 24: I had to comment this out and `python -m pip install tensorflow-macos=2.10.0`

ziofil (commented):
Should we add requirements-macos.txt ? Or could poetry help here?

sduquemesa:
Just  added a conditional dependency on `requirements.txt` that should at least help devs (but not users). Poetry is the way to go! 💪

setup.py
sduquemesa: I'm not sure how to resolve the issue on the `setup.py` file. Poetry would help with this.

ziofil: try importing `platform` (it's a built-in module) and then use the line `"tensorflow_macos==2.10.0" if platform.system() == "Darwin" else "tensorflow==2.10.1"`

ziofil: actually I can just try and see if it works

ziofil (approved):
it should work, but I can test it only locally

PR #217, opened 2023-03-06
Implement base 10 logarithm

sduquemesa:
**Context:**
There is no log10 function in tensorflow, only the loge exist ([ref](https://github.com/tensorflow/tensorflow/issues/1666))

Implements log base 10.

**Benefits:**
Well... a 10/10 implementation of log_10

**Possible Drawbacks:**
Could use `tf.function` to compile it but this function is not used in any performance-critical code so far.

**Related GitHub Issues:**
None

### DIFF
--- a/mrmustard/math/math_interface.py
+++ b/mrmustard/math/math_interface.py
@@ -469,6 +469,17 @@ class MathInterface(ABC):
         """
 
     @abstractmethod
+    def log10(self, x: Tensor) -> Tensor:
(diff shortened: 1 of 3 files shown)


ziofil (approved):

ziofil:
@sduquemesa this too, it was approved, see if we still need it

sduquemesa:
This is useful but not a deal breaker if not added to the math backend. I'll leave it to you to decide whether to add this change or close this PR.

PR #219, opened 2023-03-07
Improve callbacks and adds tensorboard

zeyueN:
**Context:**
To facilitate the investigation of the optimization process on the numerics side, as well as giving users more control over the optimization progress, the callback functionality is expanded to allow for realtime monitoring and updates. A builtin callback that tracks optimization progresses in Tensorboard is also provided.

- The `callback` argument to `Optimizer.minimize` is now `callbacks` with the new possibility of passing multiple callbacks.
- Adds tagged parameter traversal `Parametrized.traverse_trainables()` which accumulates the path of tags while traversing. This is now used by the optimizer to get unique human interpretable identifiers to parameters (for better tracking).
- Callback functions now accept as argument the optimizer, the cost value, the cost function, and the trainable parameter values as well as their gradient. This gives users a lot of run time control, but it's not necessary to use them all.
- The execution of callbacks moved to after the backprop but before the parameter update (originally at the end of the step).
- A `Callback` base class for users to extend and inherit from.
- A `TensorboardCallbacks` that does automatic tracking of costs and parameters (values and gradients) in real time.

Misc changes:

- I added a `path` item to the testing workflow so that non-code changes won't trigger it.

**Benefits:**
- Better observability.
- Finer control.
- Opens up a lot of possibilities for optimization.

**Possible Drawbacks:**
Users could introduce too much overhead if the callbacks are too heavy, thus potentially slowing down the optimization.

**Related GitHub Issues:**
None

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -4,6 +4,13 @@ on:
     branches:
       - develop
   pull_request:
+    paths:
+      - '.github/workflows/tests.yml'
+      - 'mrmustard/**'
+      - 'tests/**'
+      - 'setup.py'
+      - 'requirements*'
+      - 'pytest.ini'
 
 jobs:
   pytest:
--- a/mrmustard/training/callbacks.py
+++ b/mrmustard/training/callbacks.py
@@ -1,4 +1,4 @@
-# Copyright 2022 Xanadu Quantum Technologies Inc.
+# Copyright 2023 Xanadu Quantum Technologies Inc.
 
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
--- a/mrmustard/training/callbacks.py
+++ b/mrmustard/training/callbacks.py
@@ -32,7 +32,7 @@ class Callback:
     optimizer_step: int = 0
     callback_step: int = 0
 
-    def get_opt_step(self, optimizer, **kwargs):
+    def get_opt_step(self, optimizer, **kwargs):  # pylint: disable=unused-argument
         """Gets current step from optimizer."""
         self.optimizer_step = len(optimizer.opt_history)
         return self.optimizer_step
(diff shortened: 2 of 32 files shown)


.github/workflows/tests.yml
```diff
@@ -4,6 +4,13 @@ on:
     branches:
       - develop
   pull_request:
+    paths:
```
ziofil, line 7: what is `paths` for?

mrmustard/training/callbacks.py
```diff
@@ -0,0 +1,286 @@
+# Copyright 2022 Xanadu Quantum Technologies Inc.
```
ziofil, line 1: 2023?

mrmustard/training/callbacks.py
```diff
@@ -0,0 +1,286 @@
...
+    * `.update_grads`:
+        The custom grads modifyer, which is expected to return a list of parameter gradients
+        after modification, to be applied to the parameters.
+    * `.update_optimizer`:
+        The custom optimizer updater, which is expected to modify the optimizer inplace for
+        things like scheduling learning rates.
+
+    """
+
+    #: Custom tag for a callball instance to be used as keys in `Optimizer.callback_history`.
```
ziofil, line 103:
```suggestion
    #: Custom tag for a callback instance to be used as keys in `Optimizer.callback_history`.
```

ziofil (approved):
Fantastic. This is indeed opening up lots of possibilities 🤩

.github/workflows/tests.yml
zeyueN: It's used to specify a list of files that should trigger the workflow when changes are made to them, so that a commit that changes say changelog only won't rerun tests.

mrmustard/training/callbacks.py
zeyueN: No idea how this happened 🤦

PR #228, opened 2023-03-17
update

ziofil:
update version


sduquemesa:
We should also merge PR #230 onto main before doing the bugfix release

ziofil:
> We should also merge PR #230 onto main before doing the bugfix release

#232 now

elib20 (approved):

PR #229, opened 2023-03-17
Fix github release action

sduquemesa:
**Context:**
PyPI github action from master branch is deprecated and hence fails to upload a new release [[ref](https://github.com/marketplace/actions/pypi-publish#-master-branch-sunset-)].

Uses the `release/v1` version of the publish action as recommended by the documentation.

**Benefits:**
Hopefully automatic releases to pypi.

**Possible Drawbacks:**
Well, if it doesn't work then releases to pypi have to (still) be done manually.

**Related GitHub Issues:**
None


ziofil (approved):
👍  (how come pip install requirements is not needed?)

sduquemesa:
this action should use mrmustard's `setup.py` file to install the requirements: same as an user installing using pip from pypi

ziofil:
@sduquemesa can you close this or advise on what to do?

ziofil:
sorry - wrong button!

sduquemesa:
This is quite old and stale by now. If I'm not wrong the tooling team should have fixed that already; if not they can also do it in a way that is more consistent to the current MrMustard packaging and release workflow using poetry.

sduquemesa:
This is quite old and stale by now. If I'm not wrong the tooling team should have fixed that already; if not they can also do it in a way that is more consistent to the current MrMustard packaging and release workflow using poetry.

PR #236, opened 2023-04-05
v05 misc fixes

ziofil:
**Context:**
Various improvements

- fixed double bar when optimizing
- removed placeholder at beginning of opt_history
- fixed tests accordingly
- we now have rich tracebacks :)


tests/test_training/test_opt.py
```diff
@@ -107,16 +108,16 @@ def cost_fn():
     )
     assert np.allclose(np.cos(bs.theta.value) ** 2, k / (i + k), atol=1e-2)
     assert "null_cb" in opt.callback_history
-    assert len(opt.callback_history["null_cb"]) == (len(opt.opt_history) - 1) // 3
+    assert len(opt.callback_history["null_cb"]) == (len(opt.opt_history) + 2) // 3
```
zeyueN, line 111: I was expecting to have the `-1` removed, not replaced with a `+2`. There are a couple places callbacks depends on `len(opt_history)` , I'll take a look

zeyueN: I think we should now have an extra `+1` on [this line](https://github.com/XanaduAI/MrMustard/blob/a4ace7c6fd838dd3c2e76d79cf0ed764bc012d37/mrmustard/training/callbacks.py#L118).

ziofil: Thanks!

mrmustard/training/optimizer.py
```diff
@@ -212,13 +215,14 @@ def compute_loss_and_gradients(cost_fn: Callable, parameters: List[Parameter]):
...
         if max_steps != 0 and len(self.opt_history) > max_steps:
             return True
-        if len(self.opt_history) > 20:  # if cost varies less than 10e-6 over 20 steps
-            if (
-                sum(abs(self.opt_history[-i - 1] - self.opt_history[-i]) for i in range(1, 20))
-                < 1e-6
-            ):
+        self.__history_sum += abs(self.opt_history[-2] - self.opt_history[-1])
+        if len(self.opt_history) > 20:  # stop if cost varies less than 10e-6 over 20 steps
+            self.__history_sum -= abs(self.opt_history[-21] - self.opt_history[-20])
```
ryk-wolf, line 224: I can totally be mistaken, but it seems to me here that you're only looking at... 1 element? this:
`self.opt_history[-21] - self.opt_history[-20]`
is just 1 element afaik
is that what we want to be doing here?
I'm confused when seeing this line and this comment which mentions 20 steps
`# stop if cost varies less than 10e-6 over 20 steps`
Maybe I'm just misreading

mrmustard/training/callbacks.py
```diff
@@ -115,7 +117,7 @@ def __post_init__(self):
 
     def get_opt_step(self, optimizer, **kwargs):  # pylint: disable=unused-argument
```
ryk-wolf, line 118: As i see it, this function is doing 2 jobs: 
- setting self.optimizer_step
- returning optimizer_step

why not simply replace it by a computed property? and getting rid of the function.
Eg:
class Person():
    def __init__(self):
        self.n = "John Snow"
        self.age = 50
        self.is_old = False if self.age < 25 else True #computed property

ryk-wolf (commented):
Looking good!

I made a few style comments, nothing blocking.

mrmustard/__init__.py
```diff
@@ -33,9 +38,10 @@ def __new__(cls):  # singleton
     def __init__(self):
         self._backend = "tensorflow"
         self.HBAR = 2.0
-        self.CHOI_R = 0.881373587019543  # np.arcsinh(1.0)
         self.DEBUG = False
-        self.AUTOCUTOFF_PROBABILITY = 0.999  # capture at least 99.9% of the probability
+        self.AUTOCUTOFF_PROBABILITY = (
+            0.9999  # capture at least 99.99% of the probability by default
```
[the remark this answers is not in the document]
ziofil: my editor breaks long lines automatically

mrmustard/training/optimizer.py
ziofil: You are misreading. We maintain the difference of the last 20 steps: in line 222 we add the last difference and if opt_history is > 20 in line 224 we remove the difference beyond 20

mrmustard/utils/graphics.py
```diff
@@ -14,13 +14,19 @@
...
 
+# pylint: disable=wrong-import-order
+
 from typing import Tuple
-from rich.progress import Progress, TextColumn, BarColumn, TimeRemainingColumn
+
 import matplotlib.pyplot as plt
-from matplotlib import cm
 import numpy as np
+from matplotlib import cm
```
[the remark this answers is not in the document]
ziofil: I can't see the issue: both pyplot and cm are imported from matplotlib

tests/test_training/test_opt.py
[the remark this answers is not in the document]
ziofil: it's think it's pretty standard to have names begin with "test" (it also helps automated test discovery in vscode). But yes we could have more descriptive names.

mrmustard/training/callbacks.py
ziofil: yes it looks like it could be replaced by a property.

PR #247, opened 2023-06-01
Representations mvp project

sylviemonet:
**Context:** Enable MM can deal with different representations

**Description of the Change:** New MM Architecture about state.py file.

**Benefits:**

**Possible Drawbacks:**

**Related GitHub Issues:**


mrmustard/lab/representations/data/gaussian_data.py
```diff
@@ -0,0 +1,175 @@
...
+                batch_size = cov.shape[-3]
+                mean = math.zeros((batch_size, dim), dtype=cov.dtype)
+        else:
+            raise ValueError("You need to define at one: covariance or mean")
+
+        if coeffs is None:
+            batch_size = cov.shape[-3]
+            coeffs = math.ones((batch_size), dtype=mean.dtype)
+
+        # TODO : decide on what this is, whether we keep it do we keep it ?
```
ziofil, line 66: This hacky thingy enables this functionality `GaussianData(qdata)` where `qdata` is a `QuadraticData` object, i.e. it enables transformation between those two data types. There should be an equivalent one in `QuadraticData` do do the reverse transformation.

mrmustard/lab/representations/data/gaussian_data.py
```diff
@@ -0,0 +1,175 @@
...
+            super().__init__(mat=cov, vec=mean, coeffs=coeffs)
+
+
+        
+    @property
+    def cov(self) -> Batch[Matrix]:
+        return self.mat
+
+
+
```
ziofil, line 86: See comment above. Essentially there are two modes of operation: `GaussianData(cov, means, coeff)` and `GaussianData(qdata)`. If you prefer, we can put that code in a classmethod so we can do `GaussianData.from_QData(qdata)`.

ryk-wolf: Made it a classmethod. cheers!

mrmustard/lab/representations/data/data.py
```diff
@@ -0,0 +1,161 @@
...
+            Y (List[Union[List[Scalar],Scalar]])    : list of elements to be compared with xs
+            rtol (float)                            : relative tolerance parameter for Numpy's 
+                                                      `all_close`
+            atol (float)                            : absolute tolerance parameter for Numpy's 
+                                                      `all_close`
+
+        Returns:
+            True if all the elements compared are same within rtol/atol bounds, False otherwise
+        """
+        return all([np.allclose(x, y, rtol=rtol, atol=atol) for x, y in zip(X, Y)])
```
ziofil, line 48: why wouldn't `np.allclose(X,Y)` work?

mrmustard/lab/representations/data/data.py
```diff
@@ -0,0 +1,161 @@
...
+        r"""
+        Abstract method comparing two Data objects
+
+        Args:
+            other (Data)    : the object being compared
+
+        Returns:
+            True if both objects are equal, False otherwise
+        """
+        raise NotImplementedError()
```
ziofil, line 62: `raise NotImplementedError()` is not necessary when writing abstract methods because that line will never be executed. We can just use `pass` or even an ellipsis (three dots `...`).

mrmustard/lab/representations/data/gaussian_data.py
```diff
@@ -0,0 +1,180 @@
...
+                batch_size = cov.shape[-3]
+                mean = math.zeros((batch_size, dim), dtype=cov.dtype)
+        else:
+            raise ValueError("You need to define at one: covariance or mean")
+
+        if coeffs is None:
+            batch_size = cov.shape[-3]
+            coeffs = math.ones((batch_size), dtype=mean.dtype)
+
+        if isinstance(cov, QPolyData):
```
ziofil, line 66: do we still want to discuss about the API choice for this? (i.e. the option `GaussianData.from_QPolyData(qpoly)` vs `GaussianData(qpoly)`)

mrmustard/lab/representations/data/matvec_data.py
```diff
@@ -0,0 +1,183 @@
...
+                )
+
+            except AttributeError as e:
+                raise TypeError(f"Cannot add {self.__class__} and {other.__class__}.") from e
+
+
+
+    #@njit(parallel=True)
+    def __sub__(self, other: MatVecData, rtol: float = 1e-6, atol: float = 1e-6) -> MatVecData:
+        return self.__add__(other=other, atol=atol, rtol=rtol)
```
ziofil, line 97: add `-other` you mean?

mrmustard/lab/representations/data/matvec_data.py
```diff
@@ -0,0 +1,183 @@
...
+
+
+    #@njit(parallel=True)
+    def __add__(
+        self,
+        other: MatVecData,
+        rtol: float = 1e-6,
+        atol: float = 1e-6,
+        check_for_equality: bool = False,
+        sub:bool = False # TODO : find a more elegant way to subtract! cf below
```
ziofil, line 66: `__add__` only takes self and other as arguments. In particular, don't implement `sub` this way, just add self and -other.

mrmustard/lab/representations/data/matvec_data.py
```diff
@@ -0,0 +1,183 @@
...
+    def batch_size(self) -> int:
+        return self.coeffs.shape[0]
+
+
+    #@njit
+    def __neg__(self) -> MatVecData:
+        return self.__class__(mat=-self.mat, vec=-self.vec, coeffs=self.coeffs)
+
+
+    def __eq__(self, other: MatVecData, rtol: float = 1e-6, atol: float = 1e-6) -> bool:
```
ziofil, line 44: rtol and atol are not args of `__eq__` (see https://docs.python.org/3/reference/datamodel.html#emulating-numeric-types)

mrmustard/lab/representations/data/data.py
ryk-wolf: Because my X/Y can be matrices or vectors, afaik np.allclose works on vectors and scalars but not on matrices

mrmustard/lab/representations/data/gaussian_data.py
ryk-wolf: We've said so many things and went back on them that i don't remember what was agreed for this anymore tbh.

mrmustard/lab/representations/data/data.py
ryk-wolf: Because of what X and Y are/can be.
Np.allclose works on scalars and vectors.

X and Y can be:
X = [self.mat, self.vec, self.coeffs],
Y = [other.mat, other.vec, other.coeffs],

you can't stick that into np.allclose, you need to stick the vectors one by one.
or if you can let me know how you do it, I get errors ^^'

mrmustard/lab/representations/bargmann.py
```diff
@@ -0,0 +1,72 @@
...
+
+from mrmustard.math import Math
+from mrmustard.lab.representations.representation import Representation
+from mrmustard.lab.representations.data.qpoly_data import QPolyData
+from mrmustard.typing import Matrix, RealVector, RealMatrix, Scalar, Tensor, Vector
+
+math = Math()
+
+class Bargmann(Representation):
+    r""" Fock representation of a state.
```
ziofil, line 23:
```suggestion
    r""" Fock-Bargmann representation of a state.
```

mrmustard/lab/representations/bargmann.py
```diff
@@ -0,0 +1,72 @@
...
+        c: constants
+    """
+
+    def __init__(self, A: Matrix, b: Vector, c: Scalar) -> None:
+        self.data = QPolyData(A=A, b=b, c=c)
+
+
+    @property
+    def purity(self) -> Scalar:
+        raise NotImplementedError("Get this of this state from other representations!")
```
ziofil, line 37:
```suggestion
        raise NotImplementedError(f"This property is not available in {self.__class__.__qualname__} representation")
```

mrmustard/lab/representations/bargmann_dm.py
```diff
@@ -0,0 +1,22 @@
...
+# limitations under the License.
+
+from mrmustard.lab.representations.representation import Representation
+from mrmustard.typing import Matrix, Scalar, Vector
+
+class BargmannDM(Representation):
+
+    def __init__(self, A: Matrix, b: Vector, c: Scalar) -> None:
+        super().__init__(A=A, b=b, c=c)
+        self.num_modes = self.A.shape[-1] // 2
```
ziofil, line 22:
```suggestion
        self.num_modes = self.A.shape[-1] // 2

```

mrmustard/lab/representations/bargmann_ket.py
```diff
@@ -0,0 +1,22 @@
...
+# limitations under the License.
+
+from mrmustard.lab.representations.representation import Representation
+from mrmustard.typing import Matrix, Scalar, Vector
+
+class BargmannKet(Representation):
+
+    def __init__(self, A: Matrix, b: Vector, c: Scalar) -> None:
+        super().__init__(A=A, b=b, c=c)
+        self.num_modes = self.A.shape[-1]
```
ziofil, line 22:
```suggestion
        self.num_modes = self.A.shape[-1] 

```

mrmustard/lab/representations/data/array_data.py
```diff
@@ -0,0 +1,87 @@
...
+class ArrayData(Data):
+    """ Contains array-like data for certain Representation objects.
+
+    Args:
+        array : data to be contained in the class
+    """
+
+    def __init__(self, array:Vector) -> None:
+        self.array = array
+        self.cutoffs = self.array.shape
```
ziofil, line 33: should we keep `cutoffs` and `shape` separate? I think `cutoffs` should mean "hilbert space dimension", but shape can be more general than that (e.g. for a unitary, the shape is twice as long as the list of cutoffs)

sylviemonet: Yes you are right. That is the thing I was planning to add with Batch. Because now we don't have any Batch and the code is only related to the state, we can say the whole array's shape is the cutoffs. But when we refer to the transformations and Batch, that would be totally different meaning. Let's keep questioning this!!

mrmustard/lab/representations/data/array_data.py
```diff
@@ -0,0 +1,87 @@
...
+
+    
+    @property
+    def cutoffs(self) -> Union[int, List[int]]:
```
ziofil, line 37: is there a need for this to be a computed property?

mrmustard/lab/representations/data/array_data.py
```diff
@@ -0,0 +1,87 @@
...
+        """ The cutoffs for the Fock representation. """
+        return self.array.shape()
+
+
+    def __neg__(self) -> Data:
```
ziofil, line 42: I realized neg is the implementation of `not` so we probably don't need it

sylviemonet: Good observation. Thanks. Let me remove it.

PR #249, opened 2023-06-21
Compact fock correction

rdprins:
**Context:** There was a bug in the algorithms for faster PNR sampling from Gaussian circuits using density matrices. In the case where the cutoff of the first detector was equal to 1, the resulting density matrix was incorrect.

**Description of the Change:** The problem was solved by allowing for diagonal pivots to be used when the cutoffs[0]=1, effectively changing line 4 of Algorithm 1 in https://arxiv.org/pdf/2303.08879.pdf from `if diag_1 < C_1 − 1 then` to `if (diag_1 < C_1 − 1) or (C_1==1) then`.

### DIFF
--- a/.github/CHANGELOG.md
+++ b/.github/CHANGELOG.md
@@ -95,6 +95,9 @@
 * When projecting a Gaussian state onto a Fock state, the upper limit of the autocutoff now respect the Fock projection.
   [(#246)](https://github.com/XanaduAI/MrMustard/pull/246)
 
+* Fixed a bug for the algorithms that allow faster PNR sampling from Gaussian circuits using density matrices. When the 
+cutoff of the first detector is equal to 1, the resulting density matrix is now correct.
+
 ### Documentation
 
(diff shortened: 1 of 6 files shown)


ziofil (approved):
Thanks for the bugfix Robbe! 💪

.github/CHANGELOG.md
```diff
@@ -95,6 +95,9 @@
 * When projecting a Gaussian state onto a Fock state, the upper limit of the autocutoff now respect the Fock projection.
   [(#246)](https://github.com/XanaduAI/MrMustard/pull/246)
 
+* Fixed a bug for the algorithms that allow faster PNR sampling from Gaussian circuits using density matrices. When the 
+cutoff of the first detector is equal to 1, the resulting density matrix is now correct.
+
 ### Documentation
 
 ### Contributors
```
ziofil, line 103: don't forget to add yourself as a contributor to this release!

PR #253, opened 2023-06-26
Bugfix gradients

ziofil:
**Context:**
The `more_strategies` PR introduced several custom gradients, but we never noticed they were broken under some conditions.

- Gradients are now cast to the backend array type
- This PR also adds the numpy option to the bargmann methods

**Benefits:**
Gradients work

**Possible Drawbacks:**
None

**Related GitHub Issues:**
None

### DIFF
--- a/mrmustard/lab/abstract/state.py
+++ b/mrmustard/lab/abstract/state.py
@@ -526,8 +526,10 @@ class State:  # pylint: disable=too-many-public-methods
         self._modes = item
         return self
 
-    def bargmann(self) -> Optional[tuple[ComplexMatrix, ComplexVector, complex]]:
-        r"""Returns the Bargmann representation of the state."""
+    def bargmann(self, numpy=False) -> Optional[tuple[ComplexMatrix, ComplexVector, complex]]:
+        r"""Returns the Bargmann representation of the state.
+        If numpy=True, returns the numpy arrays instead of the backend arrays.
+        """
         if self.is_gaussian:
             if self.is_pure:
                 A, B, C = bargmann.wigner_to_bargmann_psi(self.cov, self.means)
@@ -535,6 +537,8 @@ class State:  # pylint: disable=too-many-public-methods
                 A, B, C = bargmann.wigner_to_bargmann_rho(self.cov, self.means)
         else:
             return None
+        if numpy:
+            return math.asnumpy(A), math.asnumpy(B), math.asnumpy(C)
         return A, B, C
 
     def get_modes(self, item) -> State:
--- a/tests/test_training/test_opt.py
+++ b/tests/test_training/test_opt.py
@@ -31,7 +31,7 @@ from mrmustard.lab.gates import (
     S2gate,
     Sgate,
 )
-from mrmustard.lab.states import DisplacedSqueezed, Gaussian, SqueezedVacuum, Vacuum
+from mrmustard.lab.states import DisplacedSqueezed, Fock, Gaussian, SqueezedVacuum, Vacuum
 from mrmustard.math import Math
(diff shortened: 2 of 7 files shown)


mrmustard/lab/abstract/state.py
```diff
@@ -526,15 +526,19 @@ def __getitem__(self, item) -> State:
         self._modes = item
         return self
 
-    def bargmann(self) -> Optional[tuple[ComplexMatrix, ComplexVector, complex]]:
-        r"""Returns the Bargmann representation of the state."""
+    def bargmann(self, numpy=False) -> Optional[tuple[ComplexMatrix, ComplexVector, complex]]:
```
elib20, line 529: Is there any function that needs numpy=True? I don't see one in this PR. We have several mrmustard functions that return tensorflow objects, then usually it's on the user to convert it to numpy if they need for their purposes.

tests/test_training/test_opt.py
```diff
@@ -457,3 +457,14 @@ def cost_fn():
 
     assert np.allclose(bsgate.theta.value, 0.1, atol=0.01)
     assert np.allclose(bsgate.phi.value, 0.2, atol=0.01)
+
+
+def test_squeezing_grad_from_fock():
```
elib20, line 462: Should we test the other gates that have custom gradients too?

elib20: This worked.

`from mrmustard.training import Optimizer, Parametrized
from mrmustard.lab import *

sq = SqueezedVacuum(r=1.0,r_trainable=True)

def cost_fn():
    return -((sq &Fock(1)) >> BSgate(0.5) << (Vacuum(1)&Fock(1)))

opt = Optimizer(euclidean_lr=0.05)
opt.minimize(cost_fn, by_optimizing=[sq], max_steps=100)`

elib20: This did not.

`from mrmustard.training import Optimizer, Parametrized
from mrmustard.lab import *

disp = Dgate(x=1.0,y=1.0, x_trainable=True,y_trainable=True)

def cost_fn():
    return -(Fock(2)>> disp << Vacuum(1))

opt = Optimizer(euclidean_lr=0.05)
opt.minimize(cost_fn, by_optimizing=[disp], max_steps=100)`

elib20: This worked.

`from mrmustard.training import Optimizer, Parametrized
from mrmustard.lab import *

bs = BSgate(theta=0.5,theta_trainable=True)

def cost_fn():
    return -((Fock(2,cutoffs=[3]) &Fock(1,cutoffs=[3])) >> bs << (Fock(2,cutoffs=[3]) &Fock(1,cutoffs=[3])))

opt = Optimizer(euclidean_lr=0.05)
opt.minimize(cost_fn, by_optimizing=[bs], max_steps=100)`

elib20 (commented):
Thanks for patching that so quickly!

I am still having issues with the Dgate.

sduquemesa:
Why no changelog for this change?

mrmustard/lab/abstract/state.py
ziofil: Not in this PR no, but I keep needing to transform the triple into numpy, so I just added the option here.

ziofil:
> Why no changelog for this change?

Because it's not fixing a bug in a previous version, so it wouldn't make sense to claim a bug was fixed in the upcoming release

tests/test_training/test_opt.py
ziofil: caught it!

elib20 (approved):
Nice!

PR #254, opened 2023-07-06
Fix of math.lattice.strategies.vanilla

rdprins:
### Before submitting

      test directory!

      `pip install black`.

**Context:** The lower neighbours of pivot incorrect at border of Fock representation. math.lattice.neighbours didn't check whether a certain entry of a pivot index is zero before subtracting 1 from it. Because of that, amplitudes at the opposite border of the Fock representation were read, while the neighbour should actually be disregarded.

**Description of the Change:** We now check the value of pivot indices before lowering them. 
A typo was also fixed in math.lattice.strategies.vanilla_jacobian. I still referred to paths.ndindex_path, which doesn't exist anymore. IIt was replaced by np.ndindex.

### DIFF
--- a/mrmustard/math/lattice/neighbors.py
+++ b/mrmustard/math/lattice/neighbors.py
@@ -29,7 +29,8 @@ from numba.cpython.unsafe.tuple import tuple_setitem
 def all_neighbors(pivot: tuple[int, ...]) -> Iterator[tuple[int, tuple[int, ...]]]:
     r"""yields the indices of all the neighbours of the given index."""
     for j in range(len(pivot)):  # pylint: disable=consider-using-enumerate
-        yield j, tuple_setitem(pivot, j, pivot[j] - 1)
+        if pivot[j] - 1 >= 0:
+            yield j, tuple_setitem(pivot, j, pivot[j] - 1)
         yield j, tuple_setitem(pivot, j, pivot[j] + 1)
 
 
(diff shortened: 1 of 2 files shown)


mrmustard/math/lattice/strategies/vanilla.py
```diff
@@ -68,7 +68,7 @@ def vanilla_jacobian(
     dGdc = G / c
 
     # initialize path iterator
-    path = paths.ndindex_path(G.shape)
```
ziofil, line 71: thanks, this function wasn't in use so there weren't tests to catch this

mrmustard/math/lattice/neighbors.py
```diff
@@ -29,7 +29,8 @@
 def all_neighbors(pivot: tuple[int, ...]) -> Iterator[tuple[int, tuple[int, ...]]]:
     r"""yields the indices of all the neighbours of the given index."""
     for j in range(len(pivot)):  # pylint: disable=consider-using-enumerate
-        yield j, tuple_setitem(pivot, j, pivot[j] - 1)
+        if pivot[j] != 0:
```
ziofil, line 32: Were the results actually wrong, or are they wrong only when you remove the k dependence? Because when k-1 goes negative usually the value is multiplied by sqrt(k), which should be zero.

mrmustard/math/lattice/strategies/vanilla.py
rdprins: I noticed because I used it for the parallel strategy to calculate marginals. (I needed the jacobian of the first hypercubes of the parallel simulations)

mrmustard/math/lattice/neighbors.py
rdprins: Sorry for the double communication, but it should indeed be ok for the vanilla strategy using the current form of the recurrence relation. For the renormalized recurrence relation, however, the sqrt(k) drops out.
Wouldn't it be better to check the values of the pivot index before lowering them by 1? Something like this might pop up with future Fock strategies.

ziofil (approved):
yep!

PR #255, opened 2023-07-10
ABC related modifications

sylviemonet:
Change the function of ABC in bargmann.py to the order of matrices that corresponds to the literature.
And change the order of the matrix index before using the hermite polynomial renormalized in Fock.py.


.github/CHANGELOG.md
```diff
@@ -95,14 +95,16 @@
 * When projecting a Gaussian state onto a Fock state, the upper limit of the autocutoff now respect the Fock projection.
   [(#246)](https://github.com/XanaduAI/MrMustard/pull/246)
 
+* Change the internal function to calculate ABC of a Bargmann representation into the correct ABC corresponds to the litterature and some other fix to get the correct Fock tensor. [(#255)](https://github.com/XanaduAI/MrMustard/pull/255)
```
ziofil, line 98:
```suggestion
* Fixed the internal function to calculate ABC of Bargmann representation (now corresponds to the literature) and other fixes to get the correct Fock tensor. [(#255)](https://github.com/XanaduAI/MrMustard/pull/255)
```

mrmustard/physics/bargmann.py
```diff
@@ -46,18 +46,15 @@ def wigner_to_bargmann_rho(cov, means):
...
     N = cov.shape[-1] // 2
-    A = math.matmul(
-        cayley(pq_to_aadag(cov), c=0.5), math.Xmat(N)
-    )  # X on the right, so the index order will be rho_{left,right}:
+    A = math.matmul(math.Xmat(N), cayley(pq_to_aadag(cov), c=0.5))
     Q, beta = wigner_to_husimi(cov, means)
-    B = math.solve(Q, beta)  # no conjugate, so that the index order will be rho_{left,right}
-    C = math.exp(-0.5 * math.sum(math.conj(beta) * B)) / math.sqrt(math.det(Q))
+    b = math.solve(Q, beta)
+    # B = math.matvec(math.Xmat(N), b)
```
ziofil, line 55: remove?

mrmustard/physics/bargmann.py
```diff
@@ -94,14 +91,14 @@ def wigner_to_bargmann_Choi(X, Y, d):
...
     ) / np.sqrt(2)
     A = math.matmul(math.matmul(R, A), math.dagger(R))
-    A = math.matmul(A, math.Xmat(2 * N))  # yes: X on the right
+    A = math.matmul(math.Xmat(2 * N), A)
     b = math.matvec(xi_inv, d)
     B = math.matvec(math.conj(R), math.concat([b, -math.matvec(XT, b)], axis=-1)) / math.sqrt(
         settings.HBAR, dtype=R.dtype
     )
-    B = math.concat([B[2 * N :], B[: 2 * N]], axis=-1)  # yes: opposite order
+    B = math.asnumpy(B)
```
ziofil, line 99: why convert to numpy?

mrmustard/physics/bargmann.py
```diff
@@ -94,14 +91,14 @@ def wigner_to_bargmann_Choi(X, Y, d):
...
     C = math.exp(-0.5 * math.sum(d * b) / settings.HBAR) / math.sqrt(math.det(xi), dtype=b.dtype)
-    # now A and B have order [out_l, in_l out_r, in_r].
+    # now A and B have order [out_r, in_r out_l, in_l].
```
ziofil, line 101: we want [out_l, in_l, out_r, in_r] no?

ziofil (approved):
good, just a couple of comments

mrmustard/physics/bargmann.py
sylviemonet: as I understand, we need [out_l, in_l out_r, in_r] before the Fock tensor, but here for ABC the order is [out_r, in_r out_l, in_l].

mrmustard/physics/bargmann.py
sylviemonet: I tested in a notebook with this part of code and I've got a list here not an array, so I'm afraid the type of this vector would be wrong and I just add one line to ensure it...

ziofil: but then we lose differentiability no?

mrmustard/physics/bargmann.py
ziofil: if you find any bugs, feel free to fix them as part of this PR (and add a test that would have failed)

mrmustard/physics/bargmann.py
sylviemonet: You are right, I was thinking to use the function in math would keep the differentiability... It would be better just delete this line.

PR #256, opened 2023-07-10
Utils for the Representations stage1 MVP

ryk-wolf:
**Context:**
Within the representations project, some tools were needed.

Implement a duck-style type-checker and a very generic factory method.

**Benefits:**
Allowing to check whether an object is of a certain class in a more pythonic way than `isinstance`. Allowing the creation of objects in tests (and beyond) through a dedicated factory.

**Possible Drawbacks:**
The current implementation of the duck type checker doesn't support checking python native types like booleans or ints, floats.

**Related GitHub Issues:**

### DIFF
--- a/.github/CHANGELOG.md
+++ b/.github/CHANGELOG.md
@@ -1,3 +1,16 @@
+# Release 0.6.0 (future)
+
+### New features
+* A general factory method and a duck style type-checker.
+
+### Breaking changes
+### Improvements
+### Bug fixes
+### Documentation
+### Contributors
+[Richard A. Wolf](https://github.com/ryk-wolf)
+
+
 # Release 0.5.0 (development release)
 
 ### New features
@@ -83,6 +96,7 @@
 
 * More robust implementation of cutoffs for States.
 [(#239)](https://github.com/XanaduAI/MrMustard/pull/239)
+
 
 ### Bug fixes
 
--- a/mrmustard/utils/misc_tools.py
+++ b/mrmustard/utils/misc_tools.py
@@ -42,9 +42,9 @@ def duck_type_checker(obj_a: object, obj_b: object) -> bool:
         True if both objects have exactly the same attributes, False otherwise.
 
     Raises:
-        TypeError:  If at least one of the objects given can't be compared via duck-typing, i.e.
-        lack attributes. A typical example would be python basic types such as ``int``, ``float``,
-        ``bool``, etc.
+        TypeError: If at least one of the objects given can't be compared via duck-typing, i.e.
+            lack attributes. A typical example would be python basic types such as ``int``, ``float``,
+            ``bool``, etc.
     """
     try:
         set_a = set(obj_a.__dict__.keys())
--- a/mrmustard/utils/misc_tools.py
+++ b/mrmustard/utils/misc_tools.py
@@ -1,0 +1,70 @@
+# Copyright 2023 Xanadu Quantum Technologies Inc.
+
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+
+#     http://www.apache.org/licenses/LICENSE-2.0
+
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+"""
+This module contains miscellaneous functions which are used accross the code base of MM. 
+They can be used as is in any of the classes.
+"""
+
+all = ["duck_type_checker", "general_factory"]
+
+
+def duck_type_checker(obj_a: object, obj_b: object) -> bool:
+    r"""*If it walks like a duck and it quacks like a duck, then it must be a duck.*
+
+    This function performs type-checking from a duck-type perspective, testing whether both given
+    objects share the same set of attributes. If they do, no matter their *type* -as obtained by
+    a call to ``isinstance``-, they will be considered to be of the same duck-type.
+
+    Note that we here look at the symmetric difference, hence the exact question we ask is better
+    phrased as "*Are both objects of the same type?*" than as "*Is object a of the same type as
+    object b?*".
+
+    Note also that this checks for the exact match between types -via attributes-, this system is
+    **not** meant to work with subclasses since there is a high probability that they might have
+    different attributes.
+
+    Args:
+        obj_a:  one of the objects to be checked
+        obj_b:  the other object to be checked
+
+    Returns:
+        True if both objects have exactly the same attributes, False otherwise.
+
+    Raises:
+        TypeError:  If at least one of the objects given can't be compared via duck-typing, i.e.
+        lack attributes. A typical example would be python basic types such as ``int``, ``float``,
+        ``bool``, etc.
+    """
+    try:
+        set_a = set(obj_a.__dict__.keys())
+        set_b = set(obj_b.__dict__.keys())
+        return set_a.symmetric_difference(set_b) == set()
+
+    except AttributeError as e:
+        raise TypeError(
+            f"Objects of types {type(obj_a), type(obj_b)} can't be compared via duck-type"
+        ) from e
+
+
+def general_factory(cls, *args, **kwargs) -> object:
+    r"""Factory method which generates an instance of cls parameterized by the given arguments.
+
+    Args:
+        args:   non-keyword ordered arguments used to parameterize the class instance
+        kwargs: keyword arguments used to parameterize the class instance
+
+    Returns:
+        An instance of the class cls parametrized by args and kwargs
+    """
+    return cls(*args, **kwargs)
--- a/tests/mock_classes/general_mocks.py
+++ b/tests/mock_classes/general_mocks.py
@@ -1,0 +1,38 @@
+# Copyright 2023 Xanadu Quantum Technologies Inc.
+
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+
+#     http://www.apache.org/licenses/LICENSE-2.0
+
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+"""
+This module provides general mock classes.
+"""
+
+all = ["MockAnimal", "MockFruit", "MockNoDefaultParams"]
+
+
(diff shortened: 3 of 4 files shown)


mrmustard/utils/misc_tools.py
```diff
@@ -0,0 +1,70 @@
...
+    objects share the same set of attributes. If they do, no matter their *type* -as obtained by
+    a call to ``isinstance``-, they will be considered to be of the same duck-type.
+
+    Note that we here look at the symmetric difference, hence the exact question we ask is better
+    phrased as "*Are both objects of the same type?*" than as "*Is object a of the same type as
+    object b?*".
+
+    Note also that this checks for the exact match between types -via attributes-, this system is
+    **not** meant to work with subclasses since there is a high probability that they might have
+    different attributes.
```
sduquemesa, line 35: I think within the context of this check the concept of "classes" and "subclasses" is irrelevant, meaning you can compare if objects are the same but cannot say anything about their inheritance structure. Because of this and to avoid potential confusion I suggest to remove this last bit
```suggestion
```

mrmustard/utils/misc_tools.py
```diff
@@ -0,0 +1,70 @@
...
+
+    Args:
+        obj_a:  one of the objects to be checked
+        obj_b:  the other object to be checked
+
+    Returns:
+        True if both objects have exactly the same attributes, False otherwise.
```
sduquemesa, line 42: missing return type
```suggestion
    Returns:
        bool: True if both objects have exactly the same attributes, False otherwise.
```

mrmustard/utils/misc_tools.py
```diff
@@ -0,0 +1,70 @@
...
+
+    Raises:
+        TypeError:  If at least one of the objects given can't be compared via duck-typing, i.e.
+        lack attributes. A typical example would be python basic types such as ``int``, ``float``,
+        ``bool``, etc.
```
sduquemesa, line 47: Missing indentation
```suggestion
    Raises:
        TypeError: If at least one of the objects given can't be compared via duck-typing, i.e.
            lack attributes. A typical example would be python basic types such as ``int``, ``float``,
            ``bool``, etc.
```

mrmustard/utils/misc_tools.py
```diff
@@ -0,0 +1,70 @@
...
+        set_b = set(obj_b.__dict__.keys())
+        return set_a.symmetric_difference(set_b) == set()
+
+    except AttributeError as e:
+        raise TypeError(
+            f"Objects of types {type(obj_a), type(obj_b)} can't be compared via duck-type"
+        ) from e
+
+
+def general_factory(cls, *args, **kwargs) -> object:
```
sduquemesa, line 60: Why do you need this function instead of directly calling `cls(*args, **kwargs)`? By the way, this looks a lot like a monad.

tests/mock_classes/general_mocks.py
```diff
@@ -0,0 +1,38 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+"""
+This module provides general mock classes.
+"""
+
+all = ["MockAnimal", "MockFruit", "MockNoDefaultParams"]
```
sduquemesa, line 18:
```suggestion
__all__ = ["MockAnimal", "MockFruit", "MockNoDefaultParams"]
```

tests/mock_classes/general_mocks.py
```diff
@@ -0,0 +1,38 @@
...
+
+
+class MockAnimal:
+    def __init__(self, age: int = 42, alive: bool = True, colour: str = "blue") -> None:
```
sduquemesa, line 22: A blue elephant! 🐘😅

tests/test_utils/test_misc_tools.py
```diff
@@ -0,0 +1,87 @@
...
+# limitations under the License.
+
+import pytest
+from copy import deepcopy
+from tests.mock_classes.general_mocks import *
+from mrmustard.utils.misc_tools import *
+
+
+###############################################################################
+##########################   duck_type_checker   ##############################
```
sduquemesa, line 22: Intead of a comment, you can wrap this in a class and then it will also get grouped in the pytest report
```suggestion
class TestDuckTypeChecker:
    """Tests related to the ducktype checker"""
```
You'll have to add `self` as the first argument to all the test methods.

tests/test_utils/test_misc_tools.py
```diff
@@ -0,0 +1,87 @@
...
+    assert duck_type_checker(obj_type_a, obj_type_b) == truth_val
+
+
+@pytest.mark.parametrize("x, y", [(1, 1.0), (1.0, "s"), ("s", 2.0), (True, 1.0), (MockAnimal(), 1)])
+def test_dtc_raises_TypeError_when_given_objects_without_dicts(x, y):
+    with pytest.raises(TypeError):
+        duck_type_checker(x, y)
+
+
+##########################   general_factory   ##############################
```
sduquemesa, line 62: Same comment as above

tests/test_utils/test_misc_tools.py
```diff
@@ -0,0 +1,87 @@
...
+    "cls, args, kwargs",
+    [
+        (MockAnimal, (1954, False), {"colour": "red"}),
+        (MockFruit, (), {}),
+        (MockNoDefaultParams, (), {"a": 1954, "b": "no"}),
+        (MockNoDefaultParams, (1954, "yes"), {}),
+    ],
+)
+def test_gfactory_returns_instance_of_correct_class(cls, args, kwargs):
+    assert isinstance(general_factory(cls, *args, **kwargs), cls) == True
```
sduquemesa, line 73: I still wonder what prevents you from doing `cls(*args, **kwargs)` directly 🤔 The behaviour expected from this and the following test would be exactly the same (i.e. create an instance and reaise an error when extra/wrong/unexpected args are given)

tests/test_utils/test_misc_tools.py
sduquemesa, line 1: I really like how this tests are written! Very easy to follow, granular and verbose.

sduquemesa (changes requested):
Great work @ryk-wolf! I'm leaving my first set of comments. In general I find the code very readable and the docstrings+tests make it very easy to review your code. Since I know this PR belongs to a larger project, my comments might come from the fact that I don't know how this utility functions are used within that context.

.github/CHANGELOG.md
```diff
@@ -1,3 +1,16 @@
...
+* A general factory method and a duck style type-checker.
+
+### Breaking changes
+### Improvements
+### Bug fixes
+### Documentation
+### Contributors
+[Richard A. Wolf](https://github.com/ryk-wolf)
+
+
```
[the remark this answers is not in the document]
ryk-wolf: Oh, really? Sorry, it's my first time managing a manual changelog. Where do you want me to put this?

mrmustard/utils/misc_tools.py
ryk-wolf: Ok so this brings up a wider question for me.
Whether we include the types in the doc or not.
My understanding was that we don't, to avoid redundancy of information. So while before I used to have something like:

def foo(bar:int) -> bool:
    """ Loremipsum.
    Args:
        bar (int): in dolores
    Returns:
        (bool) sint lament"""

I had changed to:
def foo(bar:int) -> bool:
    """ Loremipsum.
    Args:
        bar: in dolores
    Returns:
        sint lament"""
After reading the pennylane and a short chat with you because the way I understood it it was more valuable to avoid repeating info. So just want to make sure I understand things correctly so I can update my other code. What's our MM convention for typing in docstring? I'm happy with anything, just want to get it right ^^'.

tests/test_utils/test_misc_tools.py
ryk-wolf: What prevents me from doing it? Uhm, I'd say nothing beyond personal tastes xD. I can do that if you'd rather. Since that's exactly the one and only line of code contained in the `general_factory` anyway :').
Now my rationale was to still implement this general factory for 2 reasons:
1) There are no factory methods currently in MM afaik, and as we had briefly discussed it would be great to introduce them, so my thoughts were that this would be a first baby step in introducing the general concept of a factory there.
2) When we add more factories, we can probably inherit from the general_factory tests by replacing this method by XXfactory, which we couldn't do if I was using just the barebone line of code. I agree this would have made more sense with more than 2 tests ^^'.

So those were my `why`. They're not necessarily good/accurate ^^'.

So let me know your thoughts, I can remove the factory and the tests if you'd like, no problem.

mrmustard/utils/misc_tools.py
sduquemesa: from the [pennylane standard](https://docs.pennylane.ai/en/stable/development/guide/documentation.html#functions-and-methods) it is customary to add the types to the docstrings (even if type hints are present in the function)

```python
def func(arg1, arg2, arg3="default_value", **kwargs):
    """Single sentence that summarizes the function.

    Multi-line description of the function (optional, only if necessary).

    Args:
        arg1 (type): Description.
            Continuation line is indented if needed.
        arg2 (type): description
        arg3 (str): Description. Do not provide the default
            value, as this is already present in the signature.

    Keyword Args:
        kwarg1 (type): Description. Provide the default value if applicable.

    Returns:
        type: Description.
        Continuation line is not indented.

    Raises:
        ExceptionType: description

    .. seealso:: :func:`~.relevant_func`, :class:`~.RelevantClass` (optional)

    **Example**

    Minimal example with 1 or 2 code blocks (required).

    .. details::
        :title: Usage Details

        More complicated use cases, options, and larger code blocks (optional).

    **Related tutorials**

    Links to any relevant PennyLane demos (optional).
    """
```

tests/test_utils/test_misc_tools.py
sduquemesa: Perhaps before deciding if we should keep it or remove it it'd be good to look at its usage pattern. Can you point me to some code that uses this factory?

sduquemesa (changes requested):

mrmustard/utils/misc_tools.py
ryk-wolf: allright!

ryk-wolf: I've added the types hints, except for args/kwargs because I wasn't sure what's the best typing there.

tests/test_utils/test_misc_tools.py
ryk-wolf: For now I only use it in the tests.
if you go to the stage1_MVP_data branch, you'll find it in the fixture called DATA in all files, you can also find it inside some of the tests when I need to build a specific instance of the desired type on the fly, eg: in `test_matvec_data.py` the test function `if_coeffs_not_given_they_are_equal_to_1` calls it. There are a few other instances such as this latter one.

sduquemesa (approved):
Thanks for your changes, great work! 🚀

PR #257, opened 2023-07-12
Migration to Poetry

ggulli:
**Context:**

**Benefits:**

**Possible Drawbacks:**

**Related GitHub Issues:**


ziofil:
Fantastic, thank you Gabriele!

ziofil (approved):
👍

.devcontainer/Dockerfile
```diff
@@ -36,6 +36,8 @@ RUN sh -c "$(wget -nv -O- https://raw.githubusercontent.com/ohmyzsh/ohmyzsh/mast
 
 # upgrade pip and install package manager
 RUN python -m pip install --no-cache-dir --upgrade pip
+RUN pip install --no-cache-dir poetry==1.4.0
```
zeyueN, line 39: Any reason for version 1.4?
Also we can perhaps use [the `POETRY_VERSION` environ variable](https://python-poetry.org/docs/) if we decide to upgrade in the future.

zeyueN (approved):
🥇 Looking great! Thanks @ggulli

.devcontainer/Dockerfile
ggulli: I pinned this version to be consistent with the other projects we have. We were using this version elsewhere. 

If we introduce that variable, I'm worried we will end up adding extra complexity. Especially to make everything consistent between Dockerfiles and CI workflows.

PR #271, opened 2023-08-21
fix the variable name in apply choi

sylviemonet:
**Context:** The input variable in the function apply_choi_to_dm is not correct

**Description of the Change:** from choi_in_idx to choi_in_modes

**Benefits:**

**Possible Drawbacks:**

**Related GitHub Issues:**

### DIFF
--- a/.github/CHANGELOG.md
+++ b/.github/CHANGELOG.md
@@ -8,7 +8,7 @@
 
 ### Bug fixes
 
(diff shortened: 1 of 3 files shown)


ziofil (approved):
thanks :)

ziofil:
could you add a test that makes use of the arg names, and also change idx -> mode in the other methods?

sylviemonet:
> could you add a test that makes use of the arg names, and also change idx -> mode in the other methods?

@ziofil Hi Filippo, before merging it, please have a last look please!

PR #273, opened 2023-09-25
Ensuring that hbar can only be specified in `Settings`

SamFerracin:
### Before submitting

      test directory!

      `pip install black`.

**Context:**
While the value of `hbar` is specified in `Settings`, some methods allow passing (potentially different) values of `hbar`. There is no mechanism to keep track of the different values of `hbar` employed in a simulation, nor to ensure consistency. This leaves room for unintended behaviors.
 
All methods retrieve the value of `hbar` from the `Settings`, instead of allowing users to specify it in the inputs.

**Benefits:**
The value of `hbar` is guaranteed to be consistent throughout the session, meaning that at any point, it is equal to the value that can be found in the `Settings`.

**Possible Drawbacks:**
Users may still change the `Settings` (and specifically, the value of `hbar`) at any point during the computation. This will be disallowed in a follow-up Pull Request.

**Related GitHub Issues:**

### DIFF
--- a/mrmustard/lab/states.py
+++ b/mrmustard/lab/states.py
@@ -45,8 +45,8 @@ class Vacuum(State):
     r"""The N-mode vacuum state."""
 
     def __init__(self, num_modes: int):
-        cov = gaussian.vacuum_cov(num_modes, settings.HBAR)
-        means = gaussian.vacuum_means(num_modes, settings.HBAR)
+        cov = gaussian.vacuum_cov(num_modes)
+        means = gaussian.vacuum_means(num_modes)
         State.__init__(self, cov=cov, means=means)
 
 
@@ -109,13 +109,13 @@ class Coherent(Parametrized, State):
         self._modes = modes
         self._normalize = normalize
 
-        means = gaussian.displacement(self.x.value, self.y.value, settings.HBAR)
-        cov = gaussian.vacuum_cov(means.shape[-1] // 2, settings.HBAR)
+        means = gaussian.displacement(self.x.value, self.y.value)
+        cov = gaussian.vacuum_cov(means.shape[-1] // 2)
         State.__init__(self, cov=cov, means=means, cutoffs=cutoffs, modes=modes)
 
     @property
     def means(self):
-        return gaussian.displacement(self.x.value, self.y.value, settings.HBAR)
+        return gaussian.displacement(self.x.value, self.y.value)
 
 
 class SqueezedVacuum(Parametrized, State):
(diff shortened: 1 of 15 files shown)


mrmustard/utils/wigner.py
```diff
@@ -30,13 +29,18 @@ def wigner_discretized(rho, qvec, pvec, hbar=settings.HBAR):
...
     Retunrs:
         tuple(array, array, array): array containing the discretized Wigner function, and the Q and
             P coordinates (in meshgrid form) in which the function is calculated
     """
+    hbar = settings.HBAR
+    return _wigner_discretized(rho, qvec, pvec, hbar)
+
 
+@njit
+def _wigner_discretized(rho, qvec, pvec, hbar):
```
SamFerracin, line 42: @sylviemonet just a comment about this:

The best thing would be to have 
`@njit`
`def wigner_discretized(rho, qvec, pvec)`:
`    hbar = settings.HBAR`
`    ...`
But unfortunately, `@njit` dislikes `settings.HBAR`. 

This is why I decided to go through the `_wigner_discretized` function, which does take `hbar` as input, but is hidden and won't be used anywhere else. If you have better ideas, ket me know :)

tests/test_physics/test_fock/test_fock.py
```diff
@@ -37,6 +37,10 @@
 from mrmustard.math.lattice.strategies import displacement, grad_displacement
 from mrmustard.physics import fock
 
+from mrmustard import settings
+
+settings.HBAR = 2
```
ziofil, line 42: why do we need this here?

tests/test_physics/test_gaussian/test_symplectics.py
```diff
@@ -34,6 +40,10 @@
 from mrmustard.lab.states import TMSV, Thermal, Vacuum
 from mrmustard.physics.gaussian import controlled_X, controlled_Z
 
+from mrmustard import settings
+
+settings.HBAR = 2
```
ziofil, line 45: I guess it's for the same reason, but why do we need to set hbar here?

ziofil (approved):
Great job! I only left a couple comments

tests/test_physics/test_gaussian/test_symplectics.py
SamFerracin: Thanks for spotting this, I had forgotten it there

mrmustard/lab/states.py
```diff
@@ -421,7 +426,10 @@ def __init__(
         eigenvalues: Vector = None,
         symplectic_trainable: bool = False,
         eigenvalues_trainable: bool = False,
-        eigenvalues_bounds: Tuple[Optional[float], Optional[float]] = (settings.HBAR / 2, None),
+        eigenvalues_bounds: Tuple[Optional[float], Optional[float]] = (
+            settings.HBAR / 2,
```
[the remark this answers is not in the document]
SamFerracin: I was a bit confused here. Should the lower bound always be hbar/2, or can I replace it with `None`?

PR #274, opened 2023-09-28
Vanilla strategy with complex512 (using Julia)

rdprins:
Calculating Fock representations using the "vanilla strategy" is now more numerically stable (i.e. numerical blowups 
that result from repeatedly applying the recurrence relation are now postponed to higher cutoff values).
This is done by representing Fock amplitudes with precision complex512 rather than complex128 
(which counters the accumulation of floating-point errors). 
We run Julia code via PyJulia (where Numba was used before) to keep the code fast.

### Before submitting

- [x] All new features must include a unit test.
      test directory!

- [x] All new functions and code must be clearly commented and documented.

- [x] Ensure that the test suite passes, by running `make test`.

- [x] Ensure that code and tests are properly formatted, by running `make format` or `black -l 100
      `pip install black`.

- [x] Add a new entry to the `.github/CHANGELOG.md` file, summarizing the

- [x] The Mr Mustard source code conforms to

**Context:** Vanilla strategy for calculating Fock representations

**Description of the Change:** code is now run in Julia via PyJulia + we represent Fock amplitudes with 512 bits of precision

**Benefits:** numerical stability

**Possible Drawbacks:** slows down code

**Related GitHub Issues:**


ziofil:
oh sorry, I accidentally hit the "ready for review" button

mrmustard/math/tensorflow.py
```diff
@@ -378,7 +389,7 @@ def hermite_renormalized(
             The renormalized Hermite polynomial of given shape.
         """
         _A, _B, _C = self.asnumpy(A), self.asnumpy(B), self.asnumpy(C)
-        G = strategies.vanilla(tuple(shape), _A, _B, _C)
+        G = Main_julia.vanilla(tuple(shape), _A, _B, _C.item())
```
ziofil, line 392: Do we want every call to `hermite_renormalized` run in julia? I would think sometimes we still want the old numba code to run, where there's no need for the extra precision

mrmustard/math/__init__.py
```diff
@@ -53,8 +53,8 @@ class Math:
     def __getattribute__(self, name):
         if settings.BACKEND == "tensorflow":
             return object.__getattribute__(TFMath(), name)
-        elif settings.BACKEND == "torch":
-            return object.__getattribute__(TorchMath(), name)
+        # elif settings.BACKEND == "torch":
+        #     return object.__getattribute__(TorchMath(), name)
```
ziofil, line 57: yeah I don't know why we still leave that there... 😅

rdprins: Somehow, I got errors about it when testing the PyJulia cod. I was planning to uncomment it again later.

mrmustard/math/tensorflow.py
rdprins: I'll add a keyword argument `precision`. And if precision==float64, then I redirect to the Numba code.

mrmustard/math/__init__.py
```diff
@@ -34,15 +34,19 @@
...
 """
-
-
 import importlib
+from julia.api import Julia
 from mrmustard import settings
 
 if importlib.util.find_spec("tensorflow"):
-    from mrmustard.math.tensorflow import TFMath
+    from mrmustard.math.tensorflow_wrapper import TFMath
```
SamFerracin, line 42: Why this changes?
Just curious if it's based on personal preference, or if there's some conflict with leaving the original names :)

mrmustard/math/__init__.py
```diff
@@ -34,15 +34,19 @@
...
 if importlib.util.find_spec("torch"):
-    from mrmustard.math.torch import TorchMath
+    from mrmustard.math.torch_wrapper import TorchMath
+
+# initialize Julia
+jl = Julia(
+    compiled_modules=False
+)  # must be run before "from julia import Main as Main_julia", which is done in math/tensorflow.py
```
SamFerracin, line 49: math/tensorflow.py --> math/tensorflow_wrapper.py

mrmustard/settings.py
```diff
@@ -100,6 +100,9 @@ def __init__(self):
         self._seed = np.random.randint(0, 2**31 - 1)
         self.rng = np.random.default_rng(self._seed)
         self._default_bs_method = "vanilla"  # can be 'vanilla' or 'schwinger'
+        self.precision_bits_hermite_poly = (
```
SamFerracin, line 103: `self.preci..` --> `self._preci..`, and then add a setter and a getter, so that it appears in the docs

mrmustard/math/__init__.py
```diff
@@ -34,15 +34,19 @@
...
 
 if importlib.util.find_spec("tensorflow"):
-    from mrmustard.math.tensorflow import TFMath
+    from mrmustard.math.tensorflow_wrapper import TFMath
 if importlib.util.find_spec("torch"):
-    from mrmustard.math.torch import TorchMath
+    from mrmustard.math.torch_wrapper import TorchMath
+
+# initialize Julia
+jl = Julia(
```
SamFerracin, line 47: Are this and `from julia import Main as Main_julia` fast, or do they contribute significantly to the import time?

mrmustard/math/__init__.py
TannerRogalsky: Something about the way that Ray spins up it's new process environment and then performs imports was conflating this module with the tensorflow package. There is probably a way of fixing this that wouldn't require this rename but I couldn't figure it out after a couple hours. Here's the stackoverflow post that seemed best to indicate what the issue is: https://stackoverflow.com/questions/64509832/attributeerror-partially-initialized-module-tensorflow-has-no-attribute-conf

mrmustard/math/__init__.py
rdprins: It seems that `jl = Julia(compiled_modules=False)` is rather slow. Should I for example make a settings entry `init_julia` and only run this line when `init_julia == True`?

image

SamFerracin: Or you could import it in the setter for the setting that you added..?
`If precision is ..., import julia`

mrmustard/settings.py
```diff
@@ -269,6 +270,18 @@ def SEED(self, value):
...
+    def PRECISION_BITS_HERMITE_POLY(self):
+        r"""
+        The number of bits used to represent a single Fock amplitude when calculating Hermite polynomials.
+        Default is 128 (i.e. the Fock representation has dtype complex128).
+        """
+        return self._precision_bits_hermite_poly
+
+    @PRECISION_BITS_HERMITE_POLY.setter
+    def PRECISION_BITS_HERMITE_POLY(self, value: str):
+        self._precision_bits_hermite_poly = value
```
SamFerracin, line 283: Thanks :) 
One final suggestion: Could you check that the value is valid, and error if it's not?

rdprins: At the moment it's checked in `mrmustard.math.tensorflow_wrapper.hermite_renormalized`. I think it might be better not to define a list of allowed values in multiple places (as the list will get longer in my next PR). What do you think?

SamFerracin: I agree!
Slight preference for moving the check here (as a user, I would rather knowing that I made a mistake right away, as opposed to running some code and then getting an error), but don't waste too much time. The important thing is that there is a check in place

ziofil: I would move it here too, if anything it would be run just once as opposed at every call of `hermite_renormalized`

rdprins:
The line `from julia import (Main as Main_julia,)` pops up in CodeFactor as it's run outside of toplevel (both in `settings.py` as in `tensorflow_wrapper.py`), but I think it can't be moved, because it has to be run after running `jl = Julia(compiled_modules=False)`, which is done in `settings.py`.

ziofil (approved):

tests/test_math/test_lattice.py
```diff
@@ -17,10 +17,23 @@
 import numpy as np
 
 from mrmustard.lab import Gaussian
+from mrmustard import settings
 
 
-def test_vanilla_vs_binomial():
-    """Test that the vanilla and binomial methods give the same result"""
+def test_vanillaNumba_vs_binomial():
```
SamFerracin, line 23: `@pytest.mark.parametrize("precision", ([128, 512]))`

tests/test_math/test_lattice.py
```diff
@@ -17,10 +17,23 @@
...
+    """Test that the vanilla method (Numba with precision of complex128) and the binomial method give the same result"""
+    settings.PRECISION_BITS_HERMITE_POLY = 128
+    G = Gaussian(2)
+
+    ket_vanilla = G.ket(cutoffs=[10, 10])[:5, :5]
+    ket_binomial = G.ket(max_photons=10)[:5, :5]
+
+    assert np.allclose(ket_vanilla, ket_binomial)
+
```
SamFerracin, line 32: Set precision back: 

At the beginning of the file (or at the beginning of the func), add:
`p0 = settings.PRECISION_BITS_HERMITE_POLY`

At the beginning of the function:
` settings.PRECISION_BITS_HERMITE_POLY = precision`

At the end of the function, add:
` settings.PRECISION_BITS_HERMITE_POLY = p0`

PR #278, opened 2023-09-29
Freezing settings after their values are used

SamFerracin:
**Context:** 
Some of the attributes of `Settings` (e.g., `settings.HBAR`) should become immutable after they are used for the first time, while others (such as the cutoffs) can change at any point during the session.

A new class `ImmutableSetting` is introduced. This class takes care of "freezing" the value of certain settings after they are queried for the first time.
Additionally, the `Settings` class is restructured so that every attribute appears in the docs and is properly documented.

**Remarks:**
We should have discussions about:
- What attributes should become immutable.
- If attributes should that are not currently used, e.g. `settings.CHOI_R`, should be removed.

### DIFF
--- a/tests/random.py
+++ b/tests/random.py
@@ -112,7 +112,7 @@ def force_settings(name, value):
         >>> settings.HBAR
         1.0
     """
-    getattr(settings, name)._value = value   # pylint: disable=protected-access
+    getattr(settings, name)._value = value  # pylint: disable=protected-access
 
 
 # gates
--- a/tests/random.py
+++ b/tests/random.py
@@ -112,7 +112,7 @@ def force_settings(name, value):
         >>> settings.HBAR
         1.0
     """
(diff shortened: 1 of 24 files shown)


tests/random.py
```diff
@@ -97,6 +97,24 @@ def bounds_check(t):
 prob_bounds = st.tuples(none_or_(prob), none_or_(prob)).filter(bounds_check)
 
 
+# settings
+def force_settings(name, value):
```
ziofil, line 101: Aha, now I see what you meant. What if we change the hbar attribute of settings just for the tests? Or the ImmutableSetting class has a test_mode: bool that allows for mutability?

ziofil (approved):
I agree that the `force_settings` isn't pretty and kind of goes against what the `ImmutableSetting` is for. I believe there is a better way, but let's move on for now. We can always revisit in the future.

PR #280, opened 2023-10-05
Basic tools to simulate circuits with tensor networks

SamFerracin:
**Context:**
Developing tools to simulate circuits with tensor networks

Added core classes and methods, and namely:
- `Wire` and `Tensor`
- Functions to connect tensors and contract networks
Incorporated these classes into `Transformation`.

**Benefits:**
We can now connect unitaries and channels in a network and obtain the resulting tensor via contraction

**What remains:**
In this PR:
- Writing more tests
- Improving documentation
In future PRs:
- Incorporating tensors into `State` and measurements


mrmustard/lab/abstract/transformation.py
```diff
@@ -195,6 +132,53 @@ def d_vector_dual(self) -> Optional[RealVector]:
             return -d
         return -math.matmul(Xdual, d)
 
+    def bargmann(self, numpy=False):
+        X, Y, d = self.XYd(allow_none=False)
+        if self.is_unitary:
+            A, B, C = bargmann.wigner_to_bargmann_U(
+                X if X is not None else math.identity(d.shape[-1], dtype=d.dtype),
```
ziofil, line 139: when we call `self.XYd(allow_none=False)` the `allow_none=False` option already creates the default `X`,`Y`, and `d`, so there's no need here to have `if X is not None else ...` and same below

mrmustard/lab/abstract/transformation.py
```diff
@@ -195,6 +132,53 @@ def d_vector_dual(self) -> Optional[RealVector]:
...
+
+        X, Y, d = self.XYd(allow_none=False)
+        choi = fock.wigner_to_fock_Choi(X, Y, d, shape=shape)
+        if dual:
+            n = len(shape) // 4
+            N0 = list(range(0, n))
+            N1 = list(range(n, 2 * n))
+            N2 = list(range(2 * n, 3 * n))
+            N3 = list(range(3 * n, 4 * n))
+            choi = math.conj(math.transpose(choi, N1 + N0 + N3 + N2))  # if dual we flip out-in
```
ziofil, line 179: we will think if `dual` here is a repetition of the `DualView`

mrmustard/lab/abstract/transformation.py
```diff
@@ -383,3 +312,96 @@ def _repr_markdown_(self):
...
         return header + body
+
+
+class Unitary(Transformation):
+    def __init__(self, name: str, modes_in: list[int], modes_out: list[int]):
+        super().__init__(name=name, modes_in_ket=modes_in, modes_out_ket=modes_out)
+        self.is_unitary = True
+
+    def value(self, cutoff: int):
+        return self.U(cutoffs=[cutoff for _ in range(self.num_modes)])
```
ziofil, line 323: Apologies if this was here before. I don't think the value should force changing to fock representation)

mrmustard/lab/abstract/transformation.py
```diff
@@ -383,3 +312,96 @@ def _repr_markdown_(self):
...
+    def _transform_fock(self, state: State, dual=False) -> State:
+        op_idx = [state.modes.index(m) for m in self.modes]
+        U = self.U(cutoffs=[state.cutoffs[i] for i in op_idx])
+        if state.is_hilbert_vector:
+            return State(ket=fock.apply_kraus_to_ket(U, state.ket(), op_idx), modes=state.modes)
+        return State(dm=fock.apply_kraus_to_dm(U, state.dm(), op_idx), modes=state.modes)
+
+    def U(
+        self,
+        cutoffs: Optional[Sequence[int]] = None,
```
ziofil, line 334: we should add a shape argument that overrides the cutoffs if specified

mrmustard/math/tensor_networks/tensors.py
```diff
@@ -0,0 +1,328 @@
...
+
+
+__all__ = ["Wire", "Tensor"]
+
+
+def random_int() -> int:
+    r"""
+    Returns a random integer obtained from a UUID
+    """
+    return uuid.uuid1().int
```
ziofil, line 34: maybe uuid4 is better, but yes this is better than what I had done 😅

mrmustard/math/tensor_networks/tensors.py
```diff
@@ -0,0 +1,328 @@
...
+        self._original = tensor
+        super().__init__(
+            self._original.name,
+            self._original.input.bra.keys(),
+            self._original.output.bra.keys(),
+            self._original.input.ket.keys(),
+            self._original.output.ket.keys(),
+        )
+
+    def value(self, cutoff):
```
ziofil, line 327: should this be the conjugate?

mrmustard/math/tensor_networks/tensors.py
```diff
@@ -0,0 +1,328 @@
...
+    Args:
+        name: The name of this tensor.
+        modes_in_ket: The input modes on the ket side.
+        modes_out_ket: The output modes on the ket side.
+        modes_in_bra: The input modes on the bra side.
+        modes_out_bra: The output modes on the bra side.
+    """
+    _repr_markdown_ = None  # otherwise it takes over the repr due to mro
+
+    def __init__(
```
ziofil, line 122: should `ABC`s have an init? 🤔

mrmustard/math/tensor_networks/tensors.py
SamFerracin: About this.. If I use `uuid4` or `np.random.randint` or similar functions, the TN tests fail. If I use `uuid1`, they pass. I wonder if this only happens on my machine.. :face_in_clouds:

mrmustard/math/tensor_networks/tensors.py
SamFerracin: I never really added it, so I'd lean towards saying no. Can't find a final answer on internet to this though, although there's no super().init used in any examples provided by the official python docs:
https://docs.python.org/3/library/abc.html

mrmustard/lab/abstract/transformation.py
SamFerracin: Yeah I forgot to change this. Thanks for spotting it!

mrmustard/math/tensor_networks/tensors.py
ziofil: oh, that's interesting 🧐

PR #281, opened 2023-10-06
bumps tensorflow version to v2.14

zeyueN:
**Context:**

This PR manages to
- bump tensorflow to its latest version v2.14, on all platforms (no longer v2.12 as the branch name is suggesting)
- add native M1 Mac support! (outside of devcontainers)
- improve poetry installation without the need to reinstall tensorflow with pip

- pyproject.toml updated with better TF dependency handling (documented in comments)
- updated lock file
- TF API change: Adam -> legacy.Adam
- TF API change: concat of scalars uses stack now

**Benefits:**
NATIVE, MAC, SUPPORT

**Possible Drawbacks:**
pyproject.toml got more complex?

**Related GitHub Issues:**


mrmustard/math/tensorflow.py
```diff
@@ -341,9 +344,9 @@ def MultivariateNormalTriL(self, loc: Tensor, scale_tril: Tensor):
 
     # TODO: is a wrapper class better?
     @staticmethod
-    def DefaultEuclideanOptimizer() -> tf.keras.optimizers.Optimizer:
+    def DefaultEuclideanOptimizer() -> tf.keras.optimizers.legacy.Optimizer:
         r"""Default optimizer for the Euclidean parameters."""
-        return tf.keras.optimizers.Adam(learning_rate=0.001)
+        return tf.keras.optimizers.legacy.Adam(learning_rate=0.001)
```
heltluke, line 349: This is an easy fix, until we want to move to the new optimizer.

mrmustard/math/tensorflow.py
```diff
@@ -81,7 +81,10 @@ def clip(self, array, a_min, a_max) -> tf.Tensor:
         return tf.clip_by_value(array, a_min, a_max)
 
     def concat(self, values: Sequence[tf.Tensor], axis: int) -> tf.Tensor:
-        return tf.concat(values, axis)
+        try:
+            return tf.concat(values, axis)
+        except tf.errors.InvalidArgumentError:
+            return tf.stack(values, axis)
```
heltluke, line 87: This is obviously a hack. Either we are safe to replace all calls to `tf.concat` here with `tf.stack`, or we should likely be better about determining which to use based on the passed `values` rather than waiting for the `InvalidArgumentError`.

zeyueN: ok, basing it on 0-rank seems to have fixed it.

zeyueN:
I think this is ready for review. We got a lot further than expected.

The tests all passed on our M1 Mac mini natively with poetry installation in pyenv 🥲 
And it's actually very fast, with all tests done in 2:30...

image

heltluke (approved):

ggulli (approved):
Great work solving those Tensorflow dependencies! 🚀

mrmustard/math/tensorflow.py
```diff
@@ -81,7 +81,10 @@ def clip(self, array, a_min, a_max) -> tf.Tensor:
         return tf.clip_by_value(array, a_min, a_max)
 
     def concat(self, values: Sequence[tf.Tensor], axis: int) -> tf.Tensor:
-        return tf.concat(values, axis)
+        if any(tf.rank(v) == 0 for v in values):
```
ziofil, line 84: awesome

ziofil (approved):
outstanding 💯

PR #283, opened 2023-10-11
Fixing a bug in `choi`

SamFerracin:
**Context:**
Calling `choi` for a unitary transformation errors.

A few changes in the way the shape is handled. Also added tests for this.


ziofil (approved):
Thanks!

PR #287, opened 2023-10-16
Develop tn

SamFerracin:
**Context:**
Adding functions to generate, contract, and draw tensor networks.

- New classes (Wire, Tensor) and functions (connect, contract, draw)
- Incorporated Tensor class in Transformation


mrmustard/math/tensor_networks/tensors.py
```diff
@@ -0,0 +1,445 @@
...
+        """
+        for wire in self.wires:
+            if wire.is_connected:
+                msg = (
+                    "Cannot change nodes in a tensor when some of its wires are already connected."
+                )
+                raise ValueError(msg)
+        self._update_modes(modes_in_ket, modes_out_ket, modes_in_bra, modes_out_bra)
+
+    def shape(self, default_dim: Optional[int] = None, swap=False):
```
ziofil, line 353: instead of swap it would be clearer to use `out_in` perhaps?

ziofil (approved):
great job

PR #288, opened 2023-10-17
Diagonal strategy with higher precision (using Julia)

rdprins:
**Context:**
Calculating Fock representations using the "diagonal strategy" and "1 leftover mode strategy" is now more numerically stable (i.e. numerical blowups that result from repeatedly applying the recurrence relation are postponed to higher cutoff values).

This is done by representing Fock amplitudes with a precision that is higher than complex128 (which counters the accumulation of floating-point errors). We run Julia code via PyJulia (where Numba was used before) to keep the code fast.

The Julia code currently allows for the following precisions: Complex128, Complex256, Complex384 and Complex512. Other values could easily be added. The vanilla code will allow for all of these precisions as well, where only Complex512 was available in the previous PR.

**Benefits:** numerical stability

**Possible Drawbacks:** slows down code

**Related GitHub Issues:**


tests/test_math/test_compactFock.py
```diff
@@ -38,54 +43,63 @@ def random_ABC(draw, M):
...
-            inds_expanded = list(inds) + list(inds)  # a,b,c,a,b,c
-            ref_diag[inds] = G_ref[tuple(inds_expanded)]
-
-        # New MM
-        G_diag = math.hermite_renormalized_diagonal(
-            math.conj(-A), math.conj(B), math.conj(G0), cutoffs
-        )
-        assert np.allclose(ref_diag, G_diag)
-
+    for precision in allowed_values_precision:
```
rdprins, line 46: How do I use` @hypothesis.given` and `@pytest.mark.parametrize` at the same time?

tests/test_math/test_compactFock.py
```diff
@@ -14,6 +16,9 @@
 
 math = Math()  # use methods in math if you want them to be differentiable
 
+original_precision = settings.PRECISION_BITS_HERMITE_POLY
+allowed_values_precision = [128, 256, 384, 512]
```
rdprins, line 20: Do I add this list as an immutable settings entry?

tests/test_math/test_compactFock.py
rdprins: I found some solutions on stackoverflow, but I think they are less intuitive than what I currently have.

mrmustard/settings.py
```diff
@@ -272,6 +274,16 @@ def SEED(self, value):
...
+        return self._julia_initialized
+
+    @property
+    def ALLOWED_PRECISION_BITS_HERMITE_POLY(self):
+        r"""List of possible values for settings.PRECISION_BITS_HERMITE_POLY."""
+        return self._allowed_precision_bits_hermite_poly
+
     @property
     def PRECISION_BITS_HERMITE_POLY(self):
         r"""
```
SamFerracin, line 289: You can add here the list of allowed values, so that users know it

ziofil (approved):
Fantastic!

PR #289, opened 2023-10-17
Removing circular dependencies

SamFerracin:
**Context:**
Mr Mustard contains some circular dependencies, meaning that some packages import files from each other. This can create headaches when debugging and refactoring, and will become a bigger issue when scaling the code up.  
with_circ_dep

In this PR, files are moved around in such a way that all circular dependencies are broken.
without_circ_dep

**Benefits:**
Being able to debug, refactor, and scale up more easily.


ziofil (approved):
A+!

PR #290, opened 2023-10-18
Update docs about installing Julia

rdprins:
**Context:** Julia dependencies (added in order to stabilize the recurrence relation: https://github.com/XanaduAI/MrMustard/pull/274)

**Description of the Change:** Added Manifest.toml and Project.toml which specify the julia version and dependencies. Updated the README and development guide to instruct the user to install julia and to instantiate the toml files. This means PyJulia now runs in the environment defined by the toml files, rather than in Julia's base environment. 

- When installing via devcontainer, julia (v1.9.3) is installed and the environment is instantiaded.
- When installing via MakeFile, the environment is only instantiaded when Julia was previously installed by the user.
- When installing via `poetry install` or `pip install`, the julia installation and environment instantiation need to be performed by the user.

**Benefits:**

**Possible Drawbacks:**

**Related GitHub Issues:**


rdprins:
Waiting to merge this PR. First need to be sure that it's necessary to ask the user to install the Julia packages.

README.md
```diff
@@ -31,6 +31,16 @@ Mr Mustard supports:
 - Plug-and-play backends (TensorFlow as default)
 - An abstraction layer `XPTensor` for seamless symplectic algebra (experimental)
 
+# Julia
+MrMustard requires [Julia](https://julialang.org/) to be installed as well as the following Julia packages:
```
ziofil, line 35:
```suggestion
MrMustard requires [Julia](https://julialang.org/) to be installed, as well as the following Julia packages:
```

README.md
```diff
@@ -31,6 +31,22 @@ Mr Mustard supports:
 - Plug-and-play backends (TensorFlow as default)
 - An abstraction layer `XPTensor` for seamless symplectic algebra (experimental)
 
+# Increased numerical stability using Julia [optional]
+
+Converting phase space objects to Fock space can be numerically unstable due to accumulating floating point errors.
+To resolve this, the conversion can be performed with extended-precision arithmetic. To use this feature,
+an installation of [Julia](https://github.com/JuliaLang/juliaup#installation) is required (version 1.9.3).
```
SamFerracin, line 38: Is 1.9.3 the only one that's gonna work? 
What do people need to do if they have >1.9.3? Downgrade?

rdprins: I would think everything keeps working when newer versions come out, but I haven't checked older versions. Should I just write >1.9.3? Or should I check which ones currently work?

rdprins: 1.9.3 is the current stable release. Should I already write >= or do we wait until newer versions come out?

rdprins: Great, thanks!

ziofil (approved):
great!

PR #299, opened 2023-10-30
Skipping Julia tests if Julia is not installed.

SamFerracin:
Skipping Julia tests if Julia is not installed.
These tests are still run by GitHub


rdprins (approved):
Great! The `allowed_precision_bits_hermite_poly` are better suited where you put them in settings. Thanks for parameterizing the tests with both `A_B_G0` and `precision` as well. I wasn't sure how to do that.

PR #306, opened 2023-11-21
Remove warnings

ziofil:
Removes complex to real warnings when making the mikkel_plot, due to casting in the oscillator_eigenstate function.
We just take the real part now.


ziofil:
superseded by #307

PR #307, opened 2023-11-22
Remove warnings at the root

SamFerracin:
Removes ``ComplexWarning`` at the root (inside the backend) rather than in the functions that require casting

### DIFF
--- a/mrmustard/math/backend_tensorflow.py
+++ b/mrmustard/math/backend_tensorflow.py
@@ -42,8 +42,8 @@ class BackendTensorflow(BackendBase):  # pragma: no cover
     """
 
     int32 = tf.int32
-    float64 = tf.float64
     float32 = tf.float32
+    float64 = tf.float64
     complex64 = tf.complex64
     complex128 = tf.complex128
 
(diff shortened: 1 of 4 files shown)


sduquemesa (approved):
Thanks @SamFerracin! 

I wonder if this PR is also a good opportunity to disable tensorflow's CUDA warning messages — those that appear when importing TF and no GPU is available. What do you think? One option is to just disable them and the other is to install `tensorflow-cpu` instead of the full `tensorflow` package that includes GPU support.

SamFerracin:
@sduquemesa That's a great idea. i hate those warnings. Let me look into it :)

mrmustard/math/backend_tensorflow.py
```diff
@@ -87,6 +87,9 @@ def boolean_mask(self, tensor: tf.Tensor, mask: tf.Tensor) -> Tensor:
     def cast(self, array: tf.Tensor, dtype=None) -> tf.Tensor:
         if dtype is None:
             return array
+
+        if dtype not in [self.complex64, self.complex128, "complex64", "complex128"]:
+            array = self.real(array)
```
ziofil, line 92: what if we cast to int32?

ziofil (approved):
approved, up to you to cover casting to int32 correctly

mrmustard/math/backend_tensorflow.py
SamFerracin: I only see warnings when I convert from complex to something else, and this case is covered.
Do you have a specific example in mind that is not covered?

PR #311, opened 2023-12-04
Cleaning up the backend manager

SamFerracin:
BackendManager is a singleton. As such, we can replace many of its `cls` to `self`


ziofil (approved):
👍

PR #317, opened 2023-12-09
generaldyne pdf

ziofil:
`physics.gaussian.general_dyne()` now returns the pdf for further use (e.g. sampling)

### DIFF
--- a/mrmustard/physics/gaussian.py
+++ b/mrmustard/physics/gaussian.py
@@ -553,7 +553,6 @@ def general_dyne(
     proj_cov: Matrix,
     proj_means: Optional[Vector] = None,
     modes: Optional[Sequence[int]] = None,
-    sample_shape: Optional[Sequence[int]] = (),
 ) -> Tuple[Scalar, Scalar, Matrix, Vector, Any]:
     r"""Returns the results of a general-dyne measurement. If ``proj_means`` are not provided
     (as ``None``), they are sampled from the probability distribution.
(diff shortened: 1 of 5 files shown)


mrmustard/physics/gaussian.py
```diff
@@ -574,34 +573,35 @@ def general_dyne(
             oucome probability [units of `\hbar**N`],
             post-measurement covariace [units of `2\hbar`]
             post-measurement means vector [units of `\sqrt{\hbar}`].
+            pdf (Tensorflow probability distribution)
```
SamFerracin, line 576: If the numpy backend is used, it is not a tensorflow distribution

ziofil: Yes it is. Here we are using TF Probability, not TF.

ziofil:
> This PR adds "pdf" as an output for the generaldyne measurement and fixes small bugs.
> 
> Just have a small question for @ziofil before approving it, I haven't found any corresponding tests in tests_gaussian for this, do we need add one?

I'm not sure what you want to test here

sylviemonet (approved):

mrmustard/physics/gaussian.py
SamFerracin: No, I don't think so. Look at the MultivariateNormalTriL function in numpy_backend.py. It does not return a TF object. Let's touch base on this later, before you resolve this convo

PR #320, opened 2023-12-19
Adding wires functionality to lab_dev

SamFerracin:
**Context:**
Adding class `Wires` to lab_dev. This is the first class needed for the state refactoring.
Also adding `jupyter_sphinx` to the doc's dependencies, so that we can run jupyter notebooks while building the docs.


mrmustard/lab_dev/wires.py
```diff
@@ -0,0 +1,193 @@
...
+            self._modes = list(modes)
+
+        keys = self._modes or set(modes_in_ket + modes_out_ket + modes_in_bra + modes_out_bra)
+        self._out_bra = {m: uuid.uuid4().int if m in modes_out_bra else None for m in keys}
+        self._in_bra = {m: uuid.uuid4().int if m in modes_in_bra else None for m in keys}
+        self._out_ket = {m: uuid.uuid4().int if m in modes_out_ket else None for m in keys}
+        self._in_ket = {m: uuid.uuid4().int if m in modes_in_ket else None for m in keys}
+
+    @property
+    def in_bra(self) -> dict[Mode, Optional[Wire]]:
```
ziofil, line 88: didn't we agree on reintroducing the `.output`, `.input`, `.ket` and `.bra` properties and return a new `Wires` object?

ziofil: that will save you from having to define all the combinations as special methods (`modes_out_bra`, `modes_in_bra` etc, as well as `out_bra`, `out_ket` etc)

mrmustard/utils/typing.py
```diff
@@ -49,6 +50,8 @@
 Z = TypeVar("Z", np.int16, np.int32, np.int64)
 N = TypeVar("N", np.uint16, np.uint32, np.uint64)
 
+Mode = int
```
ziofil, line 53: I would do the same, but consider that someone may be fooled into thinking that `Mode` is some class that we wrote

tests/test_lab_dev/test_wires.py
```diff
@@ -0,0 +1,155 @@
...
+        assert None not in list(wires.out_ket.values())
+
+        assert list(wires.in_ket.keys()) == modes
+        assert list(wires.in_ket.values()) == [None] * len(modes)
+
+    def test_init_ambiguous(self):
+        r"""
+        Tests the init of ``Wires`` with ambiguous modes.
+        """
+        wires = Wires([0, 2], [9, 3], [4, 5])
```
ziofil, line 54: I think anything should be allowed at the `Wires` level

mrmustard/lab_dev/wires.py
ziofil: that will also allow you to get more general subsets (e.g. all the output wires regardless of ket/bra/mode)

SamFerracin: Last time we talked about that, we agreed on introducing only what we strictly need. 

I need to be able to **read** the dictionaries `_in_bra`, `_out_bra` etc., or the modes that go in and out, which is what these properties allow you to do. I don't need to generate new wires every time

SamFerracin: P.S. By "I don't need to generate new wires every time" I mean that I would still need these properties, in order to be able to read the dictionaries and the modes

mrmustard/utils/typing.py
SamFerracin: True.. Would you just leave `int`? 
I feel like it's maybe too generic. I find the type-hints with `int` very confusing, which is why I added `Mode`. Open to change it though

tests/test_lab_dev/test_wires.py
SamFerracin: What do you mean?

ziofil (approved):
this looks a-ok for me

ziofil:
Context for closing the PR: we added this code to the prototype branch instead of develop.

PR #325, opened 2024-01-18
Changing ordering convention of `ids` in wires

SamFerracin:
**Context:**
So that it matches the indices convention of representation

`wires.ids` returns:
`[out_bra(0), ... out_bra(N), in_bra(0), ..., in_bra(N), ...]`
With this change, it will return:
`[out_bra(m), in_bra(m), out_ket(m), in_ket(m)] for m in modes`

### DIFF
--- a/mrmustard/lab/circuit_drawer.py
+++ b/mrmustard/lab/circuit_drawer.py
@@ -61,8 +61,7 @@ def _add_op(op, layer_str, decimals):
     control = []
     if op.__class__.__qualname__ in ["BSgate", "MZgate", "CZgate", "CXgate"]:
         control = [op.modes[0]]
-    # label = op.short_name
-    label = op.__class__.__qualname__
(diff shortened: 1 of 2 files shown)


mrmustard/lab/circuit_drawer.py
```diff
@@ -61,8 +61,7 @@ def _add_op(op, layer_str, decimals):
     control = []
     if op.__class__.__qualname__ in ["BSgate", "MZgate", "CZgate", "CXgate"]:
         control = [op.modes[0]]
-    # label = op.short_name
```
SamFerracin, line 64: This undoes a change that I had done accidentally, fixing the circuit_drawer tests in prototype-tn

ziofil:
let's have a discussion about this

SamFerracin:
We decided to change the indices convention on the representations rather than on the wires

PR #329, opened 2024-01-26
Refactor devcontainer 📦

sduquemesa:
**Context:**
Doing some housekeeping here on the devcontainer side.

- The previous build script was a bit cumbersome, here it has been simplified.
- Julia installation is now manually managed with `juliacall`, meaning there is no need for passing the optional `JULIA_INSTALL` arg to `docker build`.
- Only a single Dockerfile to create the container. This Dockerfile can also be used to create a completely isolated environment using docker (useful for testing, running simulations, etc.).
- Ensures changes on the code are reflected on the interpreter as they happen. To do so the workspace folder points to /mrmustard where the source code resides.
- Installs the jupyter extension automatically.
- **LOCAL DEV CONTAINERS ONLY**: Users can now have the `MRMUSTARD_PYTHON_VERSION` env-var set in their local environment to dictate which python version the dev container will be built with. When unset, the default value is 3.10. I'd suggest doing this in your `bashrc/zshrc/whatever-rc` to ensure it's always set and therefore picked up by VS Code

**Benefits:**
- Click and go 🚀
- Seamless julia installation
- (local-only) customizable python version

**Possible Drawbacks:**
None. I tried getting the `MRMUSTARD_PYTHON_VERSION` supported in GitHub Codespaces, but it doesn't seem to take any custom build args 😢


.devcontainer/devcontainer.json
```diff
@@ -1,42 +1,35 @@
-// For format details, see https://aka.ms/devcontainer.json. For config options, see the
-// README at: https://github.com/devcontainers/templates/tree/main/src/anaconda
 {
-	"name": "Mr Mustard",
+	"name": "MrMustard 🌭",
```
ziofil, line 2: haha love it

env.Dockerfile
```diff
@@ -0,0 +1,40 @@
+# *** Base *** #
+FROM python:3.10 AS base
```
SamFerracin, line 2: Were we always using 3.10 for docker, or 3.9?

sduquemesa: We had 3.9 to match the lowest python version supported by MrMustard but @ziofil uses 3.10 more often inside devcontainers

sduquemesa:
@ziofil, I've readded oh-my-zsh to the container as default terminal

env.Dockerfile
ziofil: If it were for me I'd use 3.12! Maybe now 3.11 is supported too

timmysilv: I have updated the devcontainer definition so it will build with 3.10 _by default_, but you can specify another version if you please. To do this, just have the `MRMUSTARD_PYTHON_VERSION` environment variable set locally before building+opening the devcontainer (default will be 3.10).

Some notes:
- I feel like VS Code always refreshes something before building, so I'd put it in your `~/.bashrc`, ~/.zshrc` or whatever you use to be safe
- We can't build with 3.12 because tensorflow fails to install 🙃

PR #331, opened 2024-01-29
Tests and docs for Wires

SamFerracin:
Adds tests and docs for wires in the format agreed during the dev meeting


SamFerracin:
Carries extra things

PR #337, opened 2024-02-02
Fixing a Scipy error

SamFerracin:
**Context:**
In the latest MrM release, `scipy` was unpinned.
Syncing the `develop` branch with this latest release causes pytest errors due to improper type handling on `scipy`'s side.

Adding a `dtype` parameter to `math.sqrtm` so that it never returns arrays of an unsupported type (specifically, `complex256`)


mrmustard/math/backend_tensorflow.py
```diff
@@ -370,11 +370,16 @@ def eigvals(tensor: tf.Tensor) -> Tensor:
     def xlogy(x: tf.Tensor, y: tf.Tensor) -> Tensor:
         return tf.math.xlogy(x, y)
 
-    def sqrtm(self, tensor: tf.Tensor, rtol=1e-05, atol=1e-08) -> Tensor:
+    def sqrtm(self, tensor: tf.Tensor, dtype, rtol=1e-05, atol=1e-08) -> Tensor:
```
ziofil, line 373: just wondering if the atol and rtol are okay? they seem a bit large (but maybe I'm wrong)

SamFerracin: Good question.. 
They were set before the backend refactoring, so I never thought too much about it.

If we lower them and the tests still pass, are we "happy"?

trivy.yaml
```diff
@@ -0,0 +1,2 @@
+vulnerability:
```
ziofil, line 1: ooh! did not know this could be done

ziofil (approved):
LGTM

trivy.yaml
SamFerracin: FYI this is not a real vulnerability, it's Code Factor being paranoid.
We have `grpcio` pinned to a safe version, but ray has it pinned to `>=x`, where `x` is an unsafe version.
I can do nothing to change that, apart from adding this `trivy` file

pyproject.toml
```diff
@@ -26,14 +26,14 @@ classifiers = [
...
-thewalrus = "0.19.0"
-rich = "10.15.1"
-matplotlib = "3.5.0"
-tqdm = "^4.65.0"
-ray = { version = "2.5.0", extras = ["tune"], optional = true }
+grpcio = "1.60.0"
+numpy = "^1.23.5"
+scipy = "^1.8.0"
+numba = "^0.56.4"
+thewalrus = "~0.19.0"
```
sduquemesa, line 33: Why not bumping the walrus version? Is there any breaking change in the new releases?

.github/workflows/builds.yml
```diff
@@ -44,5 +44,13 @@ jobs:
       - name: Install only test dependencies
         run: poetry install --no-root --extras "ray" --with dev
 
+      - name: Setup Julia
```
sduquemesa, line 47: will this workflow also run the julia tests? (is there any flag you have to add to the pytest section?)

sduquemesa (approved):
Nice one @SamFerracin! Looking good to me

pyproject.toml
SamFerracin: This opens up a bigger conversation, which is that I think it would be helpful to have every dependency unpinned.
With a good nightly build one can learn about bugs that come with new versions and fix them right away.

This is the only way to make sure that people doing `pip install mrmustard` on an existing environment can have it working

PR #339, opened 2024-02-05
Removing TF warnings on import with `os` instead of `logging`

SamFerracin:
**Context:**
The previous solution based on `logging` does not appear to work on every OS


ziofil (approved):
great!

PR #341, opened 2024-02-06
Docs and tests for Fock and ArrayAnsatz, and a few minor changes to Representation docs to make them consistent

SamFerracin:
**Context:**
Last before merging Fock and ArrayAnsatz into `develop`

### DIFF
--- a/mrmustard/physics/representations.py
+++ b/mrmustard/physics/representations.py
@@ -134,10 +134,9 @@ class Bargmann(Representation):
         >>> # bargmann representation of one-mode dgate with gamma=1+0j
         >>> rep_dgate = Bargmann(*displacement_gate_Abc(1))
 
-    The inner product is defined as the contraction of two Bargmann objects across
-    marked indices.
-    (@filippo can you help me find a simple way to explain the story of the marked
-    indices?)
+    The inner product is defined as the contraction of two Bargmann objects across marked indices.
+    Indices are marked using ``__getitem__``. Once the indices are marked for contraction, they are
+    be used the next time the inner product (``@``) is called. For example:
 
     .. code-block ::
 
@@ -286,7 +285,6 @@ class Bargmann(Representation):
         Returns:
             The reordered Bargmann object.
         """
-        # @filippo, should we add error-checking? What if I pass an order that is incorrect (e.g., shorter)?
         A, b, c = bargmann.reorder_abc((self.A, self.b, self.c), order)
         return self.__class__(A, b, c)
 


mrmustard/physics/representations.py
```diff
@@ -99,47 +104,95 @@ def __and__(self, other: Representation) -> Representation:
...
-            psi3 = 1.3 * psi1 - 2.1 * psi2  # linear combination
-            assert psi3.A.shape == (2, 1, 1)  # stacked along batch dimension
-            psi4 = psi1[0] @ psi2[0]  # contract wires 0 on each (inner product)
-            assert psi4.A.shape == (1,)  # A is 0x0 now (no wires left)
-            psi5 = psi1 & psi2  # outer product (tensor product)
-            rho = psi1.conj() & psi1   # outer product (this is now the density matrix)
-            assert rho.A.shape == (1, 2, 2)  # we have two wires now
-            assert np.allclose(rho.trace((0,), (1,)), np.abs(c)**2)
+    marked indices.
+    (@filippo can you help me find a simple way to explain the story of the marked
```
sylviemonet, line 139: @ziofil

sylviemonet (approved):
I think it is a nice PR! Just point out that there a task for Filippo there! :)

tests/test_physics/test_representations.py
```diff
@@ -0,0 +1,274 @@
...
+
+        bargmann1 = Bargmann(*triple1)
+        bargmann2 = Bargmann(*triple2)
+        bargmann_add = bargmann1 - bargmann2
+
+        assert np.allclose(bargmann_add.A, math.concat([bargmann1.A, bargmann2.A], axis=0))
+        assert np.allclose(bargmann_add.b, math.concat([bargmann1.b, bargmann2.b], axis=0))
+        assert np.allclose(bargmann_add.c, math.concat([bargmann1.c, -bargmann2.c], axis=0))
+
+    # def test_trace(self):
```
SamFerracin, line 144: @ziofil could you try running this test with `--backend=tensorflow`?
I am getting an error, and I am not sure if I'm doing something wrong or there's a small bug

mrmustard/physics/representations.py
```diff
@@ -248,8 +266,27 @@ def trace(self, idx_z: tuple[int, ...], idx_zconj: tuple[int, ...]) -> Bargmann:
...
+
+            >>> assert rep_dgate1.reorder([1, 2, 0, 4, 5, 3]) == rep_dgate2
+
+        Args:
+            order: The new order.
+
+        Returns:
+            The reordered Bargmann object.
+        """
+        # @filippo, should we add error-checking? What if I pass an order that is incorrect (e.g., shorter)?
```
SamFerracin, line 289: @ziofil

mrmustard/physics/representations.py
```diff
@@ -93,73 +110,108 @@ def __and__(self, other: Representation) -> Representation:
...
-    exponentials of bilinear forms with a polynomial part:
+    The ansatz available in this representation is a linear combination of exponentials
+    of bilinear forms with a polynomial part:
 
     .. math::
         F(z) = \sum_i \textrm{poly}_i(z) \textrm{exp}(z^T A_i z / 2 + z^T b_i)
 
     This function allows for vector space operations on Bargmann objects including
-    linear combinations, outer product, and inner product.
+    linear combinations, outer product (``&``), and inner product (``@``).
```
ziofil, line 124:
```suggestion
    linear combinations (``+``), outer product (``&``), and inner product (``@``).
```

mrmustard/physics/representations.py
```diff
@@ -93,73 +110,108 @@ def __and__(self, other: Representation) -> Representation:
...
-            assert psi3.A.shape == (2, 1, 1)  # stacked along batch dimension
-            psi4 = psi1[0] @ psi2[0]  # contract wires 0 on each (inner product)
-            assert psi4.A.shape == (1,)  # A is 0x0 now (no wires left)
-            psi5 = psi1 & psi2  # outer product (tensor product)
-            rho = psi1.conj() & psi1   # outer product (this is now the density matrix)
-            assert rho.A.shape == (1, 2, 2)  # we have two wires now
-            assert np.allclose(rho.trace((0,), (1,)), np.abs(c)**2)
+    marked indices.
+    (@filippo can you help me find a simple way to explain the story of the marked
+    indices?)
```
ziofil, line 140:
```suggestion
    marked indices. Indices are marked using the getitem notation, e.g. we mark 3 indices
    for contraction like so: ``barg[1,2,3]``. Once indices are marked for contraction they will be used the next
    time the inner product (``@``) is called. For example:
     
    .. code-block ::
        >>> from mrmustard.physics.representations import Bargmann
        >>> from mrmustard.physics.triples import displacement_gate_Abc, vacuum_state_Abc
         
        >>> vac = Bargmann(*vacuum_state_Abc(1))
        >>> D = Bargmann(*displacement_gate_Abc(1))
         
        >>> vac[0] @ D[1]  # contracts index 0 of vac with index 1 (input) of displacement

    Note that ``Bargmann`` does not know what the indices refer to and will happily contract any given set of pairs
    as long as they are available (e.g. the marked index does not exceed the dimension of the Abc representation).
```

mrmustard/physics/representations.py
ziofil: I think if the size is wrong something will break eventually, so it's not strictly needed. If we go this way we could consider allowing a shorter order to "extract" only those rows/cols on top of reordering (which is already happening on the A/b part via `gather`, but then `transpose` would complain if we have an array version of c). Otherwise we can check the length and move on.

ziofil (approved):
👍

PR #342, opened 2024-02-07
Fix `math.Categorical`

SamFerracin:
**Context:**
```
probs = np.array([1e-6 for _ in range(300)])
results = [math.Categorical(probs, "") for _ in range(100)]
assert len(set(results)) > 1
```
When using the numpy backend, the code above fails: `results` contains 100 identical values.
This is because `math.Categorical` uses `np.random.multinomial`, which assumes that the probabilities sum up to `1` (if not, the last probability is increased such that the resulting probabilities sum up to `1`)

### DIFF
--- a/mrmustard/math/backend_numpy.py
+++ b/mrmustard/math/backend_numpy.py
@@ -367,7 +367,7 @@ class BackendNumpy(BackendBase):  # pragma: no cover
                 self._probs = probs
 
             def sample(self):
-                array = np.random.multinomial(1, pvals=probs)
+                array = np.random.multinomial(1, pvals=probs/sum(probs))
                 return np.where(array == 1)[0][0]
 
         return Generator(probs)
--- a/mrmustard/math/backend_numpy.py
+++ b/mrmustard/math/backend_numpy.py
@@ -367,7 +367,7 @@ class BackendNumpy(BackendBase):  # pragma: no cover
                 self._probs = probs
 
             def sample(self):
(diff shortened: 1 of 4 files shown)


mrmustard/math/backend_numpy.py
```diff
@@ -367,8 +367,8 @@ def __init__(self, probs):
                 self._probs = probs
 
             def sample(self):
-                array = np.random.multinomial(1, pvals=probs)
-                return np.where(array == 1)[0][0]
+                idx = [i for i, _ in enumerate(probs)]
```
elib20, line 370: Where is idx used?

SamFerracin: Oops. I pushed again

tests/test_lab/test_detectors.py
```diff
@@ -319,6 +319,35 @@ def test_sampling_mean_and_var(
...
+        ],
+    )
+    def test_sampling_mean_and_var_unnormalized_state(
+        self, state, kwargs, mean_expected, var_expected
+    ):
+        """Tests that the mean and variance estimates of many homodyne
+        measurements are in agreement with the expected values for the given (unnormalized) states
+        """
+        state = state(**kwargs)
+        state = State(dm=state.dm(cutoffs=[10]) / 3)
```
elib20, line 337: Since this is the only difference with the test above (that I can detect), why not add a "normalization" argument to the function above, and then add to parametrize to check for both [1,3]?

elib20 (approved):
Nice!

PR #346, opened 2024-02-13
fix the c of the attenuator triples needs to be 1

sylviemonet:
**Context:** Bug fix of the triples for attenuator

**Description of the Change:** c=1.0 for attenuator

**Benefits:**

**Possible Drawbacks:**

**Related GitHub Issues:**


SamFerracin (approved):
Approving because it matches the Attenuator on mrmustard.lab, but I trust Filippo to make sure that this is correct

ziofil:
let's first triple check the theory

ziofil:
Ok, I verified, I'm happy now :)

ziofil (approved):
good catch!

PR #347, opened 2024-02-14
Adding ``__matmul__`` to ``CircuitComponent``s

SamFerracin:
**Context:**
Adds matmul to circuit components.
Also changes/adds a bunch of stuff that is needed for it, such as:
- Adds equality on wires and circuit components (so that tests become easier)
-  Changes ``Wires.__rshift__`` into ``Wires.__matmul__``, so that all rshift (matmul) in lab_dev mean the same thing, i.e. that they do (do not) add missing adjoints

**Tests:**
For the tests, it applies ``@`` to sequences of states and gates, and compares the representation of the resulting component to those that one gets by implementing the same sequence with `mrmustard.lab`. Small differences found, which can be discussed offline


mrmustard/lab_dev/circuit_components.py
```diff
@@ -181,6 +191,48 @@ def __getitem__(self, idx: Union[Mode, Sequence[Mode]]):
...
         return ret
 
+    def __matmul__(self, other: CircuitComponent) -> CircuitComponent:
+        r"""
+        Contracts ``self`` and ``other``, without adding adjoints.
+        """
+        # initialized the ``Wires`` of the returned component
+        wires_ret = self.wires @ other.wires
+
+        # find the indices of the wires being contracted on the bra side
```
SamFerracin, line 201: @ziofil to get to this code, I copy/pasted your code and did two things:
- Added logic to find the indices of the contracted wires on bra and ket side separately. Doing it as in your PR (see the ``common`` array) gave a bug with, e.g., ``dgate @ attenuator``, where ``dgate`` has no bra wire
- There was as issue with how you were finding the indices to reorder the final representation, what used to be ``before_ids``. I fixed it
Feel free to ask questions if you find more differences

tests/test_lab_dev/test_circuit_components.py
```diff
@@ -37,6 +41,79 @@ def test_light_copy(self):
...
+        a0 = Attenuator(0.8, modes=[0])
+        a1 = Attenuator(0.8, modes=[1])
+        a2 = Attenuator(0.7, modes=[2])
+
+        result = vac012 @ d0 @ d1 @ d2
+        result = result @ result.adjoint @ a0 @ a1 @ a2
+
+        assert result.wires == Wires(modes_out_bra=[0, 1, 2], modes_out_ket=[0, 1, 2])
+        assert np.allclose(result.representation.A, 0)
+        assert np.allclose(
```
SamFerracin, line 83: @sylviemonet @ziofil 
In mrmustard.lab, one would get the same results by doing ``(vac012 >> d0 >> d1 >> d2 >> a0 >> a1 >> a2).bargmann()`. If I do this, I get the Bargmann triple of a pure state. I guess that the attenuator is mapping ket to ket? But in lab_dev, it is a channel

Just wondering if this is the reason

mrmustard/lab_dev/states.py
```diff
@@ -70,5 +70,5 @@ def representation(self) -> Bargmann:
         num_modes = len(self.modes)
         A = math.zeros(shape=(num_modes, num_modes), dtype=math.complex128)
         B = math.zeros(shape=(num_modes), dtype=math.complex128)
-        C = 1.0
+        C = math.cast(1.0, dtype=math.complex128)
```
sylviemonet, line 73: Or change it as you suggested:
```suggestion
        C = 1.0 + 0j
```

mrmustard/lab_dev/transformations.py
```diff
@@ -185,15 +185,15 @@ def representation(self) -> Bargmann:
...
+                    [np.sqrt(e), 0, 0, 1 - e],
+                    [0, 0, 0, np.sqrt(e)],
+                    [0, 1 - e, np.sqrt(e), 0],
                 ],
             ),
             math.complex128,
         )
         B = math.cast([0.0, 0.0, 0.0, 0.0], math.complex128)
-        C = math.cast(np.sqrt(e), math.complex128)
+        C = math.cast(1.0, dtype=math.complex128)
```
sylviemonet, line 197:
```suggestion
        C = 1.0 + 0j
```

mrmustard/physics/triples.py
```diff
@@ -392,7 +392,7 @@ def attenuator_Abc(eta: Union[Scalar, Iterable]) -> Union[Matrix, Vector, Scalar
 
     A = A[reshape_list, :][:, reshape_list]
     b = _vacuum_B_vector(n_modes * 2)
-    c = np.prod(eta)
+    c = 1.0
```
sylviemonet, line 395: will be replaced when the new fix PR goes in!

sylviemonet (commented):
Nice PR!
I've checked the logic in side matmul which works for me!

_Just leave an open question to discuss: if we do this matmul in simulator and with Fock in each step, would the idea to have the tensor network doesn't exist anymore?_

tests/test_lab_dev/test_circuit_components.py
```diff
@@ -37,6 +41,79 @@ def test_light_copy(self):
         assert d.y is d_copy.y
         assert d.wires is not d_copy.wires
 
+    def test_matmul_one_mode(self):
```
sylviemonet, line 44: @sylviemonet Add a test for the phase factor!.

sylviemonet: @SamFerracin @ziofil How about add tests with the projection of states as well?

SamFerracin:
> Nice PR! I've checked the logic in side matmul which works for me!
> 
> _Just leave an open question to discuss: if we do this matmul in simulator and with Fock in each step, would the idea to have the tensor network doesn't exist anymore?_

The tensor network idea still exists, but it is orthogonal to the matmul/rshift.
When we specify a path for a circuit, the setter will take care to check that the path is valid for the given circuit based on the connectivity of the components. This way, the check is done only once and outside of the simulation (remember that I may want to simulate the same circuit 1000 times)

mrmustard/lab_dev/circuit_components.py
SamFerracin: [@ziofil]: "Why do you need to find the indices for ket and bra separately?"

On Filippo's "state_barebones" branch, try:
```
from mrmustard.lab_dev.transformations import *

dgate = Dgate(1, modes=[0])
attenuator = Attenuator(1, modes=[0])

dgate @ attenuator
```

The last line gives the following error: 
```
File ~/git/MrMustard/mrmustard/physics/bargmann.py:140, in complex_gaussian_integral(Abc, idx_z, idx_zconj, measure)
    138 A, b, c = Abc
    139 if len(idx_z) != len(idx_zconj):
--> 140     raise ValueError("idx_z and idx_zconj must have the same length")
    141 n = len(idx_z)
    142 idx = tuple(idx_z) + tuple(idx_zconj)

ValueError: idx_z and idx_zconj must have the same length
```

The issue is caused by the first 4 lines in matmul:
```
common = set(self.wires.output.modes).intersection(other.wires.input.modes)
new_wires = self.wires >> other.wires
idx_z = self.wires[common].output.indices
idx_zconj = other.wires[common].input.indices
``` 
Here, ``common=[0]``, ``idx_z=[0]`` and ``idx_zconj=[1,3]``. So it's trying to contract one wire (the ket output of the Dgate) with two wires (the ket in and bra in of the attenuator). 
Finding the indices separately for ket and bra overcomes the issue, because the code will recognize that there are no wires to be contracted on the bra side

ziofil: ah I see! Thanks for the explanation

tests/test_lab_dev/test_circuit_components.py
sylviemonet: Not wrong with Attenuator. 

Just inside bargmann() function, it checks if the state is pure or not and gives the triples. 
In our lab_dev yes, we make the decision on our own not on the state to check at this point.

SamFerracin: Thanks for checking this!

Here is the other discrepancy, which I find when I project a multi-mode state:

MrMustard lab:
```
import mrmustard.lab as mm

vac0 = mm.Vacuum(1)
vac01 = mm.Vacuum(2)
vac012 = mm.Vacuum(3)
d0 = mm.Dgate(.1, .1, modes=[0])
d1 = mm.Dgate(.1, .1, modes=[1])
d2 = mm.Dgate(.1, .1, modes=[2])
a0 = mm.Attenuator(.8, modes=[0])
a1 = mm.Attenuator(.8, modes=[1])
a2 = mm.Attenuator(.7, modes=[2])

result = vac012 >> d0 >> d1 >> d2 >> a0 >> a1 >> a2 << vac012
result
```


MrMustard lab_dev:
```
import mrmustard.lab_dev as mmd

vac0 = mmd.Vacuum(1)
vac01 = mmd.Vacuum(2)
vac012 = mmd.Vacuum(3)
d0 = mmd.Dgate(.1, .1, modes=[0])
d1 = mmd.Dgate(.1, .1, modes=[1])
d2 = mmd.Dgate(.1, .1, modes=[2])
a0 = mmd.Attenuator(.8, modes=[0])
a1 = mmd.Attenuator(.8, modes=[1])
a2 = mmd.Attenuator(.7, modes=[2])

result = vac012 @ d0 @ d1 @ d2
result = (result.adjoint) @ a0 @ a1 @ a2 @ vac012.dual @ vac012.dual.adjoint
result.representation.c
```

You can observe similar discrepancies if you project onto `vac01`

sylviemonet: `result = (result @ result.adjoint) @ a0 @ a1 @ a2 @ vac012.dual @ vac012.dual.adjoint` would this be result and result adjoint at the beginning?

sylviemonet (approved):
It looks good to me now!

mrmustard/lab_dev/__init__.py
```diff
@@ -11,3 +11,10 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
+
+from .circuit_components import CircuitComponent
+from .circuits import Circuit
+from .simulator import Simulator
+from .states import *
```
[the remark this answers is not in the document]
SamFerracin: Good question. I don't think it's an issue, but I like the idea of banning wildcards for being more explicit

mrmustard/lab_dev/wires.py
```diff
@@ -300,6 +300,26 @@ def __bool__(self) -> bool:
...
+            return False
+        if self.input.bra.modes != other.input.bra.modes:
+            return False
+        if self.output.ket.modes != other.output.ket.modes:
+            return False
+        if self.input.ket.modes != other.input.ket.modes:
+            return False
+        return True
+
+    def __neq__(self, other) -> bool:
```
ziofil, line 317: this is redundant (if a class has `__eq__`, python makes `__neq__` exactly like it's written here).

tests/test_lab_dev/test_circuit_components.py
ziofil: is this resolved or does it need discussing?

mrmustard/lab_dev/circuit_components.py
```diff
@@ -181,6 +191,48 @@ def __getitem__(self, idx: Union[Mode, Sequence[Mode]]):
...
+        ket_modes = set(self.wires.ket.output.modes).intersection(other.wires.ket.input.modes)
+        idx_z += self.wires[ket_modes].ket.output.indices
+        idx_zconj += other.wires[ket_modes].ket.input.indices
+
+        # convert Bargmann -> Fock if needed
+        LEFT = self.representation
+        RIGHT = other.representation
+        if isinstance(LEFT, Bargmann) and isinstance(RIGHT, Fock):
+            raise ValueError("Cannot contract objects with different representations.")
+            # shape = [s if i in idx_z else None for i, s in enumerate(other.representation.shape)]
```
ziofil, line 216: where is the conversion taking place now?

tests/test_lab_dev/test_circuit_components.py
SamFerracin: Resolved and fixed in one of @sylviemonet's latest PRs

mrmustard/lab_dev/circuit_components.py
SamFerracin: Right now we don't support conversions. If matmul finds that the representations are different, right now it will error. The error will be removed once we have the converters in place

ziofil (approved):
LGTM!

PR #348, opened 2024-02-15
Fix the length of b Vector for all the objects and the corresponding tests

sylviemonet:
**Context:** Because of the tests were `np.allclose(b, 0)`, we haven't tested the length of the b vector just the values. Hence we have mistakes in the length of b Vector of rotation gate, thermal state and attenuator/amplifier channels.
 In this PR, we fix the b vectors of them and fix the tests with respect to the shape of the vector it should be. 
We also fix the type hint for the functions.

**Benefits:**

**Possible Drawbacks:**

**Related GitHub Issues:**


tests/test_physics/test_triples.py
```diff
@@ -48,39 +48,39 @@ def test_vacuum_state_Abc(self, n_modes):
         A, b, c = triples.vacuum_state_Abc(n_modes)
 
         assert np.allclose(A, np.zeros((n_modes, n_modes)))
-        assert np.allclose(b, np.zeros((n_modes)))
+        assert np.allclose(b, np.zeros(n_modes))
```
SamFerracin, line 51: I think it'd be best to use `np.all` instead of `np.allclose`

sylviemonet: Working on it

SamFerracin (approved):
:top:

mrmustard/math/backend_numpy.py
```diff
@@ -59,6 +59,13 @@ def __repr__(self) -> str:
     def abs(self, array: np.ndarray) -> np.ndarray:
         return np.abs(array)
 
+    def allclose(self, array1: np.array, array2: np.array, atol: float) -> bool:
```
ziofil, line 62: why is it necessary to reproduce this in the backend? There's no need for gradients

mrmustard/math/backend_tensorflow.py
```diff
@@ -83,13 +90,13 @@ def astensor(self, array: Union[np.ndarray, tf.Tensor], dtype=None) -> tf.Tensor
         return tf.convert_to_tensor(array, dtype)
 
     def atleast_1d(self, array: tf.Tensor, dtype=None) -> tf.Tensor:
-        return tf.experimental.numpy.atleast_1d(self.astensor(array, dtype))
+        return tf.experimental.numpy.atleast_1d(self.cast(self.astensor(array), dtype))
```
ziofil, line 93: why this change?

mrmustard/math/backend_numpy.py
SamFerracin: For the tests, we needed an allclose that
- takes `atol`
- checks the shapes
Numpy's and tensorflow's don't do that. So we thought it best to add it to the backend, so we can use the same `allclose` in all the test files

ziofil (approved):
ok. I don't get why we need to maintain our own `allclose`...

mrmustard/math/backend_tensorflow.py
SamFerracin: An error showed up that can be reproduced by the following snippet:

```
import tensorflow as tf

my_tensor = tf.convert_to_tensor([1, 2])
tf.convert_to_tensor(my_tensor, tf.complex128)
--> ValueError: Tensor conversion requested dtype complex128 for Tensor with dtype int32: <tf.Tensor: shape=(2,), dtype=int32, numpy=array([1, 2], dtype=int32)>
```

PR #349, opened 2024-02-16
Fix the small bugs

sylviemonet:
**Context:** There are three bugs fixed in this PR.

1. the projection onto the Vacuum state will use the generaldyne where the projection function is not correct.
For example,
`from mrmustard.lab import *;Vacuum(3) << Vacuum(3)`
The projection in the example above is not 1.0, which should be 1.0.
This has been fixed by changing the functions inside generaldyne.
3. add more text to explain about complex_gaussian_integral and other function in bargmann.py.
4. fix the problems that the state will be reassigned during the running of the circuit.
For example,
`from mrmustard.lab import *;
vac0 = Vacuum(1);
d1 = Dgate(0.1, 0.1);
vac0 >> d1;
vac0.bargmann()`

the state `vac0` state will be modified after going through the first Dgate, which is not the behaviour we want.

**Benefits:**

**Possible Drawbacks:**

**Related GitHub Issues:**

### DIFF
--- a/.github/CHANGELOG.md
+++ b/.github/CHANGELOG.md
@@ -119,7 +119,6 @@ which uses the old Numba code. When setting to a higher value, the new Julia cod
 * Fixing a bug where `scipy.linalg.sqrtm` returns an unsupported type.
 [(#337)](https://github.com/XanaduAI/MrMustard/pull/337)
 
-
 ### Documentation
 
 ### Tests
--- a/.github/CHANGELOG.md
+++ b/.github/CHANGELOG.md
@@ -114,6 +114,10 @@ which uses the old Numba code. When setting to a higher value, the new Julia cod
 [(#309)](https://github.com/XanaduAI/MrMustard/pull/309)
 * Fixing a bug where `scipy.linalg.sqrtm` returns an unsupported type.
 [(#337)](https://github.com/XanaduAI/MrMustard/pull/337)
+* Fixing a bug in `_transform_gaussian` in transformation.py that modifies the input state's cov and means.
+[(#349)](https://github.com/XanaduAI/MrMustard/pull/349)
+* Fixing a bug in `general_dyne` in physics/gaussian.py that returns the wrong probability and outcomes with given projection.
+[(#349)](https://github.com/XanaduAI/MrMustard/pull/349)
 
 ### Documentation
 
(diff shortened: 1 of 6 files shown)


mrmustard/physics/gaussian.py
```diff
@@ -585,11 +585,24 @@ def general_dyne(
...
+    if proj_means is None:
+        pdf = math.MultivariateNormalTriL(loc=b, scale_tril=math.cholesky(reduced_cov / 2))
+        outcome = (
+            pdf.sample(dtype=cov.dtype) if proj_means is None else math.cast(proj_means, cov.dtype)
+        )
+        prob = pdf.prob(outcome)
+    else:
+        # If the projector is already given: proj_means
+        # use the formula 5.139 in Serafini - Quantum Continuous Variables
+        # fixed by -0.5 on the exponential, hbar and pi with adaptation changes
```
ziofil, line 597:
```suggestion
        # fixed by -0.5 on the exponential, added hbar and removed pi due to different convention
```

ziofil (approved):
awesome!

.github/CHANGELOG.md
```diff
@@ -114,6 +114,10 @@ which uses the old Numba code. When setting to a higher value, the new Julia cod
 [(#309)](https://github.com/XanaduAI/MrMustard/pull/309)
 * Fixing a bug where `scipy.linalg.sqrtm` returns an unsupported type.
 [(#337)](https://github.com/XanaduAI/MrMustard/pull/337)
+* Fixing a bug in `_transform_gaussian` in transformation.py that modifies the input state's cov and means.
```
SamFerracin, line 117: This is not in the right place. Should be in the top section of the file (0.7.1)

SamFerracin (commented):
Can you add tests for these bugs? So we'll remember to keep an eye on these things when we merge the refactoring

.github/CHANGELOG.md
sylviemonet: Got it! I'll move it to there

.github/CHANGELOG.md
```diff
@@ -115,6 +119,7 @@ which uses the old Numba code. When setting to a higher value, the new Julia cod
 * Fixing a bug where `scipy.linalg.sqrtm` returns an unsupported type.
 [(#337)](https://github.com/XanaduAI/MrMustard/pull/337)
 
+
```
SamFerracin, line 122:
```suggestion
```

SamFerracin (approved):
Just one small change :)

PR #351, opened 2024-02-22
Updating dependencies to support python 3.11

SamFerracin:
**Context:**
Updates dependencies so that MrMustard runs with python 3.11, and makes minimal changes to make all tests pass (check SisterPam)


tests/test_lab/test_states.py
```diff
@@ -39,7 +39,7 @@
 @st.composite
 def xy_arrays(draw):
     length = draw(st.integers(2, 10))
-    return draw(arrays(dtype=np.float, shape=(2, length), elements=st.floats(-5.0, 5.0)))
+    return draw(arrays(dtype=np.float64, shape=(2, length), elements=st.floats(-5.0, 5.0)))
```
zeyueN, line 42: I think [`np.float_`](https://numpy.org/doc/stable/reference/arrays.scalars.html#numpy.float_) could also work.

zeyueN (approved):

sduquemesa (approved):
Looks good to me! I'm leaving a comment that does not prevent this from being merged.

_The comment_ — the reason this PR won't fail for python 3.9 is because you're using `from __future__ import annotation` whenever the operator `|` is used to annotate the types. This operator for type annotations is actually supported in python3.10. From the documentation 

> The future statement is intended to ease _migration_ to future versions of Python that introduce incompatible changes to the language. It allows use of the new features on a per-module basis before the release in which the feature becomes standard.

This is the reason why MrMustard in python3.9 is able to run with features of python3.10. Also, the documentation for future statements says (see [PEP 236](https://peps.python.org/pep-0236/))

> Note that there is no need to involve the future_statement machinery in new features unless they can break existing code; fully backward- compatible additions can– and should –be introduced without a corresponding future_statement.

Given that MrMustard is already supported in python3.10-3.11 and considering that the purpose of the statement `from __future__ import annotation` is to ease _migration_ to newer python versions and that it could introduce breaking changes in the future as new features are added to python, something to think about is the possibility of removing support for python3.9 —and hence removing the need for futures— if and only if no other dependent libraries use python3.9 (not the case for MrKite).

PR #354, opened 2024-02-23
Updating `pylint` in order to bump `wrapt`

SamFerracin:
**Context:**
When `poetry` is used to install Mr Mustard and its dependencies, a package called `wrapt` is installed in an outdated version `1.12.1`. When importing `tensorflow`, this outdated version causes an import error.
This did not show up on SisterPam, because `pip install mrmustard` installs a different set of dependencies

The current version of pylint requires `wrapt < 1.13`. Bumping it up allows `wrapt` to be installed in a more recent (though not the most recent) version.

I also update the tests files, so that they now test python 3.11


zeyueN (approved):
Thanks @SamFerracin !

Side note, unless there's specific pylint features we need, I've become a strong advocator of [ruff ](https://github.com/astral-sh/ruff)for linting.

sduquemesa (approved):

PR #358, opened 2024-02-29
Porting Transformations to develop

SamFerracin:
**Context:**
Porting transformations to develop


ziofil (approved):
🚀

mrmustard/lab_dev/transformations/base.py
```diff
@@ -0,0 +1,114 @@
...
+        if isinstance(other, Unitary):
+            unitary = Unitary()
+            unitary._wires = component.wires
+            unitary._representation = component.representation
+            return unitary
+        elif isinstance(other, Channel):
+            channel = Channel()
+            channel._wires = component.wires
+            channel._representation = component.representation
+            return channel
```
ziofil, line 73: With a small change (I think), we could write these as 
```python
if isinstance(other, Unitary):
    return Unitary().from_attributes(...)
elif isinstance(other, Channel):
    return Channel().from_attributes(...)

ziofil (approved):
super!

mrmustard/lab_dev/transformations/base.py
SamFerracin: Not out of the box, it would still give a `CircuitComponent`. I still have to figure out how to do it that way, I agree that it would be much cleaner

PR #359, opened 2024-03-01
Porting States to develop

SamFerracin:
**Context:**
Porting all the States in `prototype-tn` into `develop`, with the addition of `Number`.

**Note:**
Please review `Number` very well as it has not been reviewed in the prototype branch


mrmustard/lab_dev/states/states.py
```diff
@@ -0,0 +1,179 @@
...
+        :math:`n_j`.
+
+    """
+
+    def __init__(
+        self, modes: Iterable[int], n: Union[int, Iterable[int]], cutoff: Optional[int] = None
+    ) -> None:
+        super().__init__("N", modes=modes)
+        self._add_parameter(make_parameter(False, n, "n", (None, None)))
+        self._add_parameter(
```
sylviemonet, line 122: would this parameter means "trainable" ones? cutoffs is not a trainable parameter ?

tests/test_lab_dev/test_states.py
```diff
@@ -0,0 +1,225 @@
...
+        assert state.modes == [modes] if not isinstance(modes, list) else sorted(modes)
+
+    def test_init_error(self):
+        with pytest.raises(ValueError, match="Length of ``x``"):
+            Coherent(modes=[0, 1], x=[2, 3, 4])
+
+        with pytest.raises(ValueError, match="Length of ``y``"):
+            Coherent(modes=[0, 1], x=1, y=[2, 3, 4])
+
+    def test_trainable_parameters(self):
```
sylviemonet, line 135: Nice!

tests/test_lab_dev/test_states.py
```diff
@@ -0,0 +1,225 @@
...
+        ket = Coherent([0, 1], 1)
+        channel = Attenuator([1], 1)
+        dm = ket >> channel
+        dm_component = CircuitComponent.from_attributes(dm.name, dm.representation, dm.wires)
+
+        assert repr(dm) == "DM(name=None, modes=[0, 1])"
+        assert repr(dm_component) == "CircuitComponent(name=None, modes=[0, 1])"
+
+
+class TestCoherent:
```
sylviemonet, line 112: Just curious if we want to test the superposition of two coherent states, where to add that test?

sylviemonet (approved):
Looks good to me. 

Only three states here though. Do we have a plan for others? like the squeezed vacuum is something definitely useful in that notebook.

mrmustard/lab_dev/states/states.py
SamFerracin: They would not be, because of the `False`.
But I removed the `_add_parameter` to make it more explicit, and added properties and errors as needed

tests/test_lab_dev/test_states.py
SamFerracin: Integration? We can add an "Integration tests" folder on SisterPam now

SamFerracin:
> Looks good to me.
> 
> Only three states here though. Do we have a plan for others? like the squeezed vacuum is something definitely useful in that notebook.

@sylviemonet for sure! I just wanted to get a few states and transformations out to get the ball rolling, but now we can easily add every state and gate that was previously supported

PR #353, opened 2024-02-23
Adding `__rshift__` to `CircuitComponent`s

SamFerracin:
**Context:**
Adds ``>>`` to circuit components. A few other minimal changes to make it work

PR #361, opened 2024-03-05
Adding support for `to_fock` conversion of circuit components

SamFerracin:
**Context:**
- Allows converting Circuit Components with Bargmann representation into components with Fock representations.
- Extends `>>` to components with different representations

### DIFF
--- a/mrmustard/lab_dev/circuit_components.py
+++ b/mrmustard/lab_dev/circuit_components.py
@@ -236,25 +236,25 @@ class CircuitComponent:
         """
         msg = f"``__rshift__`` not supported between {self} and {other}, use ``__matmul__``."
 
-        wires_out = self.wires
-        wires_in = other.wires
+        wires_s = self.wires
+        wires_o = other.wires
 
-        if wires_out.ket and wires_out.bra:
-            if wires_in.ket and wires_in.bra:
+        if wires_s.ket and wires_s.bra:
+            if wires_o.ket and wires_o.bra:
                 return self @ other
             return self @ other @ other.adjoint
 
-        if wires_out.ket:
-            if wires_in.ket and wires_in.bra:
+        if wires_s.ket:
+            if wires_o.ket and wires_o.bra:
                 return self @ self.adjoint @ other
-            if wires_in.ket:
+            if wires_o.ket:
                 return self @ other
             raise ValueError(msg)
 
-        if wires_out.bra:
-            if wires_in.ket and wires_in.bra:
+        if wires_s.bra:
+            if wires_o.ket and wires_o.bra:
                 return self @ self.adjoint @ other
-            if wires_in.bra:
+            if wires_o.bra:
                 return self @ other
             raise ValueError(msg)
 
--- a/mrmustard/lab_dev/circuit_components.py
+++ b/mrmustard/lab_dev/circuit_components.py
@@ -220,7 +220,8 @@ class CircuitComponent:
             other.wires.ids[i] for i in range(len(other.wires.ids)) if i not in idx_zconj
         ]
         order = [contracted_idx.index(id) for id in wires_ret.ids]
-        representation_ret = representation_ret.reorder(order)
+        if order:
+            representation_ret = representation_ret.reorder(order)
 
         return CircuitComponent.from_attributes("", representation_ret, wires_ret)
 
--- a/mrmustard/lab_dev/circuit_components.py
+++ b/mrmustard/lab_dev/circuit_components.py
@@ -175,7 +175,7 @@ class CircuitComponent:
 
             >>> d = Dgate([1], x=0.1, y=0.1)
             >>> d_fock = d.to_fock(shape=3)
-            
+
             >>> assert d_fock.name == d.name
             >>> assert d_fock.wires == d.wires
             >>> assert d_fock.representation == to_fock(d.representation, shape=3)
--- a/mrmustard/lab_dev/circuit_components.py
+++ b/mrmustard/lab_dev/circuit_components.py
@@ -22,9 +22,8 @@ from __future__ import annotations
 
 from typing import Iterable, Optional, Sequence, Union
 
-from mrmustard import settings
 from ..physics.converters import to_fock
-from ..physics.representations import Bargmann, Fock, Representation
+from ..physics.representations import Representation
 from ..math.parameter_set import ParameterSet
 from ..math.parameters import Constant, Variable
 from .wires import Wires
@@ -161,8 +160,32 @@ class CircuitComponent:
         instance.__dict__["_wires"] = self.wires.copy()
         return instance
 
-    def to_fock(self, shape: Optional[Union[int, Iterable[int]]] = None):
-        return type(self).from_attributes(
+    def to_fock(self, shape: Optional[Union[int, Iterable[int]]] = None) -> CircuitComponent:
+        r"""
+        Returns a circuit component with the same attributes as this component, but
+        with ``Fock`` representation.
+
+        Uses the :meth:`mrmustard.physics.converters.to_fock` method to convert the internal
+        representation.
+
+        .. code-block::
+
+            >>> from mrmustard.physics.converters import to_fock
+            >>> from mrmustard.lab_dev import Dgate
+
+            >>> d = Dgate([1], x=0.1, y=0.1)
+            >>> d_fock = d.to_fock(shape=shape)
+            
+            >>> assert d_fock.name == d.name
+            >>> assert d_fock.wires == d.wires
+            >>> assert d_fock.representation == to_fock(d.representation, shape)
+
+        Args:
+            shape: The shape of the returned representation. If ``shape``is given as
+                an ``int``, it is broadcasted to all the dimensions. If ``None``, it
+                defaults to the value of ``AUTOCUTOFF_MAX_CUTOFF`` in the settings.
+        """
+        return CircuitComponent.from_attributes(
             self.name,
             to_fock(self.representation, shape=shape),
             self.wires,
--- a/mrmustard/lab_dev/circuit_components.py
+++ b/mrmustard/lab_dev/circuit_components.py
@@ -174,11 +174,11 @@ class CircuitComponent:
             >>> from mrmustard.lab_dev import Dgate
 
             >>> d = Dgate([1], x=0.1, y=0.1)
-            >>> d_fock = d.to_fock(shape=shape)
+            >>> d_fock = d.to_fock(shape=3)
             
             >>> assert d_fock.name == d.name
             >>> assert d_fock.wires == d.wires
-            >>> assert d_fock.representation == to_fock(d.representation, shape)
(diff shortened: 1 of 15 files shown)


mrmustard/physics/representations.py
```diff
@@ -373,10 +374,30 @@ def __getitem__(self, idx: int | tuple[int, ...]) -> Bargmann:
...
+            A ``Bargmann`` representation if ``other`` is ``Bargmann``, or a ``Fock``representation
+            if ``other`` is ``Fock``.
         """
+        idx_z = self._contract_idxs
+        idx_zconj = other._contract_idxs
+
+        if isinstance(other, Fock):
+            from .converters import to_fock
+
+            shape = [settings.AUTOCUTOFF_MAX_CUTOFF for _ in range(len(self.b[0]))]
```
ziofil, line 396:
```suggestion
            shape = [settings.AUTOCUTOFF_MAX_CUTOFF for _ in range(len(self.b[-1]))]
```

mrmustard/physics/representations.py
```diff
@@ -482,17 +503,57 @@ def __getitem__(self, idx: int | tuple[int, ...]) -> Fock:
...
         """
-        axes = [list(self._contract_idxs), list(other._contract_idxs)]
+        idx_z = self._contract_idxs
+        idx_zconj = other._contract_idxs
+
+        # convert ``other`` to ``Fock``
+        if isinstance(other, Bargmann):
+            from .converters import to_fock
+
+            shape = [settings.AUTOCUTOFF_MAX_CUTOFF for _ in range(len(other.b[0]))]
```
ziofil, line 529:
```suggestion
            shape = [settings.AUTOCUTOFF_MAX_CUTOFF for _ in range(len(other.b[-1]))]
```

mrmustard/physics/representations.py
SamFerracin: It's the same. `self.b` is batched, so `self.b[0]` and `self.b[-1]` both select one batch

SamFerracin: @ziofil `len(self.b[0])` comes up pretty often, and seems to have the same meaning as `len(self.array)` for Fock. It'd be handy to have a property  ("batch_size"?) for this concept, and it would increase readability. Would you be against it?

mrmustard/lab_dev/circuit_components.py
```diff
@@ -216,8 +236,8 @@ def __rshift__(self, other: CircuitComponent) -> CircuitComponent:
         """
         msg = f"``__rshift__`` not supported between {self} and {other}, use ``__matmul__``."
 
-        wires_out = self.wires.output
-        wires_in = other.wires.input
+        wires_out = self.wires
```
SamFerracin, line 239: @ziofil I found the reason why we don't want `input` and `output` here (see [this discussion](https://github.com/XanaduAI/MrMustard/pull/353#discussion_r1506398824)).

Without them, right-shifting states (e.g., `Vacuum([0]) >> Vacuum([1])`) errors (says to use ``@``). 
Granted, this may be sub-optimal, but it may become annoying for people using the simulator (imagine launching a simulation on AWS with a path that says to do this contraction.. I would rather see the simulation's result in a slightly longer time than getting an error)

For this reason, considering that ``>>`` is meant to be high-level and user-friendly, I think that we should get rid of input and output here

mrmustard/lab_dev/circuit_components.py
```diff
@@ -159,6 +160,37 @@ def light_copy(self) -> CircuitComponent:
         instance.__dict__["_wires"] = self.wires.copy()
         return instance
 
+    def to_fock(self, shape: Optional[Union[int, Iterable[int]]] = None) -> CircuitComponent:
```
ziofil, line 163: should we call it `bargmann_to_fock`?

mrmustard/lab_dev/circuit_components.py
```diff
@@ -159,6 +160,37 @@ def light_copy(self) -> CircuitComponent:
...
+            >>> assert d_fock.representation == to_fock(d.representation, shape=3)
+
+        Args:
+            shape: The shape of the returned representation. If ``shape``is given as
+                an ``int``, it is broadcasted to all the dimensions. If ``None``, it
+                defaults to the value of ``AUTOCUTOFF_MAX_CUTOFF`` in the settings.
+        """
+        return CircuitComponent.from_attributes(
+            self.name,
+            to_fock(self.representation, shape=shape),
```
ziofil, line 190: I think it can be confusing to use the same name as the function we're defining. Let's think of alternative naming.

mrmustard/lab_dev/circuit_components.py
ziofil: ok. then we can call them `wires_self` and `wires_other`.

mrmustard/physics/representations.py
ziofil: haha, I had misread! I thought I had read `range(self.b.shape[0])` and I thought I wrote `range(self.b.shape[0])` 😂

mrmustard/physics/representations.py
```diff
@@ -373,27 +374,53 @@ def __getitem__(self, idx: int | tuple[int, ...]) -> Bargmann:
...
+
+            # set same shape along the contracted axes, and default shape along the
+            # axes that are not being contracted
+            shape = [settings.AUTOCUTOFF_MAX_CUTOFF for _ in range(len(self.b[0]))]
+            for i, j in zip(idx_s, idx_o):
+                shape[i] = other.array.shape[1:][j]
+
+            return to_fock(self, shape=shape)[idx_s] @ other[idx_o]
+
         if self.ansatz.degree > 0 or other.ansatz.degree > 0:
```
ziofil, line 405: This is awkward, because technically the Fock representation is all polynomial..

mrmustard/physics/representations.py
```diff
@@ -373,27 +374,53 @@ def __getitem__(self, idx: int | tuple[int, ...]) -> Bargmann:
...
+
+        # if ``other`` is ``Fock``, convert ``self`` to ``Fock``
+        if isinstance(other, Fock):
+            from .converters import to_fock  # pylint: disable=import-outside-toplevel
+
+            # set same shape along the contracted axes, and default shape along the
+            # axes that are not being contracted
+            shape = [settings.AUTOCUTOFF_MAX_CUTOFF for _ in range(len(self.b[0]))]
+            for i, j in zip(idx_s, idx_o):
+                shape[i] = other.array.shape[1:][j]
```
ziofil, line 401:
```suggestion
                shape[i] = min(shape[i], other.array.shape[1:][j])
```

mrmustard/physics/representations.py
```diff
@@ -482,20 +512,71 @@ def __getitem__(self, idx: int | tuple[int, ...]) -> Fock:
...
+        idx_s = list(self._contract_idxs)
+        idx_o = list(other._contract_idxs)
+
+        # if ``other`` is ``Bargmann``, convert it to ``Fock``
+        if isinstance(other, Bargmann):
+            from .converters import to_fock  # pylint: disable=import-outside-toplevel
+
+            # set same shape along the contracted axes, and default shape along the
+            # axes that are not being contracted
+            shape = [settings.AUTOCUTOFF_MAX_CUTOFF for _ in range(len(other.b[0]))]
```
ziofil, line 540: Reminder for self: implement `__len__` in Wires, so we can do `range(len(other.wires))`

mrmustard/physics/representations.py
```diff
@@ -535,3 +616,42 @@ def reorder(self, order: tuple[int, ...] | list[int]) -> Fock:
...
+            >>> assert fock4 == Fock(array4)
+
+        Args:
+            shape: The shape of the array of the returned ``Fock``.
+        """
+        length = len(self.array.shape)
+        shape = (shape,) * length if isinstance(shape, int) else shape
+        if len(shape) != length:
+            msg = f"Expected ``shape`` of lenght {length}, "
+            msg += f"found shape of lenght {len(shape)}."
```
ziofil, line 650: I like how you break down messages 👍

mrmustard/physics/representations.py
```diff
@@ -535,3 +616,42 @@ def reorder(self, order: tuple[int, ...] | list[int]) -> Fock:
...
+            raise ValueError(msg)
+
+        ret = self.array
+        for i, s in enumerate(shape):
+            slc = (slice(None),) * i + (slice(0, s),) + (slice(None),) * (length - i - 1)
+            ret = ret[slc]
+        return Fock(array=ret, batched=True)
```
ziofil, line 657:
```suggestion
        everything = (slice(None),)
        slc = tuple(everything*i + slice(0,s) + everything*(length - i - 1) for i,s in enumerate(shape))
        return Fock(array=self.array[slc], batched=True)
```

ziofil: maybe we can do all at once

mrmustard/physics/representations.py
```diff
@@ -482,20 +512,71 @@ def __getitem__(self, idx: int | tuple[int, ...]) -> Fock:
...
+
+        # the shapes of the axes being contracted
+        shape_s_contr = [shape_s[i] for i in idx_s]
+        shape_o_contr = [shape_o[i] for i in idx_o]
+
+        # compare the shapes along the axes being contracted
+        if shape_o_contr != shape_s_contr:
+            # calculate new shapes that maintain the largest possible dimension
+            # along each of the contracted axes
+            shape = [min(s, o) for s, o in zip(shape_s_contr, shape_o_contr)]
```
ziofil, line 562: with the change above they should already have the min now

mrmustard/physics/representations.py
```diff
@@ -482,20 +512,71 @@ def __getitem__(self, idx: int | tuple[int, ...]) -> Fock:
...
+            # set same shape along the contracted axes, and default shape along the
+            # axes that are not being contracted
+            shape = [settings.AUTOCUTOFF_MAX_CUTOFF for _ in range(len(other.b[0]))]
+            for i, j in zip(idx_s, idx_o):
+                shape[j] = self.array.shape[1:][i]
+
+            return self[idx_s] @ to_fock(other, shape=shape)[idx_o]
+
+        # the number of batches in self and other
+        n_batches_s = self.array.shape[0]
```
ziofil, line 547: Reminder for self: implement `batch_size` property

ziofil (approved):
Pre-approved! 👍

mrmustard/physics/representations.py
SamFerracin: I don't think that this would work. If `shape[i] < other.array.shape[1:][j]`, then we would have different dimensions along the two axes being contracted (we'd need to mutate that of other)

mrmustard/lab_dev/circuit_components.py
SamFerracin: I changed it into `to_fock_component`, since it also works on components with `Fock` representation, not only bargmann

PR #363, opened 2024-03-11
Using Method Resolution Order (MRO) to return objects of better types.

SamFerracin:
**Context:**
In `lab_dev`, doing `Vacuum(modes=[0]).to_fock_component()` returns an object of type `CircuitComponent`. We want an object of type `Ket` instead.

The `from_attributes` method is modified so that if needed, it scans through the MRO of the class of the given object, and it returns a `Ket`, `DM`, `Unitary`, or `Channel` if possible, a `CircuitComponent` otherwise.

For performance reasons, this function does not validate the inputs and must therefore be used with care. To avoid bugs on the user's side, it is turned into a private method.

### DIFF
--- a/mrmustard/lab_dev/circuit_components.py
+++ b/mrmustard/lab_dev/circuit_components.py
@@ -95,9 +95,9 @@ class CircuitComponent:
                     break
 
         ret = ret()
-        ret._name = name
-        ret._representation = representation
-        ret._wires = wires
+        ret._name = name  # pylint: disable=protected-access
+        ret._representation = representation  # pylint: disable=protected-access
+        ret._wires = wires  # pylint: disable=protected-access
 
         return ret
 
@@ -210,7 +210,7 @@ class CircuitComponent:
             self.name,
             to_fock(self.representation, shape=shape),
             self.wires,
-        )
+        )  # pylint: disable=protected-access
 
     def __eq__(self, other) -> bool:
         r"""
--- a/mrmustard/lab_dev/circuit_components.py
+++ b/mrmustard/lab_dev/circuit_components.py
@@ -93,12 +93,15 @@ class CircuitComponent:
         Returns:
             A circuit component of type ``cls`` with the given attributes.
         """
-        try:
-            ret = cls.mro()[0]()
-        except TypeError:
-            msg = f"Cannot instantiate ``{cls.mro()[0].__name__}`` object from its attributes."
-            raise ValueError(msg)
+        ret = CircuitComponent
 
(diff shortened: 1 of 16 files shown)


mrmustard/lab_dev/circuit_components.py
```diff
@@ -47,40 +47,55 @@ class CircuitComponent:
...
 
         Returns:
-            A circuit component.
+            A circuit component of type ``cls`` with the given attributes.
         """
-        ret = CircuitComponent(name)
-        ret._wires = wires
+        types = {"Ket", "DM", "Unitary", "Channel"}
+        for tp in cls.mro():
+            if tp.__name__ in types:
```
ziofil, line 89: should we also check that the wires are compatible with `tp`?

ziofil: ah no, maybe it doesn't matter

SamFerracin: For how we use it, I think not. It's well documented and hidden by the underscore. But let's have a chat before I merge it

mrmustard/lab_dev/circuit_components.py
```diff
@@ -47,40 +47,55 @@ class CircuitComponent:
...
         Initializes a circuit component from its attributes (a name, a ``Wires``,
         and a ``Representation``).
 
+        If the Method Resolution Order (MRO) of ``cls`` contains one between ``Ket``, ``DM``,
+        ``Unitary``, and ``Channel``, then the returned component is of that type. Otherwise,
+        it is of type ``CircuitComponent``.
+
+        This function needs to be used with caution, as it does not check that the attributes
+        provided are consistent with the type of the returned component. If used improperly it
+        may initialize, e.g., ``Ket``s with both input and output wires or ``Unitary``s with
```
sylviemonet, line 76: Just a small question, "If used improperly it may initialize" means we can always use this function even though the information given are not proper right?

sylviemonet (approved):
I think it is clear for this PR, nice work!

mrmustard/lab_dev/circuit_components.py
SamFerracin: We could do something like:
```
state = Coherent([0], x=1)
unitary = BSgate([0, 1], r=0.1)

Ket._from_state(name="", representation=state.representation, wires=unitary.wires)
```
This would return an object which has the rep of a state and the wires of a unitary.

That's why it needs to be used with caution, and why it should not be shown to the users (with the underscore, it will not appear in the docs, so only devs can see it)

PR #364, opened 2024-03-13
States properties and initializers in lab_dev

SamFerracin:
**Context:**
Adding states properties (e.g. purity) and initializers (e.g. from phasespace) in lab_dev


mrmustard/lab_dev/circuit_components.py
```diff
@@ -247,6 +287,12 @@ def __matmul__(self, other: CircuitComponent) -> CircuitComponent:
 
         return CircuitComponent._from_attributes("", representation_ret, wires_ret)
 
+    def __lshift__(self, other: CircuitComponent):
+        r"""
+        Projects this component onto another component by using self's ``>>`` on ``other.dual``.
+        """
+        return self >> other.dual
```
ziofil, line 294:
```suggestion
        return (other.dual >> self.dual).dual
```

SamFerracin: I actually think that the correct one is: `return (other >> self.dual).dual`

mrmustard/lab_dev/states/base.py
```diff
@@ -86,6 +394,102 @@ def __init__(self, name: Optional[str] = None, modes: Optional[Sequence[int]] =
...
+        ret = Ket(name, modes)
+        ret._representation = Bargmann(*wigner_to_bargmann_psi(cov, means))
+        return ret
+
+    @property
+    def probability(self) -> float:
+        dm = self @ self.dual
+
+        rep = dm.representation
+        msg = "Method ``probability`` not supported for batched representations."
```
SamFerracin, line 474: @ziofil @sylviemonet 
I added these errors (and similar ones to L2) because I believe that there may be an issue with ``trace`` for ``Fock`` objects, when these objects contain batched arrays.

Below I left a minimal working example of the suspected bug. I believe that these should be addressed in a separate PR

```
from mrmustard.lab_dev.states import *
settings.AUTOCUTOFF_MAX_CUTOFF = 10

state = Coherent([0], 1) + Coherent([0], -1)
state.representation.trace([0], [0])  # this is fine

state = Coherent([0], 1) + Coherent([0], -1)
state = state.to_fock_component()
state.representation.trace([0], [0])  # this errors
```

mrmustard/lab_dev/states/base.py
```diff
@@ -86,6 +394,102 @@ def __init__(self, name: Optional[str] = None, modes: Optional[Sequence[int]] =
...
+        else:
+            if rep.A.shape[0] > 1:
+                raise ValueError(msg)
+
+        idx_ket = dm.wires.output.ket.indices
+        idx_bra = dm.wires.output.bra.indices
+
+        traced_rep = dm.representation.trace(idx_ket, idx_bra)
+        ret = traced_rep.c[0] if isinstance(traced_rep, Bargmann) else traced_rep.array
+        return math.atleast_1d(ret, math.float64)[0]
```
SamFerracin, line 487: @sylviemonet I noticed that when one takes the trace of a ``Fock``, it returns a ``np.array`` of shape (0,), such as ``np.ndarray(0.9)``, while the trace of a bargmann retuns a ``c`` with shape (1,), such as ``np.ndarray([0.9])``. This is also a little annoying and we should address it

mrmustard/lab_dev/circuit_components.py
```diff
@@ -209,6 +217,39 @@ def to_fock_component(
...
+        if self.wires != other.wires:
+            msg = "Cannot add components with different wires."
+            raise ValueError(msg)
+        rep = self.representation + other.representation
+        name = self.name if self.name == other.name else ""
+        return self._from_attributes(name, rep, self.wires)
+
+    def __mul__(self, other: Union[CircuitComponent, Scalar]):
+        r"""
+        Implements the multiplication with a scalar or another component for circuit components.
```
sylviemonet, line 233: I'm not familiar with the interpretation of a "CircuitComponent" times a "CircuitComponent".

sylviemonet (approved):
This PR looks fine to me for all the structures for states.

Note that all the methods related to fromX are for Gaussian states. The missing method "from_quadrature" has one question that needs to be discussed between us: we naturally define the Bargamnn representation with ABC Ansatz, Fock representation with Array Ansatz, would we define a new Ansatz for quadrature representation or we use ABC Ansatz directly? If so, I can suppose that we have a ABC form quadrature Gaussian state and then, in the function "from_quadrature",  I'll simply write the basis change transformation [Bargmann contraction] to generate the ABC triples for Bargmann.

mrmustard/lab_dev/circuit_components.py
ziofil: that's a good point. we need it for scalars for sure though.

mrmustard/lab_dev/circuit_components.py
```diff
@@ -247,6 +288,13 @@ def __matmul__(self, other: CircuitComponent) -> CircuitComponent:
 
         return CircuitComponent._from_attributes("", representation_ret, wires_ret)
 
+    def __lshift__(self, other: CircuitComponent):
+        r"""
+        Projects this component onto another component by computing
+        ``(other >> self.dual).dual``.
+        """
+        return (other >> self.dual).dual
```
ziofil, line 296:
```suggestion
        return (other.dual >> self.dual).dual
```
No? E.g. if I say `channel.dual << povm`e.g. to evolve the stabilizers backward like Bill does, in MM it would be `(povm.dual >> channel).dual` which makes sense because `povm.dual` is a dm.

mrmustard/lab_dev/states/base.py
```diff
@@ -36,6 +44,207 @@ class State(CircuitComponent):
...
+            A state.
+
+        Raises:
+            ValueError: If the given array has a shape that is inconsistent with the number of
+                modes.
+        """
+        raise NotImplementedError
+
+    @classmethod
+    def from_phasespace(
```
ziofil, line 128:
```suggestion
    def from_phase_space(
```

mrmustard/lab_dev/states/base.py
```diff
@@ -36,6 +44,207 @@ class State(CircuitComponent):
...
+        cls,
+        modes: Sequence[int],
+        cov: ComplexMatrix,
+        means: ComplexMatrix,
+        name: Optional[str] = None,
```
ziofil, line 133:
```suggestion
        means: ComplexMatrix,
        s: float = 0,
        name: Optional[str] = None,
```

mrmustard/lab_dev/states/base.py
```diff
@@ -36,6 +44,207 @@ class State(CircuitComponent):
...
+    ) -> State:  # pylint: disable=abstract-method
+        r"""
+        Initializes a state from the covariance matrix and the vector of means of a state in
+        phase space.
+
+        Args:
+            cov: The covariance matrix.
+            means: The vector of means.
+            modes: The modes of this states.
+            name: The name of this state.
```
ziofil, line 144:
```suggestion
        Initializes a ``State`` from the covariance matrix and the vector of means in phase space.
        This can take into account the ``s`` parameter of the representation: ``s=0`` (default) is Wigner W,
        ``s=1`` is Husimi Q, ``s=-1`` is Glauber P. Any other value is allowed, although unclear what it means.

        Args:
            cov: The covariance matrix in ``s`` representation.
            means: The vector of means in ``s`` representation.
            modes: The modes spanned by this state.
            s: The parameter of the representation: s=0 (default) is Wigner.
```

mrmustard/lab_dev/states/base.py
```diff
@@ -36,6 +44,207 @@ class State(CircuitComponent):
...
+            ValueError: If the triple cannot be calculated given the state's representation.
+        """
+        rep = self.representation
+        if isinstance(rep, Bargmann):
+            return rep.A, rep.b, rep.c
+        msg = f"Cannot compute triple from representation of type ``{rep.__class__.__name__}``."
+        raise ValueError(msg)
+
+    @property
+    def phasespace(self) -> tuple[ComplexMatrix, ComplexVector]:
```
ziofil, line 185: this one needs to be a function because it needs the s parameter

mrmustard/lab_dev/states/base.py
```diff
@@ -51,6 +260,100 @@ def __init__(self, name: Optional[str] = None, modes: Optional[Sequence[int]] =
...
+    @property
+    def probability(self) -> float:
+        rep = self.representation
+        msg = "Method ``probability`` not supported for batched representations."
+        if isinstance(rep, Fock):
+            if rep.array.shape[0] > 1:
+                raise ValueError(msg)
+        else:
+            if rep.A.shape[0] > 1:
+                raise ValueError(msg)
```
ziofil, line 344:
```suggestion
        if isinstance(rep, Fock) and rep.array.shape[0] > 1:
            raise ValueError(msg)
        elif rep.A.shape[0] > 1:
            raise ValueError(msg)
```
Don't merge, try it first

mrmustard/lab_dev/states/base.py
SamFerracin: We can add them in the future, but for the moment, the functions used for producing `Bargmann` (`wigner_to_bargmann_psi` and `wigner_to_bargmann_rho`) do not allow passing `s`. So I don't know what to do with `s` at the moment :(

mrmustard/lab_dev/circuit_components.py
SamFerracin: But I'm thinking about states. If we do `vacuum << vacuum`, in your case we'd have
- `vacuum.dual >> vacuum.dual`, which errors, because I have to objects with input wires only
In my case, it would be `vacuum >> vacuum.dual`, which is what we want

If the lshift for states a different thing than the one you're mentioning above for channels?

mrmustard/lab_dev/circuit_components.py
SamFerracin: :+1: 
I'll remove it

mrmustard/lab_dev/circuit_components.py
ziofil: But `vacuum << vacuum` shouldn't work, just like `vacuum >> vacuum`. You need to flip the in/out wires on one of them

ziofil: Hmm wait. Am I not using the API correctly?

PR #366, opened 2024-03-15
Fixing bugs/annoying issues and adding functionalities to lab_dev

SamFerracin:
- Fixes the `probability` method for states, so that we can now normalize states with batched representations (i.e., in a superposition).
- Adds `cutoffs` argument to `fock_state` in fock.py, so that the `Number` state can now be initialized with a cutoff.
- Adds `dm` method to `Ket`, to compute the `DM` that corresponds to the given ket
- Uses `astensor` in the init of `PolyExpAnsatz`, so that ``Bargmann``'s A, b, and c can be given as lists and not necessarily as np.arrays.


sylviemonet (approved):
Looks good to me.

Would it be nicer to give more information for probability calculation for ket and dm? Maybe look like:
For ket |psi>, probability = psi >> psi.dual  = <psi|psi>
For dm |psi><psi'|, probability =  Tr(dm) = <psi'|psi>. 
Becaus I was confused when look into the code directly...

PR #367, opened 2024-03-19
Adding simulator

SamFerracin:
**Context:**
Adding simulator for lab_dev

**Possible Drawbacks:**
It does not validate the path. This will be added (in another PR) after the wires PR goes in


mrmustard/lab_dev/simulator.py
```diff
@@ -0,0 +1,112 @@
...
+
+    The ``path`` attribute of ``Circuit``\s allows customising the contraction order and potentially
+    speeding up the simulation. When a ``path`` of the type ``[(i, j), (l, m), ...]`` is given, the
+    simulator creates a dictionary of the type ``{0: c0, ..., N: cN}``, where ``[c0, .., cN]``
+    is the ``circuit.component`` list. Then:
+
+    * The two components ``ci`` and ``cj`` in positions ``i`` and ``j`` are contracted. ``ci`` is
+      replaced by the resulting component ``cj >> cj``, while ``cj`` is popped.
+    * The two components ``cl`` and ``cm`` in positions ``l`` and ``m`` are contracted. ``cl`` is
+      replaced by the resulting component ``cl >> cm``, while ``cl`` is popped.
```
SamFerracin, line 66:
```suggestion
      replaced by the resulting component ``cl >> cm``, while ``cm`` is popped.
```

doc/code/lab_dev/simulator.rst
ziofil, line 1: is this autogenerated?

SamFerracin: No. This needs to be added manually every time a new module is added, courtesy of Sphinx

mrmustard/lab_dev/circuits.py
```diff
@@ -75,6 +76,154 @@ def components(self) -> Sequence[CircuitComponent]:
...
+    @property
+    def path(self) -> list[tuple[int, int]]:
+        r"""
+        A list describing the desired contraction path followed by the ``Simulator``.
+        """
+        return self._path
+
+    @path.setter
+    def path(self, value: list[tuple[int, int]]):
+        self._path = value
```
ziofil, line 88: I know we do this so that the docstring shows up in the documentation, but I'm not too fond of going anti-pythonic and adding getters and setters for no other reason. Is there an alternative way to document attributes?

SamFerracin: @ziofil here the reason (which will become apparent in the next PR) is that the setter will trigger the validation: When the user specifies a path, we want to make sure that it's a valid one, so that we know that the simulator won't fail

mrmustard/lab_dev/simulator.py
```diff
@@ -0,0 +1,112 @@
...
+            circuit.make_path()
+
+        if len(circuit.path) != len(circuit) - 1:
+            msg = f"``circuit.path`` needs to specify {len(circuit) - 1} contractions, "
+            msg += f"found {len(circuit.path)}."
+            raise ValueError(msg)
+
+        ret = dict(enumerate(circuit.components))
+        for idx0, idx1 in circuit.path:
+            ret[idx0] = ret[idx0] >> ret.pop(idx1)
```
ziofil, line 110: nice :)

tests/test_lab_dev/test_simulator.py
```diff
@@ -0,0 +1,69 @@
...
+
+        circuit = Circuit([vac12, d1, d2, d12, a1, n12])
+        circuit.path = path
+
+        res = Simulator().run(circuit)
+        exp = vac12 >> d1 >> d2 >> d12 >> a1 >> n12
+
+        assert res == exp
+
+        settings.AUTOCUTOFF_MAX_CUTOFF = autocutoff_max0
```
ziofil, line 55: maybe set the MIN too so you are sure

mrmustard/lab_dev/circuits.py
ziofil: oh ok, then here it's necessary.

ziofil (approved):
pre-approved!

PR #369, opened 2024-03-20
Better type handling for `rshift`s

SamFerracin:
**Context:**

```
from mrmustard.lab_dev import Coherent

Coherent([0, 1]) >> Coherent([0]).dual
```
At the moment, the code above returns a `CircuitComponent`. We want it to return a `Ket`.


ziofil (approved):
Nice!

PR #370, opened 2024-03-22
To phase space for states

sylviemonet:

**Context:** Add the to_phase_space function for states.

Note that this is branched out from PR368 to use the D_s_map.

**Description of the Change:** This function receives the state and changes the representation of a given mode.

Questions left:
- [ ]  What would be the best returns? If this is a N-mode state, the state has been transformed totally into phase space, would it be stored in the phase space representation somehow? If the state is just partially transformed into another representation, how to label it?

**Benefits:**

**Possible Drawbacks:**

**Related GitHub Issues:**

### DIFF
--- a/mrmustard/physics/converters.py
+++ b/mrmustard/physics/converters.py
@@ -21,7 +21,7 @@ from mrmustard import math, settings
 from mrmustard.utils.typing import Matrix, Vector
 from mrmustard.physics.triples import displacement_map_s_parametrized_Abc
 from mrmustard.physics.bargmann import complex_gaussian_integral, join_Abc
-from mrmustard.lab_dev.states import State
+from mrmustard.lab_dev.states import State, Ket, DM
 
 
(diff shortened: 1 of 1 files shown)


mrmustard/physics/converters.py
```diff
@@ -59,3 +63,51 @@ def to_fock(rep: Representation, shape: Optional[Union[int, Iterable[int]]] = No
         array = [math.hermite_renormalized(A, b, c, shape) for A, b, c in zip(rep.A, rep.b, rep.c)]
         return Fock(math.astensor(array), batched=True)
     return rep
+
+
+def to_phase_space(
```
SamFerracin, line 68: This is circular import (we cannot import `State` from lab in physics). 
I think that this logic should be moved to the (existing, not implemented) `to_phase_space`) function of `State`

sylviemonet:
I'll close this PR since this is the work after in states, unitaries or channels. My work needs to finish the triple for the transformation gate first. I'll open a new one to add the triples gate from_quad and to_quad.

PR #374, opened 2024-03-27
Logic to validate `circuit.path`

SamFerracin:
**Context:**
This PR adds logic to validate `circuit.path`. The path is now automatically (and efficiently) validated when a user sets it, making sure that simulations are not interrupted by path-related errors.


ziofil (approved):
👍

PR #377, opened 2024-03-28
silenced complex det warnings

ziofil:
**Context:**
Numpy now raises warnings when the determinant of a complex matrix is calculated. We silence the warnings only when those determinants are calculated.

Wrap det calls in a numpy context

**Benefits:**
No more warnings

**Possible Drawbacks:**
Warnings may be legit

**Related GitHub Issues:**


SamFerracin:
Cool. Just a question:
We have a class to silence warnings, that lets you un-silence them if you want to (may be useful for users to debug). It's in `utils/filters.py`. Do we want to use that, as THE way of silencing warnings?

ziofil:
I forgot we had that! But I don't like that we would silence numerical warnings for the whole kernel session. They may be relevant. Check the current solution: it's only when we compute the det in the numpy backend and a test will fail in the future when there's no need for the filter.

ziofil:
The tests fail here and pass on my machine. I think it's only my problem then? 🤔

ziofil:
closed by mistake

PR #378, opened 2024-04-01
Adding more convenience states and transformations to lab_dev

SamFerracin:
**Context:**
Adding more convenience states and transformations to lab_dev, getting closer to replicating all those that we have in `lab` currently


mrmustard/lab_dev/states/states.py
```diff
@@ -196,5 +306,48 @@ def __init__(
...
+        n_modes = len(self.modes)
+        return Bargmann(*triples.vacuum_state_Abc(n_modes))
+
+
+#  ~~~~~~~~~~~~
+#  Mixed States
+#  ~~~~~~~~~~~~
+
+
+class Thermal(Ket):
```
ziofil, line 318: Thermal is `DM`

tests/test_lab_dev/test_states.py
```diff
@@ -399,6 +507,36 @@ def test_representation(self, n_modes):
         assert math.allclose(rep.c, [1.0])
 
 
+class TestThermal:
+    r"""
+    Tests for the ``Number`` class.
```
ziofil, line 512:
```suggestion
    Tests for the ``Thermal`` class.
```

mrmustard/lab_dev/states/states.py
ziofil, line 318:
```suggestion
class Thermal(DM):
```

tests/test_lab_dev/test_states.py
```diff
@@ -399,6 +507,36 @@ def test_representation(self, n_modes):
         assert math.allclose(rep.c, [1.0])
 
 
+class TestThermal:
```
ziofil, line 510: Here we are not testing the physical properties of the thermal state, perhaps we should? Or shall we do that in integration tests?

ziofil (approved):
Approved, but we should revisit the tests in the future, maybe when we tear down lab.

tests/test_lab_dev/test_states.py
SamFerracin: That is a good question.
Those properties are inherited from `DM`, therefore they are tested in `TestDM` *for some states*.
They may not be tested *for all* the built-in states, but I don't see that as a criticality, because they are agnostic to the state (i.e., they take a representation and do some math, without needing to know the class of the object).

But if you are concerned with things being undertested in this way, I am always up for adding more tests!
And yeah, we should start talking about integration tests, in preparation for Q2

Issue #381, opened 2024-04-08
TensorFlow dependecy needs to be fixed <2.16.0


xvalcarce:
TensorFlow `>=2.16.0` has been released a month ago. Installing MrMustard with `pip install mrmustard` will install TensorFlow `2.16.1` by default. This is problematic as new releases of TensorFlow dropped support for Keras 2.0 -- see the [release note](https://github.com/tensorflow/tensorflow/releases/tag/v2.16.1). Therefore, the following call

https://github.com/XanaduAI/MrMustard/blob/d31b70f5d68d36d471858c7cffe2382b5d200f55/mrmustard/math/backend_tensorflow.py#L403

will trigger an error. For a MWE, the error can easily be triggered with the optimization given in the README 

https://github.com/XanaduAI/MrMustard/blob/d31b70f5d68d36d471858c7cffe2382b5d200f55/README.md?plain=1#L252-L285

As a temporary fix, I suggest to replace `^2.15.0` with `~2.15.0` in the following
https://github.com/XanaduAI/MrMustard/blob/d31b70f5d68d36d471858c7cffe2382b5d200f55/pyproject.toml#L52-L59

ziofil:
Thank you, nice find! Would you like to open a PR and add your name to the contributors for the next release?

xvalcarce:
Hey @ziofil , thanks for the opportunity, I opened a PR with the quick-fix as you suggested!

timmysilv:
Hi @xvalcarce! Thanks for putting this all together, super detailed and made things easy. We didn't want to force a lower TensorFlow version upon our users, so I just merged a PR to add support for TensorFlow 2.16 - you can check out the details in the PR above, but the tl;dr is MrMustard now uses the non-legacy version of the Adam optimizer when you have TF 2.16+ installed. Lmk if you have any questions/concerns about things

PR #382, opened 2024-04-10
fix: TensorFlow dependency set to ~2.15

xvalcarce:
**Context:**

See issue https://github.com/XanaduAI/MrMustard/issues/381

Restrict TensorFlow dependency to >=2.15 and <2.16.


SamFerracin:
Hi @xvalcarce ,

Thanks for opening this issue up! :) 

Personally, I am not convinced that TF should be restricted to version `< x`, because in the future we will certainly want to install `y > x`.

The solution to your issue may be using `poetry` rather than `pip install`. In your environment, do
```
pip install poetry
cd MrMustard
poetry install ...
```
optionally replace `...` with options, like we do in our `.github/workflows`. 

TF is locked to 2.15, so poetry should install 2.15

xvalcarce:
Hey @SamFerracin :)

I agree, this is not a long term viable solution, but a simple quick-fix! Instead the codebase should be updated to match TF 2.16, and set this new version as a minimum requirement. This mainly involves using  Adam from Keras 3 (as mentioned in the issue) as well as removing the need for `tensorflow-macos` as explained [here](https://github.com/tensorflow/tensorflow/releases/tag/v2.16.1). I can propose these changes in a future commit, if you are interested!

About the installation method, if an installation with cloning the repo and using `poetry` works fine thanks to the lock, I do believe that installing MrMustard with `pip install mrmustard` should provide a working module.

ziofil:
Fixed in #406. Thanks again @xvalcarce :)

PR #387, opened 2024-04-17
fix inheritance diagrams and make install

timmysilv:
- `Makefile` syntax requires tabs, not spaces, so this `make` target didn't work
- Sphinx inheritance diagrams require the RTD builder machine to have `graphviz` installed. Otherwise, [it doesn't work](https://mrmustard.readthedocs.io/en/stable/code/training.html#class-inheritance-diagram)
  - it's [fixed](https://mrmustard--387.org.readthedocs.build/en/387/code/training.html#class-inheritance-diagram) by this PR
- run numpy and tensorflow tests unconditionally

### DIFF
--- a/.github/workflows/tests_numpy.yml
+++ b/.github/workflows/tests_numpy.yml
@@ -4,14 +4,6 @@ on:
     branches:
       - develop
   pull_request:
-    paths:
-      - '.github/workflows/tests.yml'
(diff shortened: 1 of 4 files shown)


SamFerracin (approved):
Thanks, @timmysilv. I wanted to fix it for a while, but never got time to investigate

PR #389, opened 2024-04-20
Transformation repr methods

ziofil:
Fixed a few bugs and added stub methods for `Transformation`.

### DIFF
--- a/mrmustard/lab_dev/circuit_components.py
+++ b/mrmustard/lab_dev/circuit_components.py
@@ -73,12 +73,7 @@ class CircuitComponent:
         ib = tuple(sorted(modes_in_bra))
         ok = tuple(sorted(modes_out_ket))
         ik = tuple(sorted(modes_in_ket))
-        if (
-            ob != modes_out_bra
-            or ib != modes_in_bra
-            or ok != modes_out_ket
-            or ik != modes_in_ket
-        ):
+        if ob != modes_out_bra or ib != modes_in_bra or ok != modes_out_ket or ik != modes_in_ket:
             offsets = [len(ob), len(ob) + len(ib), len(ob) + len(ib) + len(ok)]
             perm = (
                 tuple(np.argsort(modes_out_bra))
@@ -349,14 +344,10 @@ class CircuitComponent:
         idx_zconj += other.wires.ket.input[ket_modes].indices
 
         # calculate the representation of the returned component
-        representation_ret = (
-            self.representation[idx_z] @ other.representation[idx_zconj]
-        )
+        representation_ret = self.representation[idx_z] @ other.representation[idx_zconj]
 
         # reorder the representation
-        representation_ret = (
-            representation_ret.reorder(perm) if perm else representation_ret
-        )
+        representation_ret = representation_ret.reorder(perm) if perm else representation_ret
         return CircuitComponent._from_attributes(None, representation_ret, wires_ret)
 
     def __rshift__(self, other: CircuitComponent) -> CircuitComponent:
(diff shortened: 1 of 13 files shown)


mrmustard/lab_dev/circuit_components.py
```diff
@@ -421,6 +447,10 @@ class DualView(CircuitComponent):
         component: The circuit component to take the view of.
     """
 
+    def __getattr__(self, name):
+        r"""send calls to the component"""
+        return getattr(self._component, name)
```
SamFerracin, line 452: :+1:

SamFerracin: In retrospective, maybe it would have been better to have a `View` class that adjoint and dual can inherit from, to avoid repeated code

mrmustard/lab_dev/transformations/base.py
```diff
@@ -36,6 +40,25 @@ class Transformation(CircuitComponent):
...
 
+    def inverse(self) -> Transformation:
+        r"""Returns the inverse of the transformation."""
+        if not isinstance(self.representation, Bargmann):
+            raise NotImplementedError("Only Bargmann representation is supported.")
+        if self.representation.ansatz.batch_size > 1:
+            raise NotImplementedError("Batched transformations are not supported.")
+        A, b, _ = self.dual.representation.conj().triple  # apply X
+        almost_inverse = self.__class__.from_bargmann(
+            [0], (math.inv(A[0]), -math.inv(A[0]) @ b[0], 1 + 0j)
```
sylviemonet, line 51: Can we somehow know the modes of this Transformation directly rather than use [0]?

mrmustard/lab_dev/transformations/base.py
```diff
@@ -36,6 +40,25 @@ class Transformation(CircuitComponent):
     Base class for all transformations.
     """
 
+    def inverse(self) -> Transformation:
+        r"""Returns the inverse of the transformation."""
+        if not isinstance(self.representation, Bargmann):
+            raise NotImplementedError("Only Bargmann representation is supported.")
```
sylviemonet, line 46: Only single mode Bargmann is supported here, am I correct?

mrmustard/lab_dev/utils.py
```diff
@@ -72,3 +73,13 @@ def reshape_params(n_modes: str, **kwargs) -> Generator:
                 msg = f"Parameter {names[i]} has an incompatible shape."
                 raise ValueError(msg)
         yield var
+
+
+def _shape_check(mat, vec, dim: int, name: str):
```
SamFerracin, line 78: No reason to keep it private. i suggest removing the underscore

mrmustard/lab_dev/utils.py
```diff
@@ -72,3 +73,13 @@ def reshape_params(n_modes: str, **kwargs) -> Generator:
...
+    r"""
+    Check that the given Gaussian representation is consistent with the given modes.
```
SamFerracin, line 80: `Args:` and `Returns:`

sylviemonet (commented):
Looks good to me in the physics side, except for some small by default settings.

mrmustard/lab_dev/circuit_components.py
ziofil: good idea. I'll put it in the todos

mrmustard/lab_dev/transformations/base.py
ziofil: [0] is the first and only element of the batch

mrmustard/lab_dev/transformations/base.py
ziofil: no, any number of modes

SamFerracin (approved):

PR #391, opened 2024-04-25
Batched methods + simplified settings

ziofil:
- L2_norm, purity, probability can now work element-wise along the batch dimension or globally. This enables using the batch dimension more freely.

- simplified docstrings in Settings class

- adds setting `CIRCUIT_DRAW_PARAMS` to set whether the circuit drawer should just draw the name of components or their parameters as well

### DIFF
--- a/mrmustard/lab_dev/states/base.py
+++ b/mrmustard/lab_dev/states/base.py
@@ -61,7 +61,6 @@ class State(CircuitComponent):
         modes: Sequence[int],
         triple: tuple[ComplexMatrix, ComplexVector, complex],
         name: Optional[str] = None,
-        batched: bool = False,
     ) -> State:
         r"""
         Initializes a state from an ``(A, b, c)`` triple defining a Bargmann representation.
@@ -84,7 +83,6 @@ class State(CircuitComponent):
             modes: The modes of this states.
             triple: The ``(A, b, c)`` triple.
             name: The name of this state.
-            batched: Whether the given triple is batched.
 
         Returns:
             A state.
@@ -223,9 +221,7 @@ class State(CircuitComponent):
         """
         return math.allclose(self.purity, 1.0)
 
-    def fock_array(
-        self, shape: Optional[Union[int, Sequence[int]]] = None
-    ) -> ComplexTensor:
+    def fock_array(self, shape: Optional[Union[int, Sequence[int]]] = None) -> ComplexTensor:
         r"""
         The array that describes this state in the Fock representation.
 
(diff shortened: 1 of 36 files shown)


mrmustard/lab_dev/states/base.py
```diff
@@ -690,19 +683,31 @@ def from_phase_space(
...
+        return math.real(rep.array)
+
+    @property
+    def probability(self) -> float:
+        r"""Probability of this DM, using the batch dimension of the Ansatz
+        as a convex combination of states."""
+        return math.sum(self._probabilities)
+
+    @property
+    def _purities(self) -> RealVector:
```
SamFerracin, line 703: why do all these properties need to be private?

ziofil: they don't, I'll change it

mrmustard/lab_dev/states/base.py
```diff
@@ -925,12 +968,21 @@ def from_quadrature(
...
-        if isinstance(rep, Bargmann):
-            return math.real(math.sum(rep.c, axes=[0]))
-        return math.real(math.sum(rep.array, axes=[0]))
+        r"""Probability of this state, where the batch dimension of the Ansatz
+        means a linear combination of states."""
+        return self.L2_norm
+
+    def purities(self) -> float:
+        r"""Purity of this state, where the batch dimension of the Ansatz
+        means a linear combination of states."""
```
ziofil, line 984:
```suggestion
        r"""Purity of each state in the batch."""
```

SamFerracin (approved):

mrmustard/physics/gaussian_integrals.py
```diff
@@ -131,20 +132,19 @@ def complex_gaussian_integral(
...
         A_post = R - math.matmul(D, math.inv(M), math.transpose(D))
         b_post = bR - math.matvec(D, math.solve(M, bM))
     else:
-        A_post = math.astensor([])
-        b_post = math.astensor([])
-
-    c_post = (
-        c * math.sqrt((-1) ** n / math.det(M)) * math.exp(-0.5 * math.sum(bM * math.solve(M, bM)))
-    )
+        A_post = math.zeros((0, 0), dtype=A.dtype)
```
sylviemonet, line 146: Why it is not empty but instead with (0,0)?

mrmustard/physics/representations.py
```diff
@@ -451,9 +451,17 @@ def __matmul__(self, other: Union[Bargmann, Fock]) -> Union[Bargmann, Fock]:
             )
 
         Abc = []
-        for A1, b1, c1 in zip(self.A, self.b, self.c):
-            for A2, b2, c2 in zip(other.A, other.b, other.c):
+        if settings.ELEMENT_WISE:
+            if self.ansatz.batch_size != other.ansatz.batch_size:
+                raise ValueError("Batch size of the two ansatze must match.")
```
sylviemonet, line 456:
```suggestion
                raise ValueError(f"Batch size of the two ansatze must match since the settings.ELEMENT_WISE is  {settings.ELEMENT_WISE}.")
```

sylviemonet (commented):
I'm ok with most of the changes in PR. Just a concern about the setting.ELEMENT_WISE, this is a label to say if we want element wise __matmul__ of the representation or not. But it is still a problem if we use batch to interpret everything in MM. For example, if we do state@measurement, if this settings would be a thing there?

SamFerracin:
> I'm ok with most of the changes in PR. Just a concern about the setting.ELEMENT_WISE, this is a label to say if we want element wise **matmul** of the representation or not. But it is still a problem if we use batch to interpret everything in MM. For example, if we do state@measurement, if this settings would be a thing there?

I do agree with @sylviemonet that these batched methods are rushed and should be thought over in the larger context of things. The risk is ending up with too many functionalities that users use in ways that were not expected, and are incompatible with each other (see the example about detectors). My advice is to encapsulate the concept of batch in its own class, as opposed to leaving it floating around and up to the interpretation of the user/developer. Specifically, it should be made clear what axis represent different "superposition" batches, and what axis represent different "zip" batches. The setting (if needed) should come after that

mrmustard/physics/gaussian_integrals.py
ziofil: so that it sill has the right len(shape)

ziofil:
@sylviemonet if we do state@measurement with `ELEMENT_WISE = True` we mean that we zip the batch dimensions of state and measurement, which will have to match. If it is False (default), then it's the kron of the batch dimensions as usual. I think this is the easiest way to implement both functionalities and it's a simple switch that you can turn on and off. @SamFerracin what example about detectors? if you have an example of how this idea could fail please let me know. For now we keep it as it's already working as it should.

ziofil:
Actually, I changed my mind. You're both right, and this is probably the solution (which was thrown around in the past) we keep a batch dimension that we *always* zip over and that grows when we concatenate batches, and a linear combination dimension that we kron over when we multiply and append to when we sum.

PR #393, opened 2024-05-06
Adding html representations for `Wires`, `Representation`s, and `CircuitComponent`

SamFerracin:
**Context:**
Adding html representations for `Wires`, `Representation`s, and `CircuitComponent`


sylviemonet (commented):
@ziofil, I have no questions on this PR.

ziofil (approved):
I flipped ket/bra. There are some aspect ratio issues, but we can address them later when we fix the aspect ratio of the single-mode state plot.

PR #394, opened 2024-05-06
use the juliacall package instead of julia

timmysilv:
**Context:**
The `julia` python package is backed by `PyCall`, and `PyCall` kinda sucks. `juliacall`/`pythoncall` is a much easier package to work with, with very simple [single-file configuration](https://juliapy.github.io/PythonCall.jl/stable/juliacall/#julia-deps)

Replace uses of the `julia` python package with `juliacall` and do a bit of re-naming/tidying.

**Benefits:**
- `juliacall` requires much less configuration up-front
- if the required julia version from the config (~1.9.3 today) can't be found, it will automatically install a valid one at runtime using juliaup. Then, if the julia project hasn't been initialized (equivalent to the `Pkg.instantiate("julia_pkg")` thing we were doing before), it will do that automatically as well!
- tl;dr the user need not do any real installation stuff anymore

**Possible Drawbacks:**
- ~I do not know if performance is worse. It might even be better?~ If a reviewer could help me prove this, that would be fantastic. That said, the creator of PyCall himself [seems to like PythonCall](https://discourse.julialang.org/t/ann-pythoncall-and-juliacall/76778/24). @ziofil sent me a pretty simple notebook to use to confirm things, and [the results](https://www.notion.so/xanaduai/Use-juliacall-in-MrMustard-72b027350b91476aaff57222820cc072) suggest that the performance is basically the same, but we might save a smidge with some import overhead, idk.
- I have to manually call `self.astensor` because for some reason, there is no auto-conversion happening from Julia to Python although it should theoretically happen with `juliacall`. See note on memory below.

**Notes on memory consumption**
Julia arrays returned to Python are wrapped in a Python object called an `ArrayValue`. According to [the FAQ](https://juliapy.github.io/PythonCall.jl/stable/faq/#Issues-when-Numpy-arrays-are-expected), the object satisfies the Python array protocol, but is not _actually_ a NumPy array. If needed (it sometimes is), the class has a `to_numpy` function, and its default value for `copy` [is True](https://github.com/JuliaPy/PythonCall.jl/blob/aca2701e472245e79397d33bb0e962aa33d9fc6c/src/JlWrap/array.jl#L338). If we don't intend on mutating array values, we can set this to `False` and just read the values from Julia's allocated memory. The performance trade-offs would require extensive benchmarking to be sure which is better.


ziofil (approved):
It works! 💯

timmysilv:
I just added [a hack](https://github.com/XanaduAI/MrMustard/pull/394/commits/73dd5983ae16f6d46c217c9bd0418ca37363bf48) to suppress the julia outputs that happen the first time you import it (in other words, the first time you increase the precision). If we don't mind the outputs happening in a fresh environment once, then I think we should revert that change. lmk what you think!

PR #395, opened 2024-05-07
Fixing codecoverage

SamFerracin:
**Context:**
A few open PRs are failing because of indirect changes in the coverage of three files, all of which are unrelated to the given PRs.
Here, I either add tests (where I know how to test) or `# pragma: no cover` to the problematic lines


mrmustard/training/trainer.py
timmysilv, line 1: is it too challenging to test these non-covered lines?

SamFerracin: I have never worked on the training package, so it would take me a bit to figure out how to do that. @ziofil should be able to add tests quickly though

ziofil: If I recall correctly that was one of @zeyueN's contributions?

zeyueN:
I think codecov is wrong....

Tried running tests locally with `print`s inserted where codecov claims to be not covered, it is actually hitting those lines.

SamFerracin:
> I think codecov is wrong....
> 
> Tried running tests locally with `print`s inserted where codecov claims to be not covered, it is actually hitting those lines.

Mmmh interesting..
So what would you suggest we do? We could force this in with `#pragma no cover`?

ziofil (approved):
👍

PR #397, opened 2024-05-09
Autoshape and fock improvements

ziofil:
**Context:**
A recommended Fock shape of the components of a circuit can often be determined by the nature of the components and their connectivity. Also, for certain components like `States` we can determine a good choice of shape by stopping e.g. at 99.9% of the norm.

- `CircuitComponent`s now have the property `.manual_shape` which can be set manually. It starts as `[None,...]` and it can be modified like e.g. `my_state.manual_shape[1]=9` and now it will be `[None, 9, None, ...]`.
- `CircuitComponent`s now have the method `.auto_shape()` which computes and returns a tuple with a guesstimate of the recommended shape to the best of our knowledge. It will respect the values in `.manual_shape` that have already been set.
- `Circuit` now has the method `propagate_shapes()` which exploits the context of each component to produce a more compact shape than `auto_shape()` would give in isolation. This gives the `Circuit` the power to contract an arrangement of components in the `Fock` representation in a close to optimal way (excluding throwing SVDs around), when combined with the optimal contraction path (upcoming).
- Faster `autocutoff` method (now `autoshape_numba`). It doesn't any longer travel back and forth between phase space and Bargmann for each mode, and it's the same as Robbe's diagonal method, but specialized for this task, so it's faster.
- General "fock-related" functionality has been refactored/improved
    - removed the converter module
    - localized the fock conversion into the `CircuitComponent` so that wire information is available
    - components keep the original bargmann data when they are converted to fock
    - can no longer matmul across representations (this is good)


874_connections.py
```diff
@@ -0,0 +1,13 @@
+from mrmustard.lab_dev.opt_contraction import optimal_path, staircase
```
SamFerracin, line 1: delete this file?

ziofil: yep!

mrmustard/lab_dev/circuit_components.py
```diff
@@ -215,6 +215,7 @@ def from_quadrature(
         name: Optional[str] = None,
     ) -> CircuitComponent:
         r"""Returns a circuit component from the given triple (A,b,c) that parametrizes the
+        quadrature wavefunction of this component in the form ``c exp(1/2 x^T A x + b^T x)``.
```
ziofil, line 218:
```suggestion
```

mrmustard/lab_dev/circuit_components.py
```diff
@@ -318,6 +318,25 @@ def dual(self) -> DualView:
...
+        r"""
+        The shape of this Component in the Fock representation. If not manually set,
+        it is a list of M ``None``s where M is the number of wires of the component.
+        The custom_shape is a list and therefore it is mutable. In fact, it can evolve
+        over time as we learn more about the component or its neighbours. For
+        each wire, the entry is either an integer or ``None``. If it is an integer, it
+        is the dimension of the corresponding Fock space. If it is ``None``, it means
+        the best shape is not known yet. ``None``s automatically become integers when
+        ``auto_shape`` is called, but the integers already set are not changed.
+        """
```
ziofil, line 332:
```suggestion
        ``auto_shape`` is called, but the integers already set are not changed.
        The order of the elements in the shape is intended the same order as the wires
        in the `.wires` attribute.
        """
```

mrmustard/lab_dev/circuit_components.py
```diff
@@ -371,35 +392,68 @@ def on(self, modes: Sequence[int]) -> CircuitComponent:
...
+        available, otherwise it defaults to the value of ``AUTOCUTOFF_MAX_CUTOFF`` in the settings.
+
+        Args:
+            shape: The shape of the returned representation. If ``shape`` is given as an ``int``,
+                it is broadcasted to all the dimensions. If not given, it is estimated.
+        Returns:
+            array: The Fock representation of this component.
+        """
+        if shape is None:
+            shape = self.auto_shape
```
ziofil, line 408:
```suggestion
        shape = shape or self.auto_shape
```

mrmustard/lab_dev/circuit_components.py
```diff
@@ -371,35 +392,67 @@ def on(self, modes: Sequence[int]) -> CircuitComponent:
...
+        available, otherwise it defaults to the value of ``AUTOCUTOFF_MAX_CUTOFF`` in the settings.
+
+        Args:
+            shape: The shape of the returned representation. If ``shape`` is given as an ``int``,
+                it is broadcasted to all the dimensions. If not given, it is estimated.
+        Returns:
+            array: The Fock representation of this component.
+        """
+        shape = shape or self.auto_shape
+        elif isinstance(shape, int):
```
ziofil, line 408:
```suggestion
        shape = shape or self.auto_shape
        if isinstance(shape, int):
```

mrmustard/lab_dev/circuit_components.py
```diff
@@ -371,35 +392,67 @@ def on(self, modes: Sequence[int]) -> CircuitComponent:
 
         return ret
 
-    def to_fock(self, shape: Optional[Union[int, Iterable[int]]] = None) -> CircuitComponent:
+    def fock(self, shape: Optional[Union[int, Sequence[int]]] = None) -> CircuitComponent:
```
ziofil, line 395:
```suggestion
    def fock(self, shape: Optional[int | Sequence[int]] = None) -> CircuitComponent:
```

mrmustard/lab_dev/states/states.py
```diff
@@ -221,6 +221,8 @@ def __init__(
             msg = f"Length of ``cutoffs`` must be 1 or {len(modes)}, found {len(self._cutoffs)}."
             raise ValueError(msg)
 
+        self._custom_shape = list(c + 1 for c in self.cutoffs)
```
ziofil, line 224:
```suggestion
        self._custom_shape = list(c + 1 for c in self.n)
```

mrmustard/math/lattice/strategies/vanilla.py
```diff
@@ -188,3 +190,139 @@ def vanilla_vjp(G, c, dLdG) -> tuple[ComplexMatrix, ComplexVector, complex]:  #
     dLdc = np.sum(G_lin.reshape(shape) * dLdG) / c
 
     return dLdA, dLdb, dLdc
+
+
+# @njit
```
ziofil, line 195:
```suggestion
```

mrmustard/math/lattice/strategies/vanilla.py
```diff
@@ -188,3 +190,139 @@ def vanilla_vjp(G, c, dLdG) -> tuple[ComplexMatrix, ComplexVector, complex]:  #
...
+        b_ = b_m - A_mn @ np.linalg.inv(A_nn - X) @ b_n
+        c_ = (
+            c
+            * np.exp(-0.5 * b_n @ np.linalg.inv(A_nn - X) @ b_n)
+            / np.sqrt(np.linalg.det(A_nn - X))
+        )
+        yield A_, b_, c_
+
+
+# @njit
```
ziofil, line 226:
```suggestion
```

ziofil, line 226:
```suggestion
@njit
```

mrmustard/physics/ansatze.py
```diff
@@ -435,10 +435,10 @@ def __eq__(self, other: Ansatz) -> bool:
         Raises:
             ValueError: If the arrays don't have the same shape.
```
ziofil, line 436:
```suggestion
```

mrmustard/math/lattice/strategies/vanilla.py
timmysilv, line 1: do we want to render these docs?

mrmustard/math/lattice/strategies/vanilla.py
```diff
@@ -188,3 +190,130 @@ def vanilla_vjp(G, c, dLdG) -> tuple[ComplexMatrix, ComplexVector, complex]:  #
...
+    """
+    # reduced DMs
+    M = len(b) // 2
+    shape = np.ones(M, dtype=np.int64)
+    A = A.reshape((2, M, 2, M)).transpose((1, 3, 0, 2))  # (M,M,2,2)
+    b = b.reshape((2, M)).transpose()  # (M,2)
+    zero = np.zeros((M - 1, M - 1), dtype=np.complex128)
+    id = np.eye(M - 1, dtype=np.complex128)
+    X = np.vstack((np.hstack((zero, id)), np.hstack((id, zero))))
+    for t, m in enumerate(range(M)):
```
timmysilv, line 276: aren't `t` and `m` the same here?

mrmustard/physics/representations.py
```diff
@@ -636,34 +627,32 @@ def __getitem__(self, idx: int | tuple[int, ...]) -> Fock:
...
         r"""
-        Implements the inner product of ansatze across the marked indices.
-
-        If ``other`` is ``Fock``, the two representations are automatically reduced
-        before being contracted. The array of the returned representation has the largest
-        possible dimension along every axis.
-
-        If ``other`` is ``Bargmann``, it is converted to ``Fock`` before the contraction.
+        Implements the inner product of fock arrays over the marked indices.
+        ..code-block::
```
timmysilv, line 631: doesn't really matter because it doesn't render, but just in case it gets moved one day
```suggestion
        Implements the inner product of fock arrays over the marked indices.

        .. code-block::
```

mrmustard/lab_dev/circuit_components.py
```diff
@@ -552,7 +646,7 @@ def __rrshift__(self, other: Scalar) -> CircuitComponent | np.array:
             return ret
 
     def __repr__(self) -> str:
-        return f"{self.__class__.__name__}(modes={self.modes}, name={self.name or None})"
+        return f"{self.__class__.__name__}(modes={self.modes}, name={self.name or None}, repr={self.representation.__class__.__name__})"
```
timmysilv, line 649: can we only add this if the representation is defined? the tests show `repr=NoneType` which is kinda noisy for users

mrmustard/lab_dev/states/base.py
```diff
@@ -648,6 +631,29 @@ def __init__(
         if representation is not None:
             self._representation = representation
 
+    def auto_shape(self) -> tuple[int, ...]:
```
timmysilv, line 634: I was surprised to see that this modifies the fock_shape of this object. There seem to be a lot of side effects throughout MrMustard, and I'm worried that they'll be hard for users to keep track of. Not for this PR, but I think it would be worth spending some time documenting (somewhere) what they all are, particularly in lab_dev before switching it over to lab.

timmysilv (commented):
will be a great feature-add! might need a more hands-on explanation of that auto-shaping algorithm 😅

aplund:
This feels like I'm just not getting it, but should `.auto_shape()` have the side-effect of setting the fock_shape attribute?  In my basic test below, I just get `[None]`:

```
import mrmustard as mm
from mrmustard.lab_dev import *

cc = Number([0], 4) >> Sgate([0], 1)
cc.auto_shape()
print(cc.fock_shape)
```

It seems that `.auto_shape()` in this instance returns `(100,)`.  Is this right?

timmysilv:
nice find @aplund! looks like for DM and Ket implementations of `auto_shape()`, we only update the object's `fock_shape` if it has a Bargmann representation. This is a good example of my above hesitation with side effects (weirdness here is that they are only sometimes applied) and why I'd prefer to avoid it

mrmustard/lab_dev/circuit_components.py
```diff
@@ -320,6 +320,36 @@ def dual(self) -> DualView:
         """
         return DualView(self)
 
+    @property
```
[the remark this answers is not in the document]
ziofil: Oh thanks, I used cached_property [on occasion](https://github.com/XanaduAI/MrMustard/blob/develop/mrmustard/lab_dev/wires.py). I'll try.

mrmustard/lab_dev/circuit_components.py
```diff
@@ -320,6 +320,36 @@ def dual(self) -> DualView:
...
+        return self._fock_shape
+
+    @fock_shape.setter
+    def fock_shape(self, shape: list[Optional[int]]):
+        r"""
+        Sets the custom shape of this component in the Fock representation.
+        """
+        if len(shape) > len(self.wires):
+            raise ValueError(f"expected a shape of length {len(self.wires)}, got {len(shape)}")
+        self._fock_shape = shape
```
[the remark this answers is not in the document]
ziofil: No, that's not the purpose of it. In fact it should be mutable. Have you read the docstring?

mrmustard/lab_dev/circuit_components.py
```diff
@@ -373,35 +403,65 @@ def on(self, modes: Sequence[int]) -> CircuitComponent:
...
+        Args:
+            shape: The shape of the returned representation. If ``shape`` is given as an ``int``,
+                it is broadcasted to all the dimensions. If not given, it is estimated.
+        Returns:
+            array: The Fock representation of this component.
+        """
+        if isinstance(shape, int):
+            shape = (shape,) * self.representation.ansatz.num_vars
+        shape = shape or self.auto_shape()
+        assert len(shape) == len(self.fock_shape)
```
ziofil:
```suggestion
        if len(shape) != len(self.fock_shape):
            raise ValueError(f"Expected Fock shape of length {len(self.fock_shape)}, got length {len(shape)}")
```

mrmustard/lab_dev/circuit_components.py
```diff
@@ -373,35 +403,65 @@ def on(self, modes: Sequence[int]) -> CircuitComponent:
...
+        try:
+            As, bs, cs = self.bargmann
+            array = [math.hermite_renormalized(A, b, c, shape) for A, b, c in zip(As, bs, cs)]
+        except AttributeError:
```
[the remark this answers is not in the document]
ziofil: Not quite, because all `CircuitComponent`s have a `.bargmann` property. The error is raised within the property if the representation has no bargmann triple.

mrmustard/lab_dev/circuit_components.py
```diff
@@ -373,35 +403,65 @@ def on(self, modes: Sequence[int]) -> CircuitComponent:
...
+        return self._from_attributes(fock, self.wires, self.name)
+
+    def auto_shape(self) -> tuple[int, ...]:
+        r"""
+        The shape of the Fock representation of this component. If the component has a Fock representation
+        then it is just the shape of the array. If the components is a State in Bargmann
+        representation the shape is calculated using autoshape using the single-mode marginals.
+        If the component is not a State then the shape is a tuple of ``AUTOSHAPE_MAX``.
+        """
+        MAX = settings.AUTOSHAPE_MAX
```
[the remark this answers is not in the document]
ziofil: Well, the thing is that we need a full shape in order to create the array for the Fock representation. So all the `None`s in `fock_shape` should be replaced by some integer. If we do not know any better (because of the properties of the component, or because of its neighbours if we are in a circuit) then we have to pick _some_ default value, and this is the one. It's not good nor bad 😅

mrmustard/lab_dev/circuit_components.py
```diff
@@ -470,11 +530,37 @@ def __matmul__(self, other: CircuitComponent | Scalar) -> CircuitComponent:
         Contracts ``self`` and ``other`` without adding adjoints.
         It allows for a more custom way of contracting components.
```
ziofil, line 532:
```suggestion
        Adds ``self`` and ``other``. If the representation is ``Fock`` it can add the arrays directly.
        Otherwise it will stack along the batch dimension in the ``Ansatz``.
```

mrmustard/lab_dev/circuit_components.py
```diff
@@ -470,11 +530,37 @@ def __matmul__(self, other: CircuitComponent | Scalar) -> CircuitComponent:
...
             return self * other
+
         wires_result, perm = self.wires @ other.wires
         idx_z, idx_zconj = self._matmul_indices(other)
-        rep = self.representation[idx_z] @ other.representation[idx_zconj]
+
+        if isinstance(self.representation, Bargmann) and isinstance(other.representation, Bargmann):
+            rep = self.representation[idx_z] @ other.representation[idx_zconj]
+            rep = rep.reorder(perm) if perm else rep
+            return CircuitComponent._from_attributes(rep, wires_result, None)
```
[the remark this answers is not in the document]
ziofil: Yep, this is an excellent question. We do, and in fact `__rshift__` is overrided in `Ket`, `DM`, `Unitary` and `Channel` because in those cases we do know more and we can create more accurate objects. In full generality (such as the generality that `__matmul__` grants us) nothing can be said.

mrmustard/lab_dev/circuit_components.py
ziofil: yes, good idea

mrmustard/lab_dev/circuits.py
```diff
@@ -76,6 +83,124 @@ def __init__(self, components: Optional[Sequence[CircuitComponent]] = None) -> N
...
+                ovlp_ket = opA.wires.output.ket.modes & opB.wires.input.ket.modes
+                if not (ovlp_bra or ovlp_ket):
+                    continue
+                iA = opA.wires.output.bra[ovlp_bra].indices + opA.wires.output.ket[ovlp_ket].indices
+                iB = opB.wires.input.bra[ovlp_bra].indices + opB.wires.input.ket[ovlp_ket].indices
+                if not out_idx.intersection(iA):
+                    continue
+                assert list(iA) == sorted(iA)
+                assert list(iB) == sorted(iB)
+                assert len(iA) == len(iB)
```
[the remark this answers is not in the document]
ziofil: they can go 👍

mrmustard/lab_dev/circuits.py
```diff
@@ -76,6 +83,124 @@ def __init__(self, components: Optional[Sequence[CircuitComponent]] = None) -> N
...
+                        c = a + b
+                        changes = True
+                    if not d or d > a + b:
+                        d = a + b
+                        changes = True
+
+                self.components[i].fock_shape = [a, b, c, d]
+
+        # TODO: propagate through where A matrix is block-antidiagonal
+        # TODO: do S2gate (easy!)
```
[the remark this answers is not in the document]
ziofil: no, they're more reminders for the future (thought the S2 gate would be indeed as easy as the block above for the BS ^^

mrmustard/lab_dev/states/base.py
```diff
@@ -648,6 +631,29 @@ def __init__(
...
+        r"""
+        A pretty enough estimate of the Fock shape of this DM, defined as the shape of the Fock
```
ziofil, line 636:
```suggestion
        A good enough estimate of the Fock shape of this DM, defined as the shape of the Fock
```

mrmustard/lab_dev/states/base.py
ziofil: Yes, sounds good let's do that and think whether we need them. This specific side effect (which is documented in both places) does not modify the entries of `fock_shape` that are already set: it only completes what a user doesn't yet know (and in any case `fock_shape` is designed to be editable, so a user can still change it in the future).

mrmustard/math/lattice/strategies/vanilla.py
ziofil: I was always writing them with the developer in mind (e.g. the letters next to the wires in the ascii art are the same as the ones in the code), but perhaps there are merits to rendering them. Is that possible to do?

mrmustard/math/lattice/strategies/vanilla.py
ziofil: hahahaha yes

mrmustard/math/lattice/strategies/vanilla.py
```diff
@@ -188,3 +190,130 @@ def vanilla_vjp(G, c, dLdG) -> tuple[ComplexMatrix, ComplexVector, complex]:  #
...
+                + A_[1, 1] * buf3[k % 2, 0] * SQRT[k]
+            ) / SQRT[k + 1]
+            buf3[(k + 1) % 2, 2] = (
+                b_[1] * buf2[(k + 1) % 2, 1]
+                + A_[1, 0] * buf3[k % 2, 2] * SQRT[k]
+                + A_[1, 1] * buf3[k % 2, 1] * SQRT[k + 1]
+            ) / SQRT[k + 2]
+            norm += np.abs(buf3[(k + 1) % 2, 1])
+            k += 1
+        shape[t] = k
```
ziofil, line 318:
```suggestion
        shape[m] = k
```

mrmustard/lab_dev/circuit_components.py
ziofil, line 650:
```suggestion
        cls = self.__class__.__name__
        repr = self.representation
        repr_name = repr.__class__.__name__
        return cls + f"(modes={self.modes}, name={self.name or None}" + ("repr=" + repr_name if repr else "")
```

mrmustard/lab_dev/circuits.py
ziofil, line 119:
```suggestion
```

mrmustard/lab_dev/circuit_components.py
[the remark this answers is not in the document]
timmysilv: haha yes, I was more thinking of the private `_fock_shape` part of it. That way, you can _only_ change individual elements by using the setter. but it's ok if we leave it as-is

mrmustard/lab_dev/circuit_components.py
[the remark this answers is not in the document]
timmysilv: I don't think we have to pick a default value. If we don't have a good/informed recommendation, we should just ask the user to kindly provide what _they_ think is a good shape, as I suspect they'll know better than this, no?

mrmustard/lab_dev/circuits.py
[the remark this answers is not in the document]
timmysilv: sounds good - do we have a general strategy for tracking TODOs in MrMustard? Notion tasks, shortcut stories, etc.

mrmustard/math/lattice/strategies/vanilla.py
timmysilv: definitely, you can just apply a diff like this one:

<details>
<summary>The Diff</summary>

```diff
diff --git a/doc/code/math.rst b/doc/code/math.rst
index d27d6af0..4729e8a0 100644
--- a/doc/code/math.rst
+++ b/doc/code/math.rst
@@ -12,6 +12,7 @@ The `math` module
     math/parameter_set
     math/parameter
     math/backend_manager
+    math/vanilla
 
 .. automodapi:: mrmustard.math
     :no-heading:
diff --git a/mrmustard/math/lattice/strategies/vanilla.py b/mrmustard/math/lattice/strategies/vanilla.py
index c0204617..b336082a 100644
--- a/mrmustard/math/lattice/strategies/vanilla.py
+++ b/mrmustard/math/lattice/strategies/vanilla.py
@@ -19,7 +19,7 @@ from mrmustard.math.lattice import paths, steps
 from mrmustard.utils.typing import ComplexMatrix, ComplexTensor, ComplexVector
 from .flat_indices import first_available_pivot, lower_neighbors, shape_to_strides
 
-__all__ = ["vanilla", "vanilla_batch", "vanilla_jacobian", "vanilla_vjp"]
+__all__ = ["vanilla", "vanilla_batch", "vanilla_jacobian", "vanilla_vjp", "autoshape_numba"]
 
 SQRT = np.sqrt(np.arange(1000))
 
@@ -208,12 +208,14 @@ def autoshape_numba(
         max_prob (float): the probability value to stop at (default 0.999)
         max_shape (int): max value before stopping (default 100)
 
-    Details:
+    **Details:**
 
     Here's how it works. First we get the reduced density matrix at the given mode. Then we
     maintain two buffers that contain values around the diagonal: buf3 and buf2, of size 2x3
     and 2x2 respectively. The buffers contain the following elements of the density matrix:
 
+    .. code-block::
+
                      ┌────┐                           ┌────┐
                      │buf3│                           │buf2│
                      └────┘                           └────┘
@@ -235,6 +237,8 @@ def autoshape_numba(
 
     The updates roughly look like this:
 
+    .. code-block::
+
                     ┌───────┐                     ┌───────┐
                     │k even │                     │ k odd │
                     └───────┘                     └───────┘
```

</details>

Then just create a file called `doc/code/math/vanilla.rst` with these contents:

```
Vanilla
==========

.. currentmodule:: mrmustard.math.lattice.strategies.vanilla

.. automodapi:: mrmustard.math.lattice.strategies.vanilla
    :no-heading:
```

and it'll appear at the top of the `mrmustard.math` section!

ziofil: thanks!

ziofil:
> This feels like I'm just not getting it, but should `.auto_shape()` have the side-effect of setting the fock_shape attribute? In my basic test below, I just get `[None]`:
> 
> ```
> import mrmustard as mm
> from mrmustard.lab_dev import *
> 
> cc = Number([0], 4) >> Sgate([0], 1)
> cc.auto_shape()
> print(cc.fock_shape)
> ```
> 
> It seems that `.auto_shape()` in this instance returns `(100,)`. Is this right?

Yep that was a bug! `CircuitComponent` was initializing `_fock_shape` in the wrong way. It should be fixed now.

mrmustard/lab_dev/circuit_components.py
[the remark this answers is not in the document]
ziofil: Yes they may know better (not always). And in fact they are free to change the default by modifying `.fock_shape`, it's just that the completely hands-off experience still works if there is a last-resort default.

mrmustard/lab_dev/circuits.py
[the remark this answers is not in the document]
ziofil: actually we should create a couple cards in the tracker 👍

mrmustard/lab_dev/states/states.py
```diff
@@ -205,39 +205,23 @@ def __init__(
...
-            msg = f"Length of ``n`` must be 1 or {len(modes)}, found {len(self._n)}."
-            raise ValueError(msg)
-
-        self._cutoffs = math.atleast_1d(cutoffs) if cutoffs else self.n
-        if len(self._cutoffs) == 1:
-            self._cutoffs = math.tile(self._cutoffs, [len(modes)])
-        if len(self._cutoffs) != len(modes):
-            msg = f"Length of ``cutoffs`` must be 1 or {len(modes)}, found {len(self._cutoffs)}."
-            raise ValueError(msg)
+        super().__init__(modes=modes, name=f"{n}")
```
elib20, line 208: I recognize this is a subjective thing, but this now looks weird to me. Why not Number(n) or Fock(n) as the name?
image

elib20: I won't die on this hill, FYI 😆

mrmustard/math/lattice/strategies/vanilla.py
```diff
@@ -52,7 +60,7 @@ def vanilla(shape: tuple[int, ...], A, b, c) -> ComplexTensor:  # pragma: no cov
     index_u_iter = np.ndindex(shape)
     next(index_u_iter)
 
-    # write vacuum amplitude
+    # write vacuum amplitude /
```
elib20, line 63:
```suggestion
    # write vacuum amplitude
```

mrmustard/lab_dev/states/base.py
```diff
@@ -648,6 +631,29 @@ def __init__(
...
+        """
+        try:  # fock
+            shape = self._representation.array.shape[1:]
+        except AttributeError:  # bargmann
+            repr = self.representation
+            shape = autoshape_numba(
+                math.asnumpy(repr.A[0]),
+                math.asnumpy(repr.b[0]),
+                math.asnumpy(repr.c[0]),
+            )
```
elib20, line 651: Wasn't exactly sure where to put this comment, but I think we should run this on the renormalized state. For example, we may have a conditional state that is subnormalized, so it will never get to 99.9% and this will always send it to the max.

Example of a subnormalized state:
`(Coherent([0],0).dm() * 0.25).auto_shape() #shape (100,100)`

mrmustard/lab_dev/states/base.py
```diff
@@ -827,6 +833,30 @@ def __init__(
...
+            shape = self._representation.array.shape[1:]
+        except AttributeError:  # bargmann
+            if None not in self.fock_shape:  # try the fock_shape if ready
+                return tuple(self.fock_shape)
+            repr = self.representation.conj() & self.representation
+            shape = autoshape_numba(
+                math.asnumpy(repr.A[0]),
+                math.asnumpy(repr.b[0]),
+                math.asnumpy(repr.c[0]),
+            )
```
elib20, line 855: See comment in DM section about subnormalized states.

mrmustard/lab_dev/states/base.py
elib20: More of an edge case (i.e. can't see how it would come about), for supernormalized states, this may truncate early.

mrmustard/math/lattice/strategies/vanilla.py
```diff
@@ -188,3 +196,134 @@ def vanilla_vjp(G, c, dLdG) -> tuple[ComplexMatrix, ComplexVector, complex]:  #
     dLdc = np.sum(G_lin.reshape(shape) * dLdG) / c
 
     return dLdA, dLdb, dLdc
+
+
+@njit
+def autoshape_numba(
+    A, b, c, max_prob=0.999, max_shape=100
```
elib20, line 203: Should max_prob be `settings.AUTOCUTOFF_PROBABILITY`?

elib20: Maybe add a test that changing that setting actually changes the cutoff.

elib20: This was the behaviour I saw
```
Coherent([0],1).auto_shape() #5
settings.AUTOCUTOFF_PROBABILITY = 0.99999
Coherent([0],1).auto_shape() #still 5
```

mrmustard/lab_dev/circuits.py
```diff
@@ -22,18 +22,25 @@
...
+    possibility of contracting them in different orders. The order in which the components
+    are contracted is specified by the ``path`` attribute.
+
+    Different orders of contraction lead to the same result, but the cost of the contraction
+    can vary significantly. The ``path`` attribute is used to specify the order in which the
+    components are contracted.
 
-    Quantum circuits store a list of ``CircuitComponent``s.
+    We supply an automatic path generator that uses a graph representation of the circuit
+    and a branch-and-bound algorithm to find the optimal path. The path can also be set manually.
```
elib20, line 43: is this true yet from this PR?

elib20 (commented):
Awesome work! 

I found a few spots with unexpected behaviour/edge case bugs, but nothing that can't be easily squashed.

mrmustard/lab_dev/circuit_components.py
```diff
@@ -368,45 +387,77 @@ def on(self, modes: Sequence[int]) -> CircuitComponent:
...
-            self.name,
-        )
+        fock = Fock(math.astensor(self.fock(shape)), batched=True)
+        fock._original_bargmann_data = self.representation.data
+        return self._from_attributes(fock, self.wires, self.name)
+
+    def auto_shape(self) -> tuple[int, ...]:
+        r"""
+        The shape of the Fock representation of this component. If the component has a Fock representation
+        then it is just the shape of the array. If the components is a State in Bargmann
```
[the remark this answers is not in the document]
ziofil: I think so. These two methods have the same name, but they are not the same method (they do things differently). I would expect `help(cc.auto_shape)` and `help(ket.auto_shape)` to give me different descriptions.

aplund:
Again, I may be "doing it wrong(TM)", but I was expecting this to change the shape as the squeezing increases.  Am I missing a step somewhere:

```python
import numpy as np
import mrmustard as mm
from mrmustard.lab_dev import *

for r in np.linspace(0.0, 3.0, 31):
    cc = Number([0], 0) >> Sgate([0], r)
    cc.auto_shape()
    f = cc.fock()
    tail = np.sqrt(np.sum(np.abs(f[..., -10:])**2))
    print(f"r={r:0.2f}, shape={cc.fock_shape}, tail={tail}")

```

Here's my output:
```
r=0.00, shape=[50], tail=0.0
r=0.10, shape=[50], tail=3.320813549783774e-21
r=0.20, shape=[50], tail=2.877796850159443e-15
r=0.30, shape=[50], tail=7.005423842389163e-12
r=0.40, shape=[50], tail=1.4442895820803e-09
r=0.50, shape=[50], tail=7.391845079536454e-08
r=0.60, shape=[50], tail=1.527182338092643e-06
r=0.70, shape=[50], tail=1.6608089924466752e-05
r=0.80, shape=[50], tail=0.0001119152559711487
r=0.90, shape=[50], tail=0.0005210655981549683
r=1.00, shape=[50], tail=0.0018108286711155058
r=1.10, shape=[50], tail=0.004974852404081372
r=1.20, shape=[50], tail=0.011292920255741735
r=1.30, shape=[50], tail=0.02193479266151498
r=1.40, shape=[50], tail=0.03749458889213709
r=1.50, shape=[50], tail=0.057708136029986395
r=1.60, shape=[50], tail=0.08148204007111474
r=1.70, shape=[50], tail=0.10718002773885009
r=1.80, shape=[50], tail=0.1330086283344037
r=1.90, shape=[50], tail=0.15735305275690456
r=2.00, shape=[50], tail=0.178984595152066
r=2.10, shape=[50], tail=0.1971322790005048
r=2.20, shape=[50], tail=0.2114532602089405
r=2.30, shape=[50], tail=0.22194764197883732
r=2.40, shape=[50], tail=0.22885591269080502
r=2.50, shape=[50], tail=0.23256342153828174
r=2.60, shape=[50], tail=0.23352358559759126
r=2.70, shape=[50], tail=0.23220268304523672
r=2.80, shape=[50], tail=0.22904418030721152
r=2.90, shape=[50], tail=0.2244485284290604
r=3.00, shape=[50], tail=0.21876408762339897
```

Should the shape not increase as `r` increases?

mrmustard/lab_dev/circuit_components.py
```diff
@@ -368,45 +387,77 @@ def on(self, modes: Sequence[int]) -> CircuitComponent:
...
+            modes_out_bra=set(modes) if ob else set(),
+            modes_in_bra=set(modes) if ib else set(),
+            modes_out_ket=set(modes) if ok else set(),
+            modes_in_ket=set(modes) if ik else set(),
         )
 
         return ret
 
-    def to_fock(self, shape: Optional[Union[int, Iterable[int]]] = None) -> CircuitComponent:
+    def fock(self, shape: Optional[int | Sequence[int]] = None) -> CircuitComponent:
```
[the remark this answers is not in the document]
ziofil: Ok, I remember thinking about this, and I concluded that I wanted to avoid padding with zeros and I'd rather chop the representations that will contract with this one instead. But you are right that it can be unexpected to put some numbers in and then get a different shape out. What if we raise an exception and return the smaller array? Then the user can make it bigger if they want to, but I don't want MM to be able to do the 'wrong thing' (to avoid uncaught regressions).

mrmustard/lab_dev/circuits.py
ziofil: yeah no this wasn't removed

mrmustard/lab_dev/states/base.py
```diff
@@ -648,6 +631,29 @@ def __init__(
         if representation is not None:
             self._representation = representation
 
+    def auto_shape(self) -> tuple[int, ...]:
+        r"""
+        A good enough estimate of the Fock shape of this DM, defined as the shape of the Fock
+        array (batch excluded) if it exists, and if it doesn't exist it is computed as the shape
+        that captures at least ``settings.AUTOCUTOFF_PROBABILITY`` of the probability mass of each
```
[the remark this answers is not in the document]
ziofil: yes actually I'll also add some other option given Austin's comments

mrmustard/lab_dev/states/base.py
ziofil: yes good idea

mrmustard/lab_dev/states/states.py
ziofil: Oh that was so that in the circuit a photon number state looks like this:

Screenshot 2024-07-12 at 9 37 14 AM

what we should do is not to use that in the visualization in your screenshot

mrmustard/lab_dev/circuit_components.py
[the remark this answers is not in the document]
elib20: What if the argument was called `max_shape` to hint that this will return the most compact thing and users only have the option to make it even more compact?

ziofil:
> Again, I may be "doing it wrong(TM)", but I was expecting this to change the shape as the squeezing increases. Am I missing a step somewhere:
> [...]
> Should the shape not increase as `r` increases?

@aplund I think you wanted this:

```python
for r in np.linspace(0.0, 3.0, 31):
    cc = Vacuum([0]) >> Sgate([0], r)
    cc.auto_shape()
    print(f"r={r:0.2f}, shape={cc.fock_shape}")  # works
```
The issue with your code was that `Number` has a `Fock` representation, so the `Sgate` needs to go to Fock itself and the best it can do is to match the shape (of 1) on the vacuum side and on the output side it has no way to determine how far to go and so it defaults to the value in the settings. On the other hand, Vacuum has a Bargmann representation so we can play all the tricks in our book.

mrmustard/lab_dev/circuit_components.py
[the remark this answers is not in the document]
ziofil: that's another option. I'll think about it.

aplund:
> The issue with your code was that `Number` has a `Fock` representation, so the `Sgate` needs to go to Fock itself and the best it can do is to match the shape (of 1) on the vacuum side and on the output side it has no way to determine how far to go and so it defaults to the value in the settings. On the other hand, `Vacuum` has a `Bargmann` representation so we can play all the tricks in our book.

So there's an implicit difference between `Number([0], 0)` and `Vacuum([0])` that makes them not interchangeable?  If so, this should probably be bolded and italicised somewhere, as I think most people would consider them the same.

When I run the code you suggested, I get the output:

```
r=0.00, shape=[None]
r=0.10, shape=[None]
r=0.20, shape=[None]
r=0.30, shape=[None]
r=0.40, shape=[None]
r=0.50, shape=[None]
r=0.60, shape=[None]
r=0.70, shape=[None]
r=0.80, shape=[None]
r=0.90, shape=[None]
r=1.00, shape=[None]
r=1.10, shape=[None]
r=1.20, shape=[None]
r=1.30, shape=[None]
r=1.40, shape=[None]
r=1.50, shape=[None]
r=1.60, shape=[None]
r=1.70, shape=[None]
r=1.80, shape=[None]
r=1.90, shape=[None]
r=2.00, shape=[None]
r=2.10, shape=[None]
r=2.20, shape=[None]
r=2.30, shape=[None]
r=2.40, shape=[None]
r=2.50, shape=[None]
r=2.60, shape=[None]
r=2.70, shape=[None]
r=2.80, shape=[None]
r=2.90, shape=[None]
r=3.00, shape=[None]
```

Is this the expected behaviour?

mrmustard/lab_dev/circuit_components.py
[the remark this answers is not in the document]
ziofil: I went for padding and warning

mrmustard/lab_dev/states/base.py
ziofil: done

mrmustard/lab_dev/states/base.py
ziofil: done

mrmustard/math/lattice/strategies/vanilla.py
ziofil: now it works. note that we're moving away from the word "cutoff" in favour of "shape" so now the setting is "AUTOSHAPE_PROBABILITY"

ziofil:
> So there's an implicit difference between `Number([0], 0)` and `Vacuum([0])` that makes them not interchangeable? If so, this should probably be bolded and italicised somewhere, as I think most people would consider them the same.
Sure, we can make it clearer in the docstrings.

> When I run the code you suggested, I get the output:
> 
> ```
> r=0.00, shape=[None]
> ...
> ```
> 
> Is this the expected behaviour?

Yes, now yes. I have removed the lines where `fock_shape` is updated by `auto_shape`, which takes about 100us per mode to run, so it's okay if we run it multiple times.

mrmustard/lab_dev/circuit_components.py
```diff
@@ -475,11 +526,37 @@ def __matmul__(self, other: CircuitComponent | Scalar) -> CircuitComponent:
         Contracts ``self`` and ``other`` without adding adjoints.
         It allows for a more custom way of contracting components.
```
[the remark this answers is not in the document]
ziofil: done

mrmustard/physics/representations.py
```diff
@@ -466,20 +466,11 @@ def __matmul__(self, other: Union[Bargmann, Fock]) -> Union[Bargmann, Fock]:
             A ``Bargmann`` representation if ``other`` is ``Bargmann``, or a ``Fock``representation
             if ``other`` is ``Fock``.
         """
-        idx_s = self._contract_idxs
-        idx_o = other._contract_idxs
-
-        # if ``other`` is ``Fock``, convert ``self`` to ``Fock``
         if isinstance(other, Fock):
-            from .converters import to_fock  # pylint: disable=import-outside-toplevel
+            raise NotImplementedError("Only matmul Bargmann with Bargmann")
```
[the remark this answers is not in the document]
ziofil: done

mrmustard/physics/representations.py
```diff
@@ -632,34 +623,33 @@ def __getitem__(self, idx: int | tuple[int, ...]) -> Fock:
...
+            >>> g = Fock(np.random.random((2, 5, 8)))
+            >>> h = f[1,2] @ g[1,2]
+            >>> assert h.array.shape == (1,3,2)  # batch size is 1
+            >>> f = Fock(np.random.random((3, 5, 10)), batched=True)
+            >>> g = Fock(np.random.random((2, 5, 8)), batched=True)
+            >>> h = f[0,1] @ g[0,1]
+            >>> assert h.array.shape == (6,)  # batch size is 3 x 2 = 6
+
+        If ``other`` is ``Bargmann``, it is converted to ``Fock`` before the contraction
+        using auto_shape where possible, or settings.AUTOSHAPE_MAX where not.
```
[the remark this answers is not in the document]
ziofil: done

mrmustard/lab_dev/circuits.py
ziofil: done

mrmustard/lab_dev/states/base.py
```diff
@@ -648,6 +636,38 @@ def __init__(
...
+            max_prob: The maximum probability mass to capture in the shape (default 0.999).
+            max_shape: The maximum shape to compute (default 50).
+            respect_fock_shape: Whether to respect the non-None values in ``fock_shape``.
+        """
+        try:  # fock
+            shape = self._representation.array.shape[1:]
+        except AttributeError:  # bargmann
+            repr = self.representation
+            A, b, c = repr.A[0], repr.b[0], repr.c[0]
+            repr = repr / trace_dm(A, b, c)
```
elib20, line 658: Out of curiosity, why not use the probability attribute of the state?

elib20 (approved):
My only remaining comment is to please include NotImplementedError's for autoshape on batched Bargmann DMs and Kets, since if the method was used as currently written, it would produce incorrect results.

Since that is an easy change to make, and since my only other comments are just catching a couple typos in docstrings, I'm providing an approval so that you don't need to wait on a follow-up review from me after you make those changes.

Not sure if @timmysilv  or @aplund wanted to have a second look either?

mrmustard/lab_dev/states/base.py
ziofil: because `.probability` triggers a long sequence of events: it is computed via the L2_norm i.e. `self >> self.dual` which calls `__rshift__`, which calls `__matmul__` which calls `gaussian_integral`. This way we have a direct fasg formula instead.

mrmustard/lab_dev/states/base.py
```diff
@@ -631,28 +636,37 @@ def __init__(
         if representation is not None:
             self._representation = representation
 
-    def auto_shape(self) -> tuple[int, ...]:
+    def auto_shape(self, max_prob=0.999, max_shape=50, respect_fock_shape=True) -> tuple[int, ...]:
```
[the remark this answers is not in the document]
timmysilv: docstring should also be updated to reflect that it now depends on settings

mrmustard/lab_dev/states/base.py
timmysilv: does this mean we can generally compute probability via representation faster? should we capture this somewhere in the Bargmann class?

mrmustard/lab_dev/states/base.py
```diff
@@ -827,6 +847,36 @@ def __init__(
...
+        r"""
+        A good enough estimate of the Fock shape of this Ket, defined as the shape of the Fock
+        array (batch excluded) if it exists, and if it doesn't exist it is computed as the shape
+        that captures at least ``settings.AUTOSHAPE_PROBABILITY`` of the probability mass of each
+        single-mode marginal (default 99.9%).
+        If the ``respect_fock_shape`` flag is set to ``True``, auto_shape will respect the
+        non-None values in ``fock_shape``.
+
+        Args:
+            max_prob: The maximum probability mass to capture in the shape (default 0.999).
```
timmysilv, line 860: again, default comes from settings, maybe we should just point there?

mrmustard/lab_dev/wires.py
```diff
@@ -336,6 +336,14 @@ def __bool__(self) -> bool:
         r"Returns ``True`` if this ``Wires`` object has any wires, ``False`` otherwise."
         return any(self.args)
 
+    def __len__(self) -> int:
+        r"The number of wires."
+        try:
+            return self._len
+        except AttributeError:
+            self._len = sum(map(len, self.args))
+            return self._len
```
timmysilv, line 345: is this expensive/does it need to be cached?

mrmustard/lab_dev/wires.py
```diff
@@ -336,6 +336,14 @@ def __bool__(self) -> bool:
         r"Returns ``True`` if this ``Wires`` object has any wires, ``False`` otherwise."
         return any(self.args)
 
+    def __len__(self) -> int:
+        r"The number of wires."
```
timmysilv, line 340:
```suggestion
        r"""The number of wires."""
```

mrmustard/math/lattice/strategies/vanilla.py
```diff
@@ -188,3 +196,134 @@ def vanilla_vjp(G, c, dLdG) -> tuple[ComplexMatrix, ComplexVector, complex]:  #
     dLdc = np.sum(G_lin.reshape(shape) * dLdG) / c
 
     return dLdA, dLdb, dLdc
+
+
+# @njit
```
timmysilv, line 201: is it not jit-friendly yet? will we make it so later?

timmysilv (approved):
looks good! just a few little things

mrmustard/lab_dev/circuit_components.py
```diff
@@ -368,45 +387,81 @@ def on(self, modes: Sequence[int]) -> CircuitComponent:
...
                 an ``int``, it is broadcasted to all the dimensions. If ``None``, it
-                defaults to the value of ``AUTOCUTOFF_MAX_CUTOFF`` in the settings.
+                defaults to the value of ``AUTOSHAPE_MAX`` in the settings.
         """
-        return self.__class__._from_attributes(
-            to_fock(self.representation, shape=shape),
-            self.wires,
-            self.name,
-        )
+        fock = Fock(math.astensor(self.fock(shape, batched=True)), batched=True)
```
apchytr, line 452: Would it be clearer to rename `self.fock` to `self.fock_array`?

mrmustard/lab_dev/circuit_components.py
```diff
@@ -368,45 +387,81 @@ def on(self, modes: Sequence[int]) -> CircuitComponent:
...
+            modes_out_bra=set(modes) if ob else set(),
+            modes_in_bra=set(modes) if ib else set(),
+            modes_out_ket=set(modes) if ok else set(),
+            modes_in_ket=set(modes) if ik else set(),
         )
 
         return ret
 
-    def to_fock(self, shape: Optional[Union[int, Iterable[int]]] = None) -> CircuitComponent:
+    def fock(self, shape: Optional[int | Sequence[int]] = None, batched=False) -> CircuitComponent:
```
apchytr, line 399: I think the type hint here should be an array rather than a CircuitComponent

mrmustard/lab_dev/states/base.py
ziofil: I don't think it's necessary for now to run after that

mrmustard/lab_dev/wires.py
ziofil: not expensive, but also will never change so might as well cache it.

mrmustard/lab_dev/circuit_components.py
ziofil: you're right that it might be clearer, but for the other representations we use only the name (bargmann, quadrature etc)

mrmustard/lab_dev/states/base.py
```diff
@@ -648,6 +635,45 @@ def __init__(
...
+        r"""
+        A good enough estimate of the Fock shape of this DM, defined as the shape of the Fock
+        array (batch excluded) if it exists, and if it doesn't exist it is computed as the shape
+        that captures at least ``settings.AUTOSHAPE_PROBABILITY`` of the probability mass of each
+        single-mode marginal (default 99.9%).
+        If the ``respect_manual_shape`` flag is set to ``True``, auto_shape will respect the
+        non-None values in ``manual_shape``.
+
+        Args:
+            max_prob: The maximum probability mass to capture in the shape (default 0.999).
```
ziofil, line 650:
```suggestion
            max_prob: The maximum probability mass to capture in the shape (default in ``settings.AUTOSHAPE_PROBABILITY``).
```

mrmustard/lab_dev/states/base.py
```diff
@@ -648,6 +635,45 @@ def __init__(
...
+            max_shape: The maximum shape to compute (default 50).
```
ziofil, line 651:
```suggestion
            max_shape: The maximum shape to compute (default in ``settings.AUTOSHAPE_MAX``).
```

mrmustard/math/lattice/strategies/vanilla.py
ziofil: oops nope, I just forgot to uncomment!

tests/test_lab_dev/test_circuit_components.py
```diff
@@ -175,28 +175,22 @@ def test_on_error(self):
         with pytest.raises(ValueError):
             Vacuum([1, 2]).on([3])
 
-    @pytest.mark.parametrize("shape", [3, [3, 2]])
-    def test_to_fock(self, shape):
+    def test_to_fock_ket(self):
         vac = Vacuum([1, 2])
-        vac_fock = vac.to_fock(shape=shape)
-        assert vac_fock.name == vac.name
-        assert vac_fock.wires == vac.wires
```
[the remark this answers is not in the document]
ziofil: tested elsewhere

elib20:
Hi Jacob. There are two settings that can be modified instead. AUTOSHAPE_PROBABILITY allows you to set a minimum norm that you want the autoshape to preserve for states. AUTOSHAPE_MAX is the max shape you want Fock objects to go to (so it won't go higher than this for states, and it will automatically go to this for things that are not states).

From https://github.com/XanaduAI/MrMustard/blob/develop/mrmustard/utils/settings.py:

`        self.AUTOSHAPE_PROBABILITY = 0.999
        "The minimum l2_norm to reach before automatically stopping the Bargmann-to-Fock conversion. Default is 0.999."

        self.AUTOCUTOFF_MAX_CUTOFF = 100  # TODO: remove in MM 1.0
        r"""The maximum value for autocutoff. Default is ``100``."""

        self.AUTOCUTOFF_MIN_CUTOFF = 1  # TODO: remove in MM 1.0
        r"""The minimum value for autocutoff. Default is ``1``."""

        self.AUTOSHAPE_MAX = 50
        r"""The max shape for the autoshape. Default is ``50``."""`

JacobHast:
So AUTOCUTOFF_MIN_CUTOFF does nothing then?

elib20:
Not in lab_dev, as far as I know. It is still there for lab.

JacobHast:
Should we maybe have self.AUTOSHAPE_PROBABILITY  default to 1 instead of 0.999? Or maybe think about some relative way of doing things when Fock measurements on Gaussian states are involved, e.g., in my example above the default behavior doesn't seem desired

JacobHast:
I.e. in my example about I get 0% of the correct output state, whereas I might have expected to get at least 0.999 of it

PR #399, opened 2024-05-10
quadrature repr in CircuitComponent

ziofil:
- Adds `quadrature` representation at the level of CircuitComponent: since it's a single-wire operation it always exists.

- Simplifies having to check for representation type when extracting data by creating a common abstraction `scalar` which means `c` for Bargmann and `array` for Fock, and `data` which is `triple` for Bargmann and `array` for Fock. This way we simplify lots of useless `if`s and `isinstance` checks.


mrmustard/lab_dev/circuit_components.py
```diff
@@ -146,11 +146,30 @@ def bargmann(self) -> tuple:
...
+            ) from e
+
+    def quadrature(self) -> tuple | ComplexTensor:
+        r"""
+        The quadrature representation of this circuit component.
+        """
+        from mrmustard.lab_dev.circuit_components_utils import (  # pylint: disable=import-outside-toplevel
+            BtoQMap,
+        )
+
```
sylviemonet, line 163:
```suggestion

# The representation change from Bargmann into quadrature is to use the BtoQMap.
# Here for a CircuitComponent, we need to add this map four times: BtoQMap on out_ket wires, BtoQMap.dual on in_ket wires, BtoQMap.adjoint on out_bra wires and BtoQMap.adjoint.dual on in_bra wires.
```

ziofil:
@elib20 can I get an approval when you have a minute?

mrmustard/lab_dev/circuit_components.py
```diff
@@ -151,11 +150,32 @@ def bargmann(self) -> tuple:
...
-        return self.representation.triple
+            ) from e
+
+    def quadrature(self) -> tuple | ComplexTensor:
+        r"""
+        The quadrature representation of this circuit component.
+        """
+        from mrmustard.lab_dev.circuit_components_utils import (  # pylint: disable=import-outside-toplevel
+            BtoQMap,
+        )
```
elib20, line 166: Why do we have to do the import here?

elib20: Should we be following the same structure as to_fock, and put a converter in the physics module?

mrmustard/lab_dev/circuit_components.py
```diff
@@ -248,21 +268,17 @@ def on(self, modes: Sequence[int]) -> CircuitComponent:
                 msg = f"Expected ``{len(modes)}`` modes, found ``{len(subset.modes)}``."
                 raise ValueError(msg)
 
-        wires = Wires(
+        ret = self.light_copy()
+        ret._wires = Wires(
             modes_out_bra=modes if ob else set(),
```
elib20, line 273: I understand this to be: if ob is not defined, make modes_out_bra equal to set(). But why wouldn't ob already be equal to set()?

mrmustard/lab_dev/circuit_components.py
```diff
@@ -162,6 +165,8 @@ def quadrature(self) -> tuple | ComplexTensor:
             BtoQMap,
         )
 
+        # The representation change from Bargmann into quadrature is to use the BtoQMap.
+        # Here for a CircuitComponent, we need to add this map four times: BtoQMap on out_ket wires, BtoQMap.dual on in_ket wires, BtoQMap.adjoint on out_bra wires and BtoQMap.adjoint.dual on in_bra wires.
```
elib20, line 169:
```suggestion
        # Here for a CircuitComponent, we need to add this map four times: BtoQMap on out_ket 
        # wires, BtoQMap.dual on in_ket wires, BtoQMap.adjoint on out_bra wires and BtoQMap.adjoint.dual 
        # on in_bra wires.
```

mrmustard/physics/representations.py
```diff
@@ -523,8 +555,9 @@ class Fock(Representation):
 
     """
 
-    def __init__(self, array: Batch[Tensor], batched=False):
+    def __init__(self, array: Batch[Tensor], batched=False, original_bargmann_data=None):
```
elib20, line 558: Is Fock user-facing? How do we ensure that the array given is consistent with original_bargmann_data?

mrmustard/lab_dev/circuit_components.py
```diff
@@ -151,11 +150,32 @@ def bargmann(self) -> tuple:
...
+        try:
+            return self.representation.triple
+        except AttributeError as e:
+            raise AttributeError(
                 f"Cannot compute triple from representation of type ``{self.representation.__class__.__qualname__}``."
-            )
-        return self.representation.triple
+            ) from e
+
+    def quadrature(self) -> tuple | ComplexTensor:
```
elib20, line 160: Is this tested somewhere?

elib20 (commented):
Thanks for waiting! I have a few questions.

mrmustard/physics/representations.py
ziofil: How do mean consistent? It cannot come from anything else, if that's what you meant

mrmustard/lab_dev/circuit_components.py
ziofil: The map is tested here, but for sure we could beef up the tests of this and other things.
https://github.com/XanaduAI/MrMustard/blob/3f804033e0ed9c505677768534b615daf83c15b5/tests/test_lab_dev/test_circuit_components_utils.py#L165C7-L165C18

mrmustard/lab_dev/circuit_components.py
ziofil: It's there to make it local to the scope, to avoid a circular import (BtoQMap is a CircuitComponent and needs to import CircuitComponent to inherit from it).

mrmustard/lab_dev/circuit_components.py
ziofil: Yes, and `bool(set())` evaluates to `False`. So if `ob` is `set()`, we make `modes_out_bra = set()`.

mrmustard/lab_dev/circuit_components.py
ziofil: And no, if anything I want to move away to having converters separate from the representations.

mrmustard/physics/representations.py
ziofil: Oh, I see the question now. Yes, this could be more privately handled.

mrmustard/lab_dev/circuit_components.py
elib20: Yes could be good to have a test of this.

elib20 (approved):
Thanks for addressing the comments! I'm approving, but left a couple remaining comments/questions to be addressed in future.

mrmustard/lab_dev/circuit_components.py
ziofil: yes, for two reasons: 1) we bring functionality together: if I'm in a representation I don't need to jump somewhere else to see how the conversion is done and 2) we avoid a few if statements each time we convert because only the appropriate conversion procedure is followed.

PR #400, opened 2024-05-13
Sauron states

ziofil:
**Context:**
This PR adds Sauron states, i.e. approximations of Number states by a superposition of coherent states on a ring.
Screenshot 2024-05-13 at 1 37 09 PM

**Benefits:**
It's like having Fock states but CV.

**Possible Drawbacks:**
Be careful that the batch dimension can grow considerably when taking products with other objects (the batch dimension distributes)

**Related GitHub Issues:**


ziofil:
PR was falling too far behind. See #433

PR #404, opened 2024-05-22
catch regressions using pytest results

timmysilv:
**Context:**
It's easy for code to slow down without devs realizing, which is why regression-testing is very important. I don't personally know what to reg-test, so I tried something a little different. `pytest` can emit test duration results, so I'm going to use those to track how test durations vary between commits.

1. When merging to `develop`, collect test durations and upload them to an S3 bucket
2. Adds a (fairly scrappy) convenience script to visualize test durations

Additional notes on this:
- both the numpy and tensorflow test results are uploaded, each to their own folders in the S3 bucket
- The file structure should always follow `<test_type>/<branch_name>/durations_<epoch_timestamp>_<commit_short_sha>.txt`. For example, `numpy_tests/develop/durations_1716824126_3d8c81c.txt`. This organization helps with visualization/analysis because the script requires a data folder as an arg, so you can say `python visualize.py numpy_tests/develop` to assess that one set of data. The epoch timestamp is important (and auto-set by the upload workflow) because the analysis script sorts by file name and that value should be monotonic increasing

**Benefits:**
- Easy way to quickly see if any particular tests are slowing down between commits
- Does not require hand-picked benchmarks, works with existing tests and pytest
- Easily extensible in the future:
  - can improve viz scripts
  - can add automated regression detection (check if line jumps)
  - can add more data that follows this format, eg. custom/private results

Example screenshot of the viz results:
image

**Possible Drawbacks:**
- more code to maintain
- the additional CI stuff only runs on merges to `develop` as of this PR, so we might not realize if the automation breaks at some point


.github/workflows/tests_numpy.yml
```diff
@@ -35,5 +37,39 @@ jobs:
...
+      # - name: Reformat durations file
+      #   if: github.event_name == 'push'
+      #   run: grep ' call ' durations.txt | awk '{print $3,$1}' > ${{ steps.record_file.outputs.filename }}
+
+      # - name: Upload benchmark results
+      #   if: github.event_name == 'push'
+      #   uses: actions/upload-artifact@v3
+      #   with:
+      #     name: timings
+      #     path: ${{ steps.record_file.outputs.filename }}
```
ziofil, line 62: if this commented block is not needed let's remove it

ziofil (approved):
super!

.github/workflows/tests_numpy.yml
timmysilv: I'll remove it!

tests/test_lab/test_gates_fock.py
```diff
@@ -75,6 +77,15 @@ def test_Dgate_2mode(state, xxyy):
     assert state_out == state
 
 
+def test_additive_noise_equal_to_circuit():
```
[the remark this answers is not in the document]
timmysilv: the test is unrelated actually, codecov was just failing for unrelated reasons (due to how `hypothesis` picks test cases, see Slack discussion) so I added this test to reliably cover `AdditiveNoise` construction

ziofil (approved):
👍

PR #405, opened 2024-05-29
Operator and Map [merge before QtoB]

ziofil:
**Context:**
The `Unitary` class isn't enough, because we have things that are operators but not unitary. So we introduce `Operator` and `Map` to stay between `CircuitComponent` and `Unitary` and `Channel`, respectively.

- Added code of `Operator` and `Map`, updated names of `BtoQMap` to `BtoQ` and now it inherits from `Operator` (so no need for "map" in the name). Also `DsMap` now becomes `BtoPS` and it inherits from `Map`.
- Moved the `name` field to the end in several methods.

**Benefits:**
Clearer structure, less mistakes going forward, getting things ready for the `MultiRepresentation` class.


mrmustard/lab_dev/wires.py
```diff
@@ -175,6 +175,10 @@ def __init__(
         # Adds elements to the cache when calling ``__getitem__``
         self._mode_cache = {}
 
+    def __len__(self) -> int:
+        r"The number of wires."
+        return sum(len(s) for s in self.args)
```
timmysilv, line 180: wouldn't many wires be counted twice? not sure what our definition of a unique wire is

mrmustard/physics/gaussian.py
```diff
@@ -421,9 +421,7 @@ def CPTP(
         Y = math.single_mode_to_multimode_mat(Y, len(transf_modes))
     if d is not None and d.shape[-1] == 2:
         d = math.single_mode_to_multimode_vec(d, len(transf_modes))
-    indices = [
-        state_modes.index(i) for i in transf_modes
-    ]  # TODO: do this when calling the method instead of here?
```
timmysilv, line 426: does this TODO not need further attention?

timmysilv (commented):
what was the motivation for moving the `name` to the end, because we want it to be optional and it is often omitted?

ziofil:
> what was the motivation for moving the `name` to the end, because we want it to be optional and it is often omitted?

Precisely, it was forcing to to think of a name or pass an empty string every time.

mrmustard/physics/gaussian.py
ziofil: Nope, it's about the old MM and I wanted codefactor to see one fewer TODO..

mrmustard/lab_dev/wires.py
ziofil: why twice? When I define Wires with `Wires(a,b,c,d)` the total number of wires is `len(a) + len(b) + len(c) + len(d)`

mrmustard/lab_dev/transformations/base.py
```diff
@@ -42,25 +42,69 @@ class Transformation(CircuitComponent):
...
         if not isinstance(self.representation, Bargmann):
             raise NotImplementedError("Only Bargmann representation is supported.")
         if self.representation.ansatz.batch_size > 1:
             raise NotImplementedError("Batched transformations are not supported.")
+
+        # compute the inverse
         A, b, _ = self.dual.representation.conj().triple  # apply X
-        almost_inverse = self.__class__.from_bargmann(
-            [0], (math.inv(A[0]), -math.inv(A[0]) @ b[0], 1 + 0j)
+        almost_inverse = self._from_attributes(
```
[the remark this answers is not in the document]
ziofil: I can't remember this specific change, I think they're both ok here.

mrmustard/lab_dev/transformations/base.py
```diff
@@ -71,9 +115,9 @@ class Unitary(Transformation):
...
         super().__init__(
-            name or "U" + "".join(str(m) for m in modes),
-            modes_in_ket=modes,
-            modes_out_ket=modes,
+            modes_in=modes,
+            modes_out=modes,
+            name=name or "U" + "".join(str(m) for m in modes),
         )
 
     def __rshift__(self, other: CircuitComponent) -> CircuitComponent:
```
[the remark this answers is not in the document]
ziofil: all this is solved in the next PR

mrmustard/lab_dev/transformations/base.py
```diff
@@ -107,7 +151,28 @@ def from_bargmann(
         c = math.astensor(triple[2])
         shape_check(A, b, 2 * len(modes), "Bargmann")
```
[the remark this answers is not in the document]
ziofil: I think this is solved too in the next PR. I didn't want to do too many changes in this one

mrmustard/lab_dev/wires.py
timmysilv: just my poor understanding of MrMustard :D I figured the modes_in and modes_out would be the same in some cases, hence the double counting, but it doesn't seem that's the case

mrmustard/lab_dev/transformations/base.py
```diff
@@ -42,25 +42,69 @@ class Transformation(CircuitComponent):
...
         almost_identity = (
             self >> almost_inverse
         )  # TODO: this is not efficient, need to get c from formula
         invert_this_c = almost_identity.representation.c
-        actual_inverse = self.__class__.from_bargmann(
-            [0], (math.inv(A[0]), -math.inv(A[0]) @ b[0], 1 / invert_this_c)
+        actual_inverse = self._from_attributes(
+            Bargmann(math.inv(A[0]), -math.inv(A[0]) @ b[0], 1 / invert_this_c),
+            self.wires,
+            self.name + "_inv",
```
elib20, line 68: To confirm, this still uses the "shortcut" that we are using the Bargmann representation to store A,b,c's that don't necessarily correspond to the Bargmann representation?

mrmustard/lab_dev/transformations/base.py
```diff
@@ -42,25 +42,69 @@ class Transformation(CircuitComponent):
...
         )
         return actual_inverse
 
 
-class Unitary(Transformation):
+class Operator(Transformation):
+    r"""A CircuitComponent with input and output wires, only on the ket side.
+    This class essentially relaxes the requirement that Unitaries have that
+    the input and output modes must be the same."""
```
elib20, line 76: Can you clarify more what you mean by the input and output modes don't need to be the same? E.g. for the BtoQ map, the input and output modes attribute could be the same, it's that the representation is different on the input and output right?

mrmustard/lab_dev/transformations/base.py
```diff
@@ -107,7 +151,28 @@ def from_bargmann(
...
         shape_check(A, b, 2 * len(modes), "Bargmann")
         s = set(modes)
-        return Unitary._from_attributes(name, Bargmann(A, b, c), Wires(set(), set(), s, s))
+        return Unitary._from_attributes(
+            representation=Bargmann(A, b, c), wires=Wires(set(), set(), s, s), name=name
+        )
+
+
+class Map(Transformation):
+    r"""A CircuitComponent with input and output wires and same modes on bra and ket sides.
```
elib20, line 160: I am confused. If modes have to be the same, why do you specify modes_out and modes_in? Please reword.

elib20 (commented):
I'm a little confused by some of the docstrings, so once that's clarified it will be good to go!

mrmustard/lab_dev/transformations/base.py
ziofil: Yes, it's not an actual Bargmann representation (only on the input side). This and a few others will be updated when we have MultiRepresentation

mrmustard/lab_dev/transformations/base.py
ziofil: No, here I mean that an Operator can have e.g. input on mode [0] and output on modes [0,1]. A Unitary cannot do that.
The other open PR is on top of this one, and I'm working on docstrings in that one because some methods are moving around between the two PRs and I don't want to do things twice.

mrmustard/lab_dev/transformations/base.py
ziofil: Oh, it's the old docstring (when Channel had only `modes`), I'll go through all the docstrings in the next pr (#406)

elib20 (approved):

PR #406, opened 2024-05-30
enable new TF and pip installation

### LINKED ISSUE #381: TensorFlow dependecy needs to be fixed <2.16.0

timmysilv:
**Context:**
If you install MrMustard with `pip`, it will install the latest tensorflow, but that does not allow the use of `tf.keras.optimizers.legacy.Adam`. We need the non-legacy version, but it's [slow on M1/M2 Macs](https://stackoverflow.com/questions/77222463/is-there-a-way-to-change-adam-to-legacy-when-using-mac-m1-m2-in-tensorflow). Some manual effort is required to support the legacy optimizer with TF 2.16+, and it seems out of the door anyway. I'd advise Mac M1/M2 users to install TF 2.15 in advance if they want the better optimizer for them (until TF fixes it, at least).

- Add the `tf` extra to `tensorflow-probability`. This one's a bit hacky, because that extra doesn't exist until 0.24+, but if you are using 0.24+ (which you would be with a `pip install`), you need it to use Keras optimizers. The good news is that `poetry` doesn't seem to mind, and totally ignores it for our lockfile (which has it pinned to 0.22)
- Use the non-legacy optimizer for TF 2.16+ since it will work out of the box
- Raise a warning to advise M1/M2 Mac users to downgrade to TF 2.15 (if they're on 2.16+)

**Benefits:**
If users pip-install MrMustard, it will still work (even with TF 2.16)

**Possible Drawbacks:**
Mac M1/M2 users will have slow optimizations if they get a fresh environment and `pip install mrmustard`. They can avoid this by pre-installing TF 2.15 to their environment. I've added a warning to tell them clearly what to do.

**Related GitHub Issues:**
Fixes #381 by adding 2.16 support

### DIFF
--- a/.github/CHANGELOG.md
+++ b/.github/CHANGELOG.md
@@ -22,6 +22,9 @@
 
 * Ensure all symplectic eigenvalues are returned by the `symplectic_eigenvals` function.
 
+* Ensured support for TensorFlow 2.16+, which would be chosen when installing with `pip`.
+  [(#406)](https://github.com/XanaduAI/MrMustard/pull/406)
+
 
 ### Documentation
 
--- a/mrmustard/math/backend_tensorflow.py
+++ b/mrmustard/math/backend_tensorflow.py
@@ -408,7 +408,6 @@ class BackendTensorflow(BackendBase):  # pragma: no cover
     # ~~~~~~~~~~~~~~~~~
 
     def DefaultEuclideanOptimizer(self) -> tf.keras.optimizers.legacy.Optimizer:
-        print(f"{distribution('tensorflow').version}\n{platform.system()}\n{platform.processor()}")
         use_legacy = Version(distribution("tensorflow").version) < Version("2.16.0")
         AdamOpt = tf.keras.optimizers.legacy.Adam if use_legacy else tf.keras.optimizers.Adam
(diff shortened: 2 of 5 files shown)


ziofil (commented):
very nice!

ziofil (approved):

PR #407, opened 2024-06-01
QtoB and other improvements

ziofil:
**Context:**
We can use the inverse of BtoQ to define a QtoB map. This way the code to initialize from quadrature is drastically simplified, and it can be placed higher in the hierarchy. 

Introduces two classes: Operation <- Unitary and Map <- Channel (in the sense that Operation and Map are the parents of Unitary and Channel). This sets things up for Kraus operators and allows us to correctly type objects that are not strictly unitaries or channels.

**Benefits:**
Simpler code, more robust. 

**Possible Drawbacks:**
More classes

### DIFF
--- a/mrmustard/lab_dev/circuit_components.py
+++ b/mrmustard/lab_dev/circuit_components.py
@@ -77,7 +77,12 @@ class CircuitComponent:
         ib = tuple(sorted(modes_in_bra))
         ok = tuple(sorted(modes_out_ket))
         ik = tuple(sorted(modes_in_ket))
-        if ob != modes_out_bra or ib != modes_in_bra or ok != modes_out_ket or ik != modes_in_ket:
+        if (
+            ob != modes_out_bra
+            or ib != modes_in_bra
+            or ok != modes_out_ket
+            or ik != modes_in_ket
+        ):
             offsets = [len(ob), len(ob) + len(ib), len(ob) + len(ib) + len(ok)]
             perm = (
                 tuple(np.argsort(modes_out_bra))
@@ -160,6 +165,54 @@ class CircuitComponent:
                 f"Cannot compute triple from representation of type ``{self.representation.__class__.__qualname__}``."
             ) from e
 
+    @classmethod
+    def from_quadrature(
+        cls,
+        triple: tuple,
+        modes_out_bra: Sequence[int] = (),
+        modes_in_bra: Sequence[int] = (),
+        modes_out_ket: Sequence[int] = (),
+        modes_in_ket: Sequence[int] = (),
+        name: Optional[str] = None,
+    ) -> CircuitComponent:
+        r"""
+        Initializes a circuit component from its quadrature representation.
+
+        Args:
+            triple: The quadrature representation of the component.
+            modes_out_bra: The output modes on the bra side of this component.
+            modes_in_bra: The input modes on the bra side of this component.
+            modes_out_ket: The output modes on the ket side of this component.
+            modes_in_ket: The input modes on the ket side of this component.
+            name: The name of this component.
+            kwargs: keyword arguments like modes_in, modes_out and modes.
+
+        Returns:
+            A circuit component with the given quadrature representation.
+        """
+        from mrmustard.lab_dev.circuit_components_utils import BtoQ  # pylint: disable=import-outside-toplevel
+
+        wires = Wires(
+            set(modes_out_bra),
+            set(modes_in_bra),
+            set(modes_out_ket),
+            set(modes_in_ket),
+        )
+        Q = CircuitComponent._from_attributes(
+            Bargmann(*triple), wires
+        )  # this is actually in quadrature
+        kets_done = (
+            BtoQ(wires.input.ket.modes).inverse().dual
+            @ Q
+            @ BtoQ(wires.output.ket.modes).inverse()
+        )
+        all_done = (
+            BtoQ(wires.input.bra.modes).inverse().adjoint.dual
+            @ kets_done
+            @ BtoQ(wires.output.bra.modes).inverse().adjoint
+        )
+        return cls._from_attributes(all_done.representation, wires, name)
+
     def quadrature(self) -> tuple | ComplexTensor:
         r"""
         The quadrature representation of this circuit component.
@@ -172,7 +225,11 @@ class CircuitComponent:
         # Here for a CircuitComponent, we need to add this map four times: BtoQ on out_ket
         # wires, BtoQ.dual on in_ket wires, BtoQ.adjoint on out_bra wires and BtoQ.adjoint.dual
         # on in_bra wires.
-        kets_done = BtoQ(self.wires.input.ket.modes).dual @ self @ BtoQ(self.wires.output.ket.modes)
+        kets_done = (
+            BtoQ(self.wires.input.ket.modes).dual
+            @ self
+            @ BtoQ(self.wires.output.ket.modes)
+        )
         all_done = (
             BtoQ(self.wires.input.bra.modes).adjoint.dual
             @ kets_done
@@ -281,7 +338,9 @@ class CircuitComponent:
 
         return ret
 
-    def to_fock(self, shape: Optional[Union[int, Iterable[int]]] = None) -> CircuitComponent:
+    def to_fock(
+        self, shape: Optional[Union[int, Iterable[int]]] = None
+    ) -> CircuitComponent:
         r"""
         Returns a circuit component with the same attributes as this component, but
         with ``Fock`` representation.
@@ -360,7 +419,9 @@ class CircuitComponent:
         """
         return self.representation == other.representation and self.wires == other.wires
 
-    def _matmul_indices(self, other: CircuitComponent) -> tuple[tuple[int, ...], tuple[int, ...]]:
+    def _matmul_indices(
+        self, other: CircuitComponent
+    ) -> tuple[tuple[int, ...], tuple[int, ...]]:
         r"""
         Finds the indices of the wires being contracted on the bra and ket sides of the components.
         """
@@ -426,7 +487,9 @@ class CircuitComponent:
         wires_temp = Template(filename=os.path.dirname(__file__) + "/assets/wires.txt")  # nosec
         wires_temp_uni = wires_temp.render_unicode(wires=self.wires)
         wires_temp_uni = (
-            wires_temp_uni.replace("<body>", "").replace("</body>", "").replace("h1", "h3")
+            wires_temp_uni.replace("<body>", "")
+            .replace("</body>", "")
+            .replace("h1", "h3")
         )
 
         rep_temp = (
@@ -437,7 +500,11 @@ class CircuitComponent:
             )  # nosec
         )
         rep_temp_uni = rep_temp.render_unicode(rep=self.representation)
-        rep_temp_uni = rep_temp_uni.replace("<body>", "").replace("</body>", "").replace("h1", "h3")
+        rep_temp_uni = (
+            rep_temp_uni.replace("<body>", "")
+            .replace("</body>", "")
+            .replace("h1", "h3")
+        )
 
         display(HTML(temp.render(comp=self, wires=wires_temp_uni, rep=rep_temp_uni)))
 
--- a/mrmustard/lab_dev/circuit_components.py
+++ b/mrmustard/lab_dev/circuit_components.py
@@ -77,12 +77,7 @@ class CircuitComponent:
         ib = tuple(sorted(modes_in_bra))
         ok = tuple(sorted(modes_out_ket))
         ik = tuple(sorted(modes_in_ket))
-        if (
-            ob != modes_out_bra
-            or ib != modes_in_bra
-            or ok != modes_out_ket
-            or ik != modes_in_ket
-        ):
+        if ob != modes_out_bra or ib != modes_in_bra or ok != modes_out_ket or ik != modes_in_ket:
             offsets = [len(ob), len(ob) + len(ib), len(ob) + len(ib) + len(ok)]
             perm = (
                 tuple(np.argsort(modes_out_bra))
@@ -177,11 +172,7 @@ class CircuitComponent:
         # Here for a CircuitComponent, we need to add this map four times: BtoQ on out_ket
         # wires, BtoQ.dual on in_ket wires, BtoQ.adjoint on out_bra wires and BtoQ.adjoint.dual
         # on in_bra wires.
-        kets_done = (
-            BtoQ(self.wires.input.ket.modes).dual
-            @ self
-            @ BtoQ(self.wires.output.ket.modes)
-        )
+        kets_done = BtoQ(self.wires.input.ket.modes).dual @ self @ BtoQ(self.wires.output.ket.modes)
         all_done = (
             BtoQ(self.wires.input.bra.modes).adjoint.dual
             @ kets_done
@@ -290,9 +281,7 @@ class CircuitComponent:
 
         return ret
 
-    def to_fock(
-        self, shape: Optional[Union[int, Iterable[int]]] = None
-    ) -> CircuitComponent:
+    def to_fock(self, shape: Optional[Union[int, Iterable[int]]] = None) -> CircuitComponent:
         r"""
         Returns a circuit component with the same attributes as this component, but
         with ``Fock`` representation.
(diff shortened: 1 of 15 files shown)


timmysilv (commented):
will review the rest of the files tomorrow. also wondering, do we build docs on PRs? would be useful to validate the docstrings somehow, otherwise i'll check it out and build it locally

mrmustard/lab_dev/transformations/base.py
```diff
@@ -131,98 +150,82 @@ def __rshift__(self, other: CircuitComponent) -> CircuitComponent:
...
-        shape_check(A, b, 2 * len(modes), "Bargmann")
-        s = set(modes)
+        r"""Initialize a Unitary from the given symplectic matrix in qqpp basis.
+        I.e. the axes are ordered as [q0, q1, ..., p0, p1, ...].
+        """
+        if symplectic.shape[-2:] != (2 * len(modes), 2 * len(modes)):
+            raise ValueError(
+                "Symplectic matrix and number of modes don't match. "
+                + f"Modes imply shape {(2 * len(modes), 2 * len(modes))}, "
+                + f"but shape is {symplectic.shape[-2:]}."
```
timmysilv, line 173: no need to change here, but python pro tip - strings are auto-concatenated:
```pycon
>>> "foo""bar" == "foobar"
True
```
so you can do things like:
```python
x = (
    "my first long line "
    f"and my second longer line with variable {y}"
)
```

mrmustard/lab_dev/transformations/transformations.py
```diff
@@ -200,18 +200,18 @@ class Rgate(Unitary):
     def __init__(
         self,
         modes: Sequence[int],
-        theta: Union[float, list[float]] = 0.0,
-        theta_trainable: bool = False,
-        theta_bounds: Tuple[Optional[float], Optional[float]] = (0.0, None),
+        phi: Union[float, list[float]] = 0.0,
+        phi_trainable: bool = False,
+        phi_bounds: Tuple[Optional[float], Optional[float]] = (0.0, None),
```
timmysilv, line 205: i see there has been some doc-confusion about calling this `phi` vs `theta`, but I'm wondering now if users have been explicitly specifying this as `theta=0.1` (with the kwarg name) because if so, this will break their code. either way, if we really want to rename this set of kwargs, we should change the docstring on lines 182-183 from `theta` to `phi` as well

mrmustard/lab_dev/circuit_components.py
```diff
@@ -160,25 +199,58 @@ def bargmann(self) -> tuple:
...
-            BtoQ(self.wires.input.bra.modes).adjoint.dual
-            @ kets_done
-            @ BtoQ(self.wires.output.bra.modes).adjoint
-        )
-        return all_done.representation.data
+        BtoQ_ob = BtoQ(self.wires.output.bra.modes, phi).adjoint
+        BtoQ_ib = BtoQ(self.wires.input.bra.modes, phi).adjoint.dual
+        BtoQ_ok = BtoQ(self.wires.output.ket.modes, phi)
+        BtoQ_ik = BtoQ(self.wires.input.ket.modes, phi).dual
+        QQQQ = BtoQ_ib @ (BtoQ_ik @ self @ BtoQ_ok) @ BtoQ_ob
```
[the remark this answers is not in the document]
ziofil: yes indeed, otherwise it would start with an outer product between `BtoQ_ib` and `BtoQ_ik`

mrmustard/lab_dev/states/base.py
```diff
@@ -330,14 +332,10 @@ def phase_space(self, s: float) -> tuple:
                 The covariance matrix, the mean vector and the coefficient of the state in s-parametrized phase space.
         """
         if not isinstance(self.representation, Bargmann):
-            raise ValueError(f"Can not calculate phase space for ``{self.name}`` object.")
+            raise ValueError("Can calculate phase space only for Bargmann states.")
```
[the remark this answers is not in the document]
ziofil: No, it's that in this case we don't have a method in place yet to calculate the phase space representation (e.g. wigner function) if the starting representation is Fock.

mrmustard/lab_dev/transformations/base.py
```diff
@@ -52,14 +91,12 @@ def inverse(self) -> Transformation:
...
+        A, b, _ = self.dual.representation.conj().triple  # apply X(.)X
         almost_inverse = self._from_attributes(
-            Bargmann(math.inv(A[0]), -math.inv(A[0]) @ b[0], 1 + 0j),
-            self.wires,
-            "",
+            Bargmann(math.inv(A[0]), -math.inv(A[0]) @ b[0], 1 + 0j), self.wires
         )
         almost_identity = (
-            self >> almost_inverse
+            self @ almost_inverse
```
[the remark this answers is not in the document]
ziofil: that we call directly what `>>` would have called after a couple of if statements

mrmustard/lab_dev/circuit_components.py
```diff
@@ -96,50 +97,53 @@ def _from_attributes(
...
+        types = {"Ket", "DM", "Unitary", "Operation", "Channel", "Map"}
         for tp in cls.mro():
             if tp.__name__ in types:
                 ret = tp()
                 break
         else:
             ret = CircuitComponent()
-
-        ret._name = name or tp.__name__ + "".join(str(m) for m in sorted(wires.modes))
+        ret._name = name or cls.__name__ + "".join(str(m) for m in sorted(wires.modes))
```
ziofil, line 131:
```suggestion
        ret._name = name or tp.__name__ + "".join(str(m) for m in sorted(wires.modes))
```

mrmustard/lab_dev/circuit_components.py
```diff
@@ -116,15 +117,15 @@ def _from_attributes(
...
+        types = {"Ket", "DM", "Unitary", "Operation", "Channel", "Map"}
         for tp in cls.mro():
             if tp.__name__ in types:
                 ret = tp()
                 break
         else:
             ret = CircuitComponent()
 
-        ret._name = name or tp.__name__ + "".join(str(m) for m in sorted(wires.modes))
+        ret._name = name or cls.__name__ + "".join(str(m) for m in sorted(wires.modes))
```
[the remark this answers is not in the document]
ziofil: it was correct, changing back 👍

mrmustard/lab_dev/circuit_components_utils.py
```diff
@@ -88,37 +90,36 @@ def __init__(
...
             modes_out=modes,
             modes_in=modes,
-            name="BtoQ",
+            representation=Bargmann(*triples.bargmann_to_quadrature_Abc(len(modes))),
         )
 
-    @property
-    def representation(self) -> Bargmann:
-        return Bargmann(*triples.bargmann_to_quadrature_Abc(len(self.modes)))
+        Operation.__init__(
```
ziofil, line 119:
```suggestion
        super().__init__(
```

mrmustard/lab_dev/transformations/base.py
```diff
@@ -69,42 +106,31 @@ def inverse(self) -> Transformation:
...
 
+    def __repr__(self) -> str:
+        return super().__repr__().replace("CircuitComponent", self.__class__.__name__)
+
 
-class Operator(Transformation):
-    r"""A CircuitComponent with input and output wires, only on the ket side.
-    This class essentially relaxes the requirement that Unitaries have that
-    the input and output modes must be the same."""
+class Operation(Transformation):
```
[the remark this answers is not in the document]
ziofil: because I figured the previous name was confusing (DM are Operators too, and Operation gives more the sense of a transformation).

mrmustard/lab_dev/transformations/transformations.py
ziofil: Historically it's always been phi. Sam coded this up using theta, so I reverted it back. I don't think anyone has code written with the new `Rgate` yet for this change to be an issue.

mrmustard/lab_dev/circuit_components.py
```diff
@@ -39,8 +38,10 @@
 
 class CircuitComponent:
     r"""
-    A base class for the components (states, transformations, and measurements, or potentially
-    unphysical ``wired'' objects) that can be placed in Mr Mustard's quantum circuits.
+    A base class for the circuit components (states, transformations, measurements,
+    and any component made by combining CircuitComponents). CircuitComponents are
+    defined by their ``representation`` and ``wires`` attributes. See the ``Wires``
+    and ``Representation`` classes (and their subclasses) for more details.
```
ziofil, line 44:
```suggestion
    defined by their ``representation`` and ``wires`` attributes. See the :class:`Wires`
    and :class:`Representation` classes (and their subclasses) for more details.
```

mrmustard/lab_dev/circuit_components.py
```diff
@@ -148,10 +149,48 @@ def _add_parameter(self, parameter: Union[Constant, Variable]):
...
+        )
+        return cls._from_attributes(repr, wires, name)
+
     @property
     def bargmann(self) -> tuple:
         r"""
-        The Bargmann parametrization of this circuit component, if available.
+        The Bargmann parametrization of this component, if available.
+        It return a triple (A, b, c) such that the Bargmann representation of this component is
+        ``F(z) = c exp(1/2 z^T A z + b^T z)``.
```
ziofil:
```suggestion
        The Bargmann parametrization of this component, if available.
        It returns a triple (A, b, c) such that the Bargmann function of this component is
        :math:`F(z) = c \exp\left(\frac{1}{2} z^T A z + b^T z\right)`
```
```

Comment (ziofil) [inline mrmustard/lab_dev/circuit_components.py]:
```suggestion
        The Bargmann parametrization of this component, if available.
        It returns a triple (A, b, c) such that the Bargmann function of this component is
        :math:`F(z) = c \exp\left(\frac{1}{2} z^T A z + b^T z\right)`
```
```

File: mrmustard/lab_dev/circuit_components.py [hunk 12]
```diff
@@ -160,25 +201,56 @@ def bargmann(self) -> tuple:
...
+        modes_out_bra: Sequence[int],
+        modes_in_bra: Sequence[int],
+        modes_out_ket: Sequence[int],
+        modes_in_ket: Sequence[int],
+        triple: tuple,
+        phi: float = 0.0,
+        name: Optional[str] = None,
+    ) -> CircuitComponent:
+        r"""Returns a circuit component from the given triple (A,b,c) that parametrizes the
+        quadrature wavefunction of this component in the form ``c exp(1/2 x^T A x + b^T x)``.
```
timmysilv, line 216:
```suggestion
        quadrature wavefunction of this component in the form :math:`c * exp(1/2 x^T A x + b^T x)`.
```

mrmustard/lab_dev/states/base.py
```diff
@@ -616,112 +621,53 @@ class DM(State):
...
-        means: ComplexMatrix,
+        triple: tuple,
         name: Optional[str] = None,
-        atol_purity: Optional[float] = 1e-3,
+        s: float = 0,  # pylint: disable=unused-argument
     ) -> DM:
+        r"""
+        Initializes a density matrix from the covariance matrix, vector of means and a coefficient,
+        which parametrize the s-parametrized phase space function
+        ``coeff * exp((x-means)^T cov^-1 (x-means))``.
```
timmysilv, line 658:
```suggestion
        :math:`coeff * exp((x-means)^T cov^{-1} (x-means))`.
```

mrmustard/lab_dev/transformations/base.py
```diff
@@ -22,27 +22,68 @@
...
+        cls,
+        modes_out: Sequence[int],
+        modes_in: Sequence[int],
+        triple: tuple,
+        phi: float = 0,
+        name: Optional[str] = None,
+    ) -> Operation:
+        r"""Initialize an Operation from the given quadrature triple (A, b, c).
+        The triple parametrizes the quadrature representation of the transformation as
+        ``c * exp(0.5*x^T A x + b^T x)``."""
```
timmysilv, line 54:
```suggestion
        :math:`c * exp(0.5*x^T A x + b^T x)`."""
```

mrmustard/lab_dev/transformations/base.py
```diff
@@ -22,27 +22,68 @@
...
+    def from_bargmann(
+        cls,
+        modes_out: Sequence[int],
+        modes_in: Sequence[int],
+        triple: tuple,
+        name: Optional[str] = None,
+    ) -> Operation:
+        r"""Initialize a Transformation from the given Bargmann triple (A,b,c)
+        which parametrizes the Bargmann function of the transformation as
+        ``c * exp(0.5*z^T A z + b^T z)``."""
```
timmysilv, line 73:
```suggestion
        :math:`c * exp(0.5*z^T A z + b^T z)`."""
```

timmysilv (approved):
nice! just left some suggestions to render math in docstrings, but otherwise looks good

mrmustard/lab_dev/states/base.py
elib20, line 1: Looks like having the option to initialize Ket and DM with representations simplifies a lot! Nice!

mrmustard/lab_dev/transformations/base.py
```diff
@@ -22,27 +22,68 @@
 representation.
```
elib20, line 22: Is the second paragraph still true?

mrmustard/lab_dev/transformations/base.py
```diff
@@ -70,159 +107,134 @@ def inverse(self) -> Transformation:
...
-            return Channel._from_attributes(ret.representation, ret.wires, "")
+            return Channel._from_attributes(ret.representation, ret.wires)
         return ret
 
-    def __repr__(self) -> str:
-        return super().__repr__().replace("CircuitComponent", "Unitary")
-
     @classmethod
-    def from_bargmann(
+    def from_symplectic(
```
elib20, line 165: Do we have the theory needed to have an analogous `from_XY` for channels? That would be useful for e.g. measurement-based squeezing where we already know the XY map and want to make an MM object for it!

elib20 (commented):
Nice!

mrmustard/lab_dev/circuit_components.py
```diff
@@ -225,31 +303,40 @@ def wires(self) -> Wires:
...
         r"""
-        The ``DualView`` of this component.
+        The dual of this component obtained by conjugating the representation and swapping
+        the input and output wires. The returned object is a view of the original component which
+        applies a conjugation and a swap of the wires, but does not copy the data in memory.
         """
         return DualView(self)
 
-    def light_copy(self) -> CircuitComponent:
+    def _light_copy(self, wires: Optional[Wires] = None) -> CircuitComponent:
```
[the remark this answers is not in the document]
timmysilv: I should think that if you call a private method (especially one called `_light_copy`) and act as if it was a deep copy, it's fair to say you might get some unexpected behaviour. do we want to provide a `__deepcopy__` as well though?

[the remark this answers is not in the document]
timmysilv: oh but users don't know that it's being used behind the scenes. ignore me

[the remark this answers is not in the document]
elib20: Yes that was my question. If a user uses _light_copy directly, then yes the _ alone should mean "user beware", but I was more concerned about adjoint and dual

mrmustard/lab_dev/transformations/base.py
ziofil: Arsalan is on it! 💪

mrmustard/lab_dev/circuit_components.py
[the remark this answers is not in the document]
ziofil: Hey, sorry I'm late to the party. So, yes the `_light_copy` is intended for avoiding copies of a potentially large representation (e.g. a big fock array). As you point out, it has an underscore in the name and I would add that its behaviour is documented, so if a user uses it incorrectly it's kind of on them.

The more important part of your question I think is whether this happens in ways that could fool users, i.e. they get a DualView back from some method and they use it as a CircuitComponent. Let's keep thinking of the implications. For context, copying Bargmann data isn't a problem (Kb) as much as it is copying Fock data (which can be really large (Gb)).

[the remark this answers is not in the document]
timmysilv: my main concern here is that views should be read-only, but it seems that users have a good chance of manipulating one. If I were a user, I would not realize that the `.dual` property is a `DualView` instance - I'd suspect it's a fresh object. If we expect our users to maintain that expectation, then we should implement a sort of [copy-on-write](https://en.wikipedia.org/wiki/Copy-on-write) technique to get the best of both worlds. I'll look into adding this!

PR #410, opened 2024-06-07
Adding twomodesqueezing

kaspernielsen96:
**Context:**
Adding two mode squeezing 
Added both the two mode squeezing gate, and the two mode squeezed vacuum state


mrmustard/lab_dev/transformations/transformations.py
```diff
@@ -315,6 +315,70 @@ def representation(self) -> Bargmann:
         return Bargmann(*triples.identity_Abc(n_modes))
 
 
+class TMSgate(Unitary):
```
ziofil, line 318: I admit I like this name, but historically we were calling it `S2gate`

mrmustard/physics/triples.py
```diff
@@ -187,6 +187,37 @@ def displaced_squeezed_vacuum_state_Abc(
...
+        r: The squeezing magnitudes.
+        phi: The squeezing angles.
+
+    Returns:
+        The ``(A, b, c)`` triple of the squeezed vacuum states.
+    """
+    r, phi = list(_reshape(r=r, phi=phi))
+    n_modes = 2 * len(r)
+    O = math.zeros((len(r), len(r)), math.complex128)
+    tanhr = math.diag(-math.exp(1j * phi) * math.sinh(r) / math.cosh(r))
```
ziofil, line 212: we don't have tanh?

tests/test_lab_dev/test_states.py
```diff
@@ -769,6 +770,57 @@ def test_representation_error(self):
             SqueezedVacuum(modes=[0], r=[0.1, 0.2]).representation
 
 
+class TestTwoModeSqueezedVacuum:
```
ziofil, line 773: there are some other nice tests, like that if you trace one subsystem e.g. using `state[0]` you should get a thermal state.

ziofil (approved):
very nice! just a quick name change for consistency with previous libraries should suffice 👍

PR #411, opened 2024-06-13
rshift of Ket and DM returns scalar if no Hilbert spaces

ziofil:
Right-shifting two components so that no wires are left, like `Coherent([0], 1.0) >> Vacuum([0]).dual`, now returns a scalar, which as Jacob points out is more intuitive.

This is a tricky thing to do in code because it's easier if the return type of a method (`__rshift__` here) is always the same and not dependent on context. However, I think I did things carefully and I kept the behaviour confined to the `>>` operator: `Coherent([0], 1.0) @ Vacuum([0]).dual` returns a components with no wires and scalar representation. This way we can still use `@` internally and not worry about scalars being introduced.

There is one (possibly rare) edge case that is not possible to handle: things like `Coherent([0], 1.0) >> Vacuum([0]).dual >> Coherent([0], 1.0)`, which should just return a rescaled coherent state now gives a nonsensical result, because the `__rshift__` of the array resulting from the first `>>` gets called first and we cannot avoid that. If the scalar comes _after_ the circuit component then it works.


mrmustard/lab_dev/states/base.py
```diff
@@ -266,17 +266,16 @@ def _L2_norms(self) -> RealVector:
         The `L2` norm (squared) of a ``Ket``, or the Hilbert-Schmidt norm of a ``DM``, element-wise along the batch dimension.
         """
         settings.UNSAFE_ZIP_BATCH = True
-        rep = (self >> self.dual).representation
+        rep = self >> self.dual
         settings.UNSAFE_ZIP_BATCH = False
-        return math.real(rep.c if isinstance(rep, Bargmann) else rep.array)
+        return rep
```
ziofil, line 271:
```suggestion
        return math.real(rep)
```

tests/test_lab_dev/test_simulator.py
```diff
@@ -53,6 +58,6 @@ def test_run(self, path):
         res = Simulator().run(circuit)
         exp = vac12 >> d1 >> d2 >> d12 >> a1 >> n12
 
-        assert res == exp
+        assert np.isclose(res, exp)
```
timmysilv, line 61: not super concerned, but do you know why this isn't exact anymore?

timmysilv (approved):
looks good! I can't comment on what users would find more intuitive, but I _think_ this makes sense. I'm generally hesitant to have functions return different types with different inputs, but with sufficient docs/examples that should be fine

mrmustard/lab_dev/states/base.py
```diff
@@ -266,17 +266,16 @@ def _L2_norms(self) -> RealVector:
...
+        return math.real(rep)
 
     @property
     def L2_norm(self) -> float:
         r"""
         The `L2` norm (squared) of a ``Ket``, or the Hilbert-Schmidt norm of a ``DM``.
         """
-        rep = (self >> self.dual).representation
-        return math.sum(math.real(rep.scalar))
+        return math.sum(math.real(self >> self.dual))
```
elib20, line 278: Do we want to specify a MM type for what this returns?

elib20 (commented):
Nice! I'm a bit confused about how batch dimensions and typing will work here.

mrmustard/lab_dev/states/base.py
```diff
@@ -266,17 +266,16 @@ def _L2_norms(self) -> RealVector:
         The `L2` norm (squared) of a ``Ket``, or the Hilbert-Schmidt norm of a ``DM``, element-wise along the batch dimension.
         """
         settings.UNSAFE_ZIP_BATCH = True
-        rep = (self >> self.dual).representation
+        rep = self >> self.dual
         settings.UNSAFE_ZIP_BATCH = False
-        return math.real(rep.c if isinstance(rep, Bargmann) else rep.array)
+        return math.real(rep)
```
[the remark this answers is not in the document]
ziofil: The typing conventions are trying to do too much (capturing type, dtype and shape). We are all waiting for python 3.12 to be fully compatible with all the libraries to take advantage of the new features in the typing module which will simplify the way we indicate types in MM. While we wait for that we ought to explain things well in the docstrings.

mrmustard/lab_dev/states/base.py
ziofil: the return type is actually `float`

tests/test_lab_dev/test_circuit_components_utils.py
```diff
@@ -58,12 +58,8 @@ def test_trace_out_bargmann_states(self):
         assert state >> TraceOut([0]) == Coherent([1, 2], x=1).dm()
         assert state >> TraceOut([1, 2]) == Coherent([0], x=1).dm()
 
-        no_state = state >> TraceOut([0, 1, 2])
-        assert no_state.modes == []
-        assert no_state.wires == Wires()
-        assert np.allclose(no_state.representation.A, [])
-        assert np.allclose(no_state.representation.b, [])
-        assert np.allclose(no_state.representation.c, 1)
+        trace = state >> TraceOut([0, 1, 2])
```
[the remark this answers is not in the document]
ziofil: it isn't, it's a float

tests/test_lab_dev/test_simulator.py
ziofil: it is exact to float precision (1e-18), it's just that one shouldn't compare floats with ==

mrmustard/physics/triples.py
```diff
@@ -546,39 +546,31 @@ def fock_damping_Abc(n_modes: int) -> Union[Matrix, Vector, Scalar]:
...
+         phi: The quadrature angle. 0 corresponds to the `x` quadrature, and :math:`\pi/2` to the `p` quadrature.
 
     Returns:
         The ``(A, b, c)`` triple of the map from bargmann representation with ABC Ansatz to quadrature representation with ABC Ansatz.
     """
     hbar = settings.HBAR
-    In = math.eye(n_modes, math.complex128)
-    A = math.block(
+    Id = np.eye(n_modes, dtype=np.complex128)
+    e = np.exp(1j * phi / 2)
```
timmysilv, line 565: can the tensorflow backend not benefit from this?

ziofil: No, this is not a trainable gate.

ziofil:
> looks good! I can't comment on what users would find more intuitive, but I _think_ this makes sense. I'm generally hesitant to have functions return different types with different inputs, but with sufficient docs/examples that should be fine

I 100% agree. Now there exist edge cases (probably not too relevant, see explanation above), but still. We need to be careful going forward.

ziofil:
> Nice! I'm a bit confused about how batch dimensions and typing will work here.

Batch dimension is kept:
```python
(Coherent([0], 1.0) + Coherent([0],[-2.0]) >> Vacuum([0]).dual) 
# array([0.6065+0.j, 0.1353+0.j])
```

elib20:
> There is one (possibly rare) edge case that is not possible to handle: things like Coherent([0], 1.0) >> Vacuum([0]).dual >> Coherent([0], 1.0), which should just return a rescaled coherent state now gives a nonsensical result, because the __rshift__ of the array resulting from the first >> gets called first and we cannot avoid that. If the scalar comes after the circuit component then it works.

Would `(Coherent([0], 1.0) >> Vacuum([0]).dual) * Coherent([0], 1.0)` work? If so, I think that's good. We wouldn't want the `>>` to also mean multiplication by a scalar right?

mrmustard/lab_dev/circuit_components_utils.py
```diff
@@ -71,6 +70,28 @@ def __init__(
...
+        We know that the trace in Bargmann is a Gaussian integral, and in
+        Fock it's a trace (rather than an inner product with the identity).
+        So we write two shortcuts here, and ``__rrshift__`` will be called first if
+        present in the ``__rshift__`` method of the first object (``other`` here).
+        """
+        ket = other.wires.output.ket
+        bra = other.wires.output.bra
+        idx_zconj = [bra[m].indices[0] for m in self.wires.modes & bra.modes]
+        idx_z = [ket[m].indices[0] for m in self.wires.modes & ket.modes]
+        if not ket or not bra:
```
elib20, line 85: If there's a possibility for ket or bra to not exist, what does bra[m] and ket[m] do in the lines above?

mrmustard/lab_dev/circuit_components_utils.py
```diff
@@ -97,8 +118,9 @@ def __init__(
 
 
 class BtoQ(Operation):
-    r"""The kernel for the change of representation from ``Bargmann`` into quadrature.
-    By default it's defined on the output ket side.
+    r"""The Operation that changse the representation of an object from ``Bargmann`` into quadrature.
```
elib20, line 121:
```suggestion
    r"""The Operation that changes the representation of an object from ``Bargmann`` into quadrature.
```

mrmustard/lab_dev/states/base.py
```diff
@@ -656,8 +656,10 @@ def from_phase_space(
         r"""
         Initializes a density matrix from the covariance matrix, vector of means and a coefficient,
         which parametrize the s-parametrized phase space function
+        ``coeff * exp((x-means)^T cov^-1 (x-means))``.
```
elib20, line 659: Should there be a -1/2 in the exponential? Otherwise, I worry the function isn't bounded. the `:math: `below should also be changed if this is true.

```suggestion
        ``coeff * exp((-1/2)*(x-means)^T cov^-1 (x-means))``.
```

mrmustard/lab_dev/states/base.py
```diff
@@ -726,17 +728,17 @@ def expectation(self, operator: CircuitComponent):
...
+            result = self >> operator.dual
             if leftover_modes:
                 result >>= TraceOut(leftover_modes)
         else:
             result = (self @ operator) >> TraceOut(self.modes)
 
-        return result.representation.scalar
+        return result
 
     def __rshift__(self, other: CircuitComponent) -> CircuitComponent:
```
elib20, line 743: Update the docstring to reflect the new possible returns.

mrmustard/lab_dev/states/base.py
```diff
@@ -937,13 +941,12 @@ def __rshift__(self, other: CircuitComponent) -> CircuitComponent:
         with those of a ``DM`` or of a ``Ket``. Returns a ``CircuitComponent`` otherwise.
```
elib20, line 941: Update the docstring of this method to reflect new possible return type

tests/test_lab_dev/test_circuit_components_utils.py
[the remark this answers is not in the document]
elib20: The documentation says it can be complex? Do we want to have a test where the answer can be complex too?

mrmustard/lab_dev/states/base.py
```diff
@@ -1037,10 +1036,10 @@ def __rshift__(self, other: CircuitComponent) -> CircuitComponent:
...
         if not ret.wires.input:
             if not ret.wires.bra:
-                return Ket._from_attributes(ret.representation, ret.wires, "")
-            if ret.wires.bra.modes == ret.wires.ket.modes:
-                return DM._from_attributes(ret.representation, ret.wires, "")
-        return ret
+                ret = Ket._from_attributes(ret.representation, ret.wires, "")
+            elif ret.wires.bra.modes == ret.wires.ket.modes:
+                ret = DM._from_attributes(ret.representation, ret.wires, "")
+        return ret.representation.scalar if len(ret.wires) == 0 else ret
```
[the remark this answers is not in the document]
elib20: Please remember to make the changes before marking comments as resolved.

mrmustard/lab_dev/circuit_components.py
```diff
@@ -497,20 +515,35 @@ def __rshift__(self, other: CircuitComponent) -> CircuitComponent:
...
-            return (self @ other) @ other.adjoint
+            ret = (self @ other) @ other.adjoint
+            return ret.representation.scalar if len(ret.wires) == 0 else ret
 
         raise ValueError(msg)
 
+    def __rrshift__(self, other: Scalar) -> CircuitComponent | np.array:
+        r"""
+        Multiplies a scalar with a circuit component when written as ``scalar >> component``.
+        This is needed when the "component" on the left is the result of a contraction that leaves
```
elib20, line 538: on the left or the right?

mrmustard/lab_dev/circuit_components.py
```diff
@@ -497,20 +515,35 @@ def __rshift__(self, other: CircuitComponent) -> CircuitComponent:
...
         other_needs_bra = (self.wires.bra and self.wires.ket) and not other.wires.bra
         other_needs_ket = (self.wires.bra and self.wires.ket) and not other.wires.ket
         if other_needs_bra or other_needs_ket:
-            return (self @ other) @ other.adjoint
+            ret = (self @ other) @ other.adjoint
+            return ret.representation.scalar if len(ret.wires) == 0 else ret
 
         raise ValueError(msg)
 
+    def __rrshift__(self, other: Scalar) -> CircuitComponent | np.array:
```
elib20, line 535: I can't figure out where this is ever called?

elib20 (changes requested):
Please make sure that docstrings reflect latest changes.

I have some confusion over `__rrshift__`, and I think we should avoid overloading rshift to also do multiplication if possible.

elib20:
> > Nice! I'm a bit confused about how batch dimensions and typing will work here.
> 
> Batch dimension is kept:
> 
> ```python
> (Coherent([0], 1.0) + Coherent([0],[-2.0]) >> Vacuum([0]).dual) 
> # array([0.6065+0.j, 0.1353+0.j])
> ```

Is this what we expect? If a user wants to make a superposition of two coherent states, then start taking inner products of that (displaced) cat state, wouldn't we want that to return a scalar?

mrmustard/lab_dev/circuit_components.py
```diff
@@ -463,31 +464,48 @@ def _matmul_indices(self, other: CircuitComponent) -> tuple[tuple[int, ...], tup
...
         Contracts ``self`` and ``other`` (output of self going into input of other).
-        It adds the adjoints when they are missing. An error is raised if these
-        cannot be deduced from the wires of the components. For example this
-        allows ``Ket``s to be right-shifted into ``Channel``s and automatically
-        the result is a ``DM``:
+        It adds the adjoints when they are missing (e.g. if ``self`` is a Ket and
+        ``other`` is a Channel). An error is raised if these cannot be deduced from
+        the wires of the components. For example this allows ``Ket``s to be right-shifted
+        into ``Channel``s and automatically the result is a ``DM``. If the result has
+        no wires left, it returns the (batched) scalar value of the representation.
```
ziofil, line 493:
```suggestion
        no wires left, it returns the (batched) scalar value of the representation.
        Note that a ``CircuitComponent`` is allowed to right-shift into scalars because the scalar part may result from an automated contraction subroutine that involves several components).
```

mrmustard/lab_dev/circuit_components.py
```diff
@@ -463,31 +464,49 @@ def _matmul_indices(self, other: CircuitComponent) -> tuple[tuple[int, ...], tup
...
         return CircuitComponent._from_attributes(rep, wires_result, None)
 
-    def __rshift__(self, other: CircuitComponent) -> CircuitComponent:
+    def __rmatmul__(self, other: Scalar) -> CircuitComponent:
+        r"""
+        Multiplies a scalar with a circuit component when written as ``scalar @ component``.
+        """
+        return self * other
+
+    def __rshift__(self, other: CircuitComponent) -> CircuitComponent | np.ndarray:
```
ziofil, line 486:
```suggestion
    def __rshift__(self, other: CircuitComponent | numbers.Number) -> CircuitComponent | np.ndarray:
```

mrmustard/lab_dev/circuit_components.py
ziofil: It's called by python when the object on the left has no `__rshift__`. Then python looks for ``__rrshift__`` on the object on the right.

mrmustard/lab_dev/circuit_components.py
ziofil: left.

mrmustard/lab_dev/circuit_components.py
```diff
@@ -497,20 +515,35 @@ def __rshift__(self, other: CircuitComponent) -> CircuitComponent:
...
 
+    def __rrshift__(self, other: Scalar) -> CircuitComponent | np.array:
+        r"""
+        Multiplies a scalar with a circuit component when written as ``scalar >> component``.
+        This is needed when the "component" on the left is the result of a contraction that leaves
+        no wires and the component is returned as a scalar.
+        """
+        ret = self * other
+        try:
+            return ret.representation.scalar
```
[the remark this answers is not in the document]
ziofil: No, because if `other` has an Abc triple, then its ``__rshift__`` exists and it's called before. This is for objects on the left that don't have rshift like the float `1.414`.

mrmustard/lab_dev/circuit_components_utils.py
ziofil: Then they return a `Wires` object with no wires. You could try, you know.

mrmustard/lab_dev/circuit_components.py
```diff
@@ -497,20 +516,35 @@ def __rshift__(self, other: CircuitComponent) -> CircuitComponent:
...
+            ret = (self @ other) @ other.adjoint
+            return ret.representation.scalar if len(ret.wires) == 0 else ret
 
         raise ValueError(msg)
 
+    def __rrshift__(self, other: Scalar) -> CircuitComponent | np.array:
+        r"""
+        Multiplies a scalar with a circuit component when written as ``scalar >> component``.
+        This is needed when the "component" on the left is the result of a contraction that leaves
+        no wires and the component is returned as a scalar.
```
ziofil, line 540:
```suggestion
        no wires and the component is returned as a scalar. Note that there is an edge case if the object on the left happens to have the ``__rshift__`` method, but it's not the one we want (usually `>>` is about bit shifts) like a numpy array. In this case in an expression with types ``np.ndarray >> CircuitComponent`` the method ``__rrshift__`` will not be called, and something else will be returned.
```

mrmustard/lab_dev/states/base.py
```diff
@@ -750,10 +751,12 @@ def __rshift__(self, other: CircuitComponent) -> CircuitComponent:
         those of a ``DM``, a ``CircuitComponent`` otherwise.
```
ziofil, line 751:
```suggestion
        those of a ``DM``, a ``CircuitComponent`` otherwise, and a scalar if there are no wires left.
```

mrmustard/lab_dev/states/base.py
```diff
@@ -938,10 +941,12 @@ def __rshift__(self, other: CircuitComponent) -> CircuitComponent:
         with those of a ``DM`` or of a ``Ket``. Returns a ``CircuitComponent`` otherwise.
```
ziofil, line 941:
```suggestion
        with those of a ``DM`` or of a ``Ket``. Returns a ``CircuitComponent`` otherwise.
```
```suggestion
        with those of a ``DM`` or of a ``Ket``. Returns a ``CircuitComponent`` otherwise and a scalar if there are no wires left.
```

mrmustard/lab_dev/states/base.py
```diff
@@ -935,13 +938,15 @@ def __rshift__(self, other: CircuitComponent) -> CircuitComponent:
         not needed and the method returns a new ``Ket``.
 
         Returns a ``DM`` or a ``Ket`` when the wires of the resulting components are compatible
-        with those of a ``DM`` or of a ``Ket``. Returns a ``CircuitComponent`` otherwise.
+        with those of a ``DM`` or of a ``Ket``. Returns a ``CircuitComponent`` otherwise and a scalar if there are no wires left.
```
ziofil, line 941:
```suggestion
        with those of a ``DM`` or of a ``Ket``. Returns a ``CircuitComponent`` otherwise and a (batched) scalar if there are no wires left.
```

mrmustard/lab_dev/states/base.py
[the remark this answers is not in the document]
ziofil: done

tests/test_lab_dev/test_circuit_components_utils.py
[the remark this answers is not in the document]
ziofil: added a test.

ziofil:
> > > Nice! I'm a bit confused about how batch dimensions and typing will work here.
> > 
> > 
> > Batch dimension is kept:
> > ```python
> > (Coherent([0], 1.0) + Coherent([0],[-2.0]) >> Vacuum([0]).dual) 
> > # array([0.6065+0.j, 0.1353+0.j])
> > ```
> 
> Is this what we expect? If a user wants to make a superposition of two coherent states, then start taking inner products of that (displaced) cat state, wouldn't we want that to return a scalar?

You are right, in that case we would want a scalar. I wanted to keep the batch dimension to allow users to do what they want, but we can also assume they always want to sum the inner products. I can change that.

elib20 (approved):
Couple last questions/minor modifications to answer, but once they are answered/modified, we should be good to go. Thanks!

mrmustard/lab_dev/circuit_components.py
ziofil: That's why I'm using quotes, it's not a `CircuitComponent`, it's something that doesn't have `__rshift__`. The thing on the right is a `CircuitComponent`.

mrmustard/lab_dev/circuit_components_utils.py
ziofil: 1. when you get subsets of wires using `.ket`, `.bra`, `.output` etc... and `__getitem__` (e.g. `wires[2,3,4]`) you get back a `Wires` object, regardless of how many wires it contains, so it can have zero.
2. Wires have a `__bool__` method that allows you to use the entire object as a bool which `False` if there are no wires, so `not bra` if `bra` is a Wires object is triggered when there are no wires on the bra side. So `not ket or not bra` is triggered when an object only has wires on one side, and we need to add the other side.

elib20: Cool! Thanks.

mrmustard/lab_dev/states/base.py
```diff
@@ -657,8 +658,10 @@ def from_phase_space(
         r"""
         Initializes a density matrix from the covariance matrix, vector of means and a coefficient,
         which parametrize the s-parametrized phase space function
+        ``coeff * exp((-1/2)*(x-means)^T cov^-1 (x-means))``.
```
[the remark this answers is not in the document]
ziofil: actually this is a repeated line. I'll remove one.

PR #412, opened 2024-06-13
easy cast

ziofil:
Allows circuit component with a class initializer to take another circuit component as only argument and cast it to its type.
For example:
```python
cc = Some >> Circuit  # say this returns a CircuitComponent
channel = Channel(cc)   # fails if wires are incompatible
```


mrmustard/lab_dev/wires.py
```diff
@@ -310,6 +310,38 @@ def dual(self) -> Wires:
...
+        return not self.args[0] and not self.args[1]
+
+    @cached_property
+    def map_like(self) -> bool:
+        r"Returns ``True`` if the wires are compatible with a map."
+        return self.args[0] == self.args[2] and self.args[1] == self.args[3]
+
+    @cached_property
+    def u_like(self) -> bool:
+        return self.op_like and len(self.args[2]) == len(self.args[3])
```
ziofil, line 335:
```suggestion
    def u_like(self) -> bool:
        r"Returns ``True`` if the wires are compatible with a unitary."
        return self.op_like and len(self.args[2]) == len(self.args[3])
```

mrmustard/lab_dev/wires.py
```diff
@@ -310,6 +310,38 @@ def dual(self) -> Wires:
...
+
+    @cached_property
+    def channel_like(self) -> bool:
+        return (
```
ziofil, line 339:
```suggestion
    def channel_like(self) -> bool:
        r"Returns ``True`` if the wires are compatible with a channel."
        return (
```

ziofil:
falling too far back behind, not really needed. Closing.

PR #414, opened 2024-06-18
fixed isinstance check

ziofil:
An `isinstance` check fails on some machines. Hopefully this fixes it.


timmysilv (approved):
looks good! we can look into thorough 3.9 testing too if necessary

PR #421, opened 2024-06-19
skip training on numpy, improve drawing

timmysilv:
**Context:**
The drawing tool is not very useful today, so I touched it up a bit by adding `plotly` support.

1. Skip all `test_training` tests when using `numpy` because they do nothing
2. Add the `plotly` and `plotly-grouped` modes to the test visualizer because it's much better
3. filter out `test_about` and any tests that always return 0.0 (in other words, that always finish in 50ms or less) because we can't get anything useful from that

**Benefits:**
- Easier to glean some information from the visualizer
- No time wasted running silly numpy tests that do nothing

**Possible Drawbacks:**
N/A

<details>
<summary>How to use it (from a notebook)</summary>

```
# authenticate with AWS first!
!aws s3 sync s3://<top-secret-bucket-name>/ /some/path/you/choose/
%run .github/workflows/scripts/visualize_timings.py /some/path/you/choose/numpy_tests/develop/
%run .github/workflows/scripts/visualize_timings.py /some/path/you/choose/tf_tests/develop/ --mode plotly-grouped --ncols 2 --short-name
```
</details>

### DIFF
--- a/.github/CHANGELOG.md
+++ b/.github/CHANGELOG.md
@@ -15,6 +15,7 @@
 * Save pytest timings to an S3 bucket for regression analysis. Also add a script to help
   visualize the timing results quickly.
   [(#404)](https://github.com/XanaduAI/MrMustard/pull/404)
+  [(#421)](https://github.com/XanaduAI/MrMustard/pull/421)
 
 ### Bug fixes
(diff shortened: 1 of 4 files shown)


ziofil (approved):
Nice! Thanks

PR #422, opened 2024-06-21
BtoQ fix

kaspernielsen96:
**Context:**
Fixing a mistake in BtoQ which resulted in swapping q (sometimes called x) and p quadratures. 
Updated the BtoQ triples and added a test which can catch this mistake. Also added a test that can catch swapping p and -p (and q and -q) 
**Benefits:**


ziofil (approved):
super!

PR #425, opened 2024-06-26
improve visualizations in lab_dev

### LINKED ISSUE #423: State visualizations are not shown

### Expected behavior

When calling .visualize_2d() I expect a state to be shown 

### Actual behavior

A blank box is displayed instead of a figure. If I add the keyword argument return_fig=True, the figure gets shown as desired. 

image

When e.g. calling a state in a jupyter notebook, also nothing is shown:
image

### Reproduces how often

100 % for me

### System information

```shell
Mr Mustard: a differentiable bridge between phase space and Fock space.
Copyright 2021 Xanadu Quantum Technologies Inc.

Python version:            3.9.19
Platform info:             Windows-10-10.0.19045-SP0
Installation path:         
c:\Users\jacob\miniconda3\envs\mrmustard_dev\lib\site-packages\mrmustard
Mr Mustard version:        0.7.3
Numpy version:             1.26.4
Numba version:             0.59.1
Scipy version:             1.13.1
The Walrus version:        0.19.0
TensorFlow version:        2.16.1
```

### Source code

_No response_

### Tracebacks

_No response_

### Additional information

_No response_

timmysilv:
**Context:**
There have been some issues with visualizations in MrMustard, and this should address a good amount of them

- Remove mako-based implementations for `_repr_html_`, and replace them with ipywidgets-based implementations with `_ipython_display_`. I mainly did this because ipywidgets is easier to work with than raw HTML, and should be more portable (eg. jupyterlab vs VS Code). Also, the switch to `_ipython_display_` allows for side-effects, and doesn't also print the default repr (unless I do it explicitly)
- Simplified the funky HTML+CSS to a consistent and basic look. Easier to update, and not distracting at all.
- fix the default `return_fig=False` option for `State.visualize_*()` methods by not converting to HTML manually
- some little things to please codefactor

**Benefits:**
- Works everywhere!
- Easier to maintain
- Uses highly trusted ipywidgets package
- No more double-repr

**Possible Drawbacks:**
- N/A

**Related GitHub Issues:**
Fixes #423


mrmustard/lab_dev/circuit_components.py
```diff
@@ -454,7 +453,7 @@ def to_fock(self, shape=None):
         fock._original_bargmann_data = self.representation.data
         return self._from_attributes(fock, self.wires, self.name)
 
-    def auto_shape(self, **kwargs) -> tuple[int, ...]:
+    def auto_shape(self, **_) -> tuple[int, ...]:
```
apchytr, line 456: Hmm shouldnt the kwargs be removed if its not used?

timmysilv: perhaps.. tbh I'm not sure why codefactor didn't catch this in Filippo's PR so I just made a judgement call here. I decided to do this because DM and Ket (child classes of this one) re-implement it, and I don't think we need a type error if somehow the super() implementation gets called with the kwargs propagated.

ziofil: I added them because auto_shape exists in Ket, DM and CircuitComponent, but in CircuitComponent it cannot use the arguments that can be passed to the version in Ket and DM, but I didn't want it to hang in case the args are kept and it's called on non-states. I won't die on this hill 😅

apchytr: Looks like the proper thing to do is to just keep the **kwargs https://stackoverflow.com/questions/6034662/python-method-overriding-does-signature-matter

mrmustard/lab_dev/circuit_components.py
```diff
@@ -701,6 +697,7 @@ class CCView(CircuitComponent):
 
     @property
     def short_name(self):
+        """A shortened version of the component name."""
```
apchytr, line 700:
```suggestion
        r"""
        A shortened version of the component name.
        """
```

mrmustard/lab_dev/circuit_components.py
timmysilv: will leave it as-is then 👍 (I changed it from `kwargs` to `_` because pylint complains that you're not using it, and underscore is the conventional placeholder var name)

mrmustard/lab_dev/circuit_components.py
timmysilv: is this preferred for any particular reason?

apchytr: Mostly just personal preference I find starting on a new line makes things look marginally cleaner. But I'm happy as long as all the docs are consistent between each other 😄

ziofil (approved):
very nice! thanks :)

ziofil (approved):
👍

PR #426, opened 2024-06-27
Unitary override inverse 

apchytr:
**Context:**
`Transformation`'s `inverse()` method numerically computes the inverse. We can avoid this on `Unitary` by calling `self.dual`.

`Unitary` now overrides `inverse()` to call `self.dual`. Additionally, small formatting changes are included and a fix to the Makefile.

### DIFF
--- a/mrmustard/lab_dev/transformations/base.py
+++ b/mrmustard/lab_dev/transformations/base.py
@@ -1,4 +1,4 @@
-# Copyright 2023 Xanadu Quantum Technologies Inc.
+# Copyright 2024 Xanadu Quantum Technologies Inc.
 
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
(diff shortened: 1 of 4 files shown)


mrmustard/lab_dev/transformations/base.py
```diff
@@ -1,4 +1,4 @@
-# Copyright 2023 Xanadu Quantum Technologies Inc.
+# Copyright 2024 Xanadu Quantum Technologies Inc.
```
ziofil, line 1: I think we should leave the original year, otherwise copyright is covered only from 2024

mrmustard/lab_dev/transformations/base.py
```diff
@@ -184,9 +194,13 @@ def from_symplectic(
             name=name,
         )
 
+    def inverse(self) -> Transformation:
```
ziofil, line 197:
```suggestion
    def inverse(self) -> Unitary:
```

ziofil (approved):
Nice :)

mrmustard/lab_dev/transformations/base.py
apchytr: Changed to `DualView` as this is the return type of `self.dual`

PR #427, opened 2024-06-27
adds method to contract schwinger BS

ziofil:
We can now apply a BS directly on an array via the Schwinger representation, without having to create the 4D array of the BS: just the 2D arrays of the unitaries in the SU(2) irreps :)

### DIFF
--- a/mrmustard/math/lattice/strategies/beamsplitter.py
+++ b/mrmustard/math/lattice/strategies/beamsplitter.py
@@ -184,7 +184,9 @@ def beamsplitter_schwinger(shape, theta, phi, max_N=None):
     """
     c1, c2, c3, c4 = shape
     if c1 != c3 or c2 != c4:
-        raise ValueError("The Schwinger method only supports shapes of the form (i,k,i,k).")
+        raise ValueError(
+            "The Schwinger method only supports shapes of the form (i,k,i,k)."
+        )
     # create output tensor
     U = np.zeros(shape, dtype="complex128")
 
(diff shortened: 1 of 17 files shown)


mrmustard/math/lattice/strategies/beamsplitter.py
```diff
@@ -202,3 +202,74 @@ def beamsplitter_schwinger(shape, theta, phi, max_N=None):
...
+    11           3,2  |-- 5 photons
+    14           2,3  |
+
+    15           3,3  |-- 6 photons
+
+    The left column is the flattened order. This function returns the indices in
+    left column for the N-th block. E.g. sector_idx(3, (4,4)) is [3,6,9,12].
+    """
+    return [
+        np.ravel_multi_index((i, N - i), shape) for i in range(N + 1) if max(i, N - i) < max(shape)
```
apchytr, line 245: Is the `if max(i, N - i) < max(shape)` intended to handle out of bound indices? If so, could it be handled by the `mode` param in `np.ravel_multi_index` instead?

ziofil: Good question, I didn't think of `mode`. I had a look, but it doesn't seem any of the modes allows you to skip some indices?

ziofil: Oh I see, we could use `mode='clip'` and then filter out the zeros

mrmustard/math/lattice/strategies/beamsplitter.py
```diff
@@ -202,3 +202,74 @@ def beamsplitter_schwinger(shape, theta, phi, max_N=None):
...
+    # step 1: reshape the pair of indices to which the BS is attached
+    order = [k for k in range(array.ndim) if k not in [i, j]] + [i, j]
+    array = array.transpose(order)  # move the indices to the end
+    shape_rest, shape = array.shape[:-2], array.shape[-2:]
+    array = array.reshape(shape_rest + (-1,))  # flatten the last two dimensions
+    # step 2: apply each unitary to the corresponding indices
+    for N in range(sum(shape) - 1):  # is this right?
+        flat_idx = sector_idx(N, shape)
+        u = sector_u(N, theta, phi)
+        subset = [k for k in range(N + 1) if k < shape[0] and N - k < shape[1]]
```
ziofil, line 270: @apchytr this too then can probably be done by `np` then

apchytr: @ziofil It's possible although unclear to me which method can be used to substitute this behaviour

ziofil: I think another call to rave/unravel should do it. thanks again for the idea :)

tests/test_math/test_lattice.py
```diff
@@ -108,3 +112,18 @@ def test_diagonalbatchNumba_vs_diagonalNumba(batch_size):
 
     for nb in range(batch_size):
         assert np.allclose(G_ref, G_batched[:, :, :, nb])
+
+
+def test_bs_schwinger():
```
apchytr, line 117: Can you please add unit tests for `sector_u` and `sector_idx`?

mrmustard/math/lattice/strategies/beamsplitter.py
ziofil: hmm I tried, but it clips the upper boundary to the max index, so not zero. I'm afraid `mode` doesn't help.

mrmustard/math/lattice/strategies/beamsplitter.py
ziofil: see above

mrmustard/math/lattice/strategies/beamsplitter.py
```diff
@@ -202,3 +202,74 @@ def beamsplitter_schwinger(shape, theta, phi, max_N=None):
...
+
+    The left column is the flattened order. This function returns the indices in
+    left column for the N-th block. E.g. sector_idx(3, (4,4)) is [3,6,9,12].
+    """
+    return [
+        np.ravel_multi_index((i, N - i), shape) for i in range(N + 1) if max(i, N - i) < max(shape)
+    ]
+
+
+def sector_u(N, theta, phi):
```
apchytr, line 249: Can you include the type hints for the arguments here (in `apply_BS_schwinger` also)

PR #428, opened 2024-07-02
shape of c

kaspernielsen96:
**Context:**
Making c in the Abc triple ready for polynomials
Updating the shape of c, such that the first index is always the batch dimension which is the same way that A and b are currently implemented. The following dimensions will be the (potentially multivariate) polynomial. 
If the shape of c is `(batch,1)`, it correspinds to c just being a scalar.
**Benefits:**
Prepares for the updated poly exp ansatz.
**Possible Drawbacks:**
Involves changes to a number of central mm functions, which may break some parts of mm.

### DIFF
--- a/mrmustard/math/lattice/strategies/vanilla.py
+++ b/mrmustard/math/lattice/strategies/vanilla.py
@@ -53,8 +53,8 @@ def vanilla(shape: tuple[int, ...], A, b, c) -> ComplexTensor:  # pragma: no cov
     next(index_u_iter)
 
     # write vacuum amplitude
-    # ret[0] = c
-    ret[0] = 1
+    ret[0] = c
+
 
     # iterate over the rest of the indices
     for index_u in index_u_iter:
@@ -73,7 +73,7 @@ def vanilla(shape: tuple[int, ...], A, b, c) -> ComplexTensor:  # pragma: no cov
             value_at_index += A[i, j] * np.sqrt(index_u[j]) * ret[n]
         ret[index] = value_at_index / np.sqrt(index_u[i])
 
-    return ret.reshape(shape) * c
+    return ret.reshape(shape)
 
 
 @njit
--- a/mrmustard/math/lattice/strategies/vanilla.py
+++ b/mrmustard/math/lattice/strategies/vanilla.py
@@ -53,7 +53,8 @@ def vanilla(shape: tuple[int, ...], A, b, c) -> ComplexTensor:  # pragma: no cov
     next(index_u_iter)
 
     # write vacuum amplitude
-    ret[0] = c
+    # ret[0] = c
+    ret[0] = 1
 
     # iterate over the rest of the indices
     for index_u in index_u_iter:
@@ -72,7 +73,7 @@ def vanilla(shape: tuple[int, ...], A, b, c) -> ComplexTensor:  # pragma: no cov
             value_at_index += A[i, j] * np.sqrt(index_u[j]) * ret[n]
         ret[index] = value_at_index / np.sqrt(index_u[i])
 
-    return ret.reshape(shape)
+    return ret.reshape(shape) * c
 
 
 @njit
--- a/mrmustard/math/lattice/strategies/vanilla.py
+++ b/mrmustard/math/lattice/strategies/vanilla.py
@@ -55,7 +55,6 @@ def vanilla(shape: tuple[int, ...], A, b, c) -> ComplexTensor:  # pragma: no cov
     # write vacuum amplitude
     ret[0] = c
 
-
     # iterate over the rest of the indices
     for index_u in index_u_iter:
         # update index
--- a/mrmustard/physics/ansatze.py
+++ b/mrmustard/physics/ansatze.py
@@ -394,20 +394,10 @@ class PolyExpAnsatz(PolyExpBase):
             The tensor product of this ansatz and other.
         """
 
-        def andc(c1, c2):
-            if c1.shape == (1,) and c2.shape == (1,):
-                c3 = math.outer(c1, c2).reshape(-1)
-            elif c1.shape == (1,):
-                c3 = math.outer(c1, c2).reshape(c2.shape)
-            elif c2.shape == (1,):
-                c3 = math.outer(c1, c2).reshape(c1.shape)
-            else:
-                c3 = math.outer(c1, c2).reshape(c1.shape + c2.shape)
-            return c3
 
         As = [math.block_diag(a1, a2) for a1 in self.A for a2 in other.A]
         bs = [math.concat([b1, b2], axis=-1) for b1 in self.b for b2 in other.b]
-        cs = [andc(c1, c2) for c1 in self.c for c2 in other.c]
+        cs = [math.outer(c1, c2).reshape(-1) for c1 in self.c for c2 in other.c]
         return self.__class__(As, bs, cs)
 
 
--- a/mrmustard/physics/ansatze.py
+++ b/mrmustard/physics/ansatze.py
@@ -393,9 +393,20 @@ class PolyExpAnsatz(PolyExpBase):
         Returns:
             The tensor product of this ansatz and other.
         """
+        def andc(c1,c2):
+            if c1.shape==(1,) and c2.shape==(1,):
+                c3 = math.outer(c1,c2).reshape(-1)
+            elif c1.shape==(1,):
+                c3 = math.outer(c1,c2).reshape(c2.shape)
+            elif c2.shape==(1,):
+                c3 = math.outer(c1,c2).reshape(c1.shape)
+            else:
+                c3 = math.outer(c1,c2).reshape(c1.shape+c2.shape)
+            return c3
+
         As = [math.block_diag(a1, a2) for a1 in self.A for a2 in other.A]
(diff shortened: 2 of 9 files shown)


mrmustard/math/lattice/strategies/vanilla.py
```diff
@@ -72,7 +73,7 @@ def vanilla(shape: tuple[int, ...], A, b, c) -> ComplexTensor:  # pragma: no cov
             value_at_index += A[i, j] * np.sqrt(index_u[j]) * ret[n]
         ret[index] = value_at_index / np.sqrt(index_u[i])
 
-    return ret.reshape(shape)
+    return ret.reshape(shape) * c
```
ziofil, line 76: this leads to the same result, but this way we perform a lot more multiplications no? (one per entry of `ret`)

mrmustard/physics/ansatze.py
```diff
@@ -393,9 +393,10 @@ def __and__(self, other: PolyExpAnsatz) -> PolyExpAnsatz:
         Returns:
             The tensor product of this ansatz and other.
         """
+
         As = [math.block_diag(a1, a2) for a1 in self.A for a2 in other.A]
         bs = [math.concat([b1, b2], axis=-1) for b1 in self.b for b2 in other.b]
-        cs = [math.outer(c1, c2) for c1 in self.c for c2 in other.c]
+        cs = [math.outer(c1, c2).reshape(-1) for c1 in self.c for c2 in other.c]
```
ziofil, line 399: If `c1` and `c2` can potentially be large arrays we shouldn't actually call `outer`, but somehow store them separately and use them as individual units (e.g. as part of a larger tensor contraction). We can leave it for now, but keep in mind that you will probably have to come back to this at some point.

ziofil, line 399: Let's make a mental note to revisit all the times we call `outer`: if the arrays can be large there should be a mechanism to store them separately because their product may not fit in memory.

mrmustard/physics/gaussian_integrals.py
```diff
@@ -149,9 +149,27 @@ def complex_gaussian_integral(
     return A_post, b_post, c_post
 
 
+def join_cs(
```
ziofil, line 152: Ok, perhaps instead of digging ourselves deeper into calling `outer` as a primitive operation, let's revisit why we need to call it. I think it's because we defined the gaussian integral as a function of a single tensor. If we had a version of `gaussian_integral` that takes two tensors then we wouldn't need to join them and call `outer` in the first place. What do you think?

mrmustard/math/lattice/strategies/vanilla.py
kaspernielsen96: I agree. But when c is no longer a scalar, but a polynomial we need to do an inner product between c and some of the dimensions of ret (what is returned from hermite renorm). This essentially updates c. Do you think one can incorporate that inner product in hermite renorm?

mrmustard/physics/gaussian_integrals.py
kaspernielsen96: I agree. It is what we discussed, that it may be more efficient (at least in memory) to not call outer, but just store the two arrays that we would do the outer products on. 
The join_cs is just taking care of the three cases where both c are scalars, both are arrays, and one scalar one array. The question is if we need a separate `gaussian_integral `for each of these cases or we can make a solution that fits all, while not slowing down the "simple scalar" case.

mrmustard/math/lattice/strategies/vanilla.py
ziofil: Oh I see. But still, is it going work the way you wrote it? Because the `*` operator needs `ret` and `c` to have compatible shapes or one of them to be a scalar. is that the case?

mrmustard/physics/gaussian_integrals.py
ziofil: They are surely related, but not the same thing. This is much simpler. At the moment `join_cs` is just a step inside the inner product. But we don't need to take this outer product, and instead carry out the computation of the final result from the two parts. This requires modifying the way `gaussian_integral` takes arguments and works. It shouldn't be difficult, I'm happy to help.

mrmustard/math/lattice/strategies/vanilla.py
kaspernielsen96: No. Right now we have to call hermite renorm with `c=1`, and then do the inner product afterwards. `ret` and `c` will typically have different dimensions and the inner product is performed between some of these dimensions. I will change it back to the way it was when c is just a scalar to reduce the number of multiplications. We can always look into maybe doing something smarter when c is a polynomial.

mrmustard/physics/ansatze.py
```diff
@@ -141,7 +141,7 @@ class PolyExpBase(Ansatz):
     def __init__(self, mat: Batch[Matrix], vec: Batch[Vector], array: Batch[Tensor]):
         self.mat = math.atleast_3d(math.astensor(mat))
         self.vec = math.atleast_2d(math.astensor(vec))
-        self.array = math.atleast_1d(math.astensor(array))
+        self.array = math.atleast_2d(math.astensor(array))
```
apchytr, line 144: Realizing now that `math.astensor` is unnecessary here (as well as in 142 & 143) as its called within `math.atleast_2d`

kaspernielsen96: That shaved off about 3 secs when running `make test`. These small unnecessary calls adds up.

mrmustard/physics/converters.py
```diff
@@ -60,7 +60,10 @@ def to_fock(rep: Representation, shape: Optional[Union[int, Iterable[int]]] = No
             msg += f"the number of variables of this ansatz ({rep.ansatz.num_vars})."
             raise ValueError(msg)
 
-        array = [math.hermite_renormalized(A, b, c, shape) for A, b, c in zip(rep.A, rep.b, rep.c)]
+        array = [
+            math.hermite_renormalized(A, b, complex(c), shape)
```
apchytr, line 64: `atleast_2D` has a `dtype` option. Could we just specify `dtype=complex` and avoid having to cast here?

kaspernielsen96: Making c at least 2d means c here is now an array, but `hermite_renomalized` only takes scalars for c. I think `c[0]` should be used instead. We will however also need to be careful once the shape of c is no longer `(batch,1)`, since one will have to do an inner product between c and the array in line 63)

mrmustard/physics/gaussian_integrals.py
```diff
@@ -149,9 +149,27 @@ def complex_gaussian_integral(
...
+    c1: ComplexTensor,
+    c2: ComplexTensor,
+):
+    r"""Joins two multivariate polynomial arrays into a single multivariate polynomial array."""
+    c1 = math.astensor(c1)
+    c2 = math.astensor(c2)
```
apchytr, line 158: I think you should be able to avoid these calls to `astensor` (avoids additional copies of the tensors)

mrmustard/physics/converters.py
apchytr: Right makes sense. Looks like `hermite_renormalized` will just need to be refactored to account for `!=(batch,1)` cs

mrmustard/physics/gaussian_integrals.py
apchytr: Easiest is probably the `astensor` call but it seems to me the "proper" solution would be to do the rewrite? This would avoid these calls, be consistent with the type hint and consistent with the bargmann

apchytr: Side question: do we expect users to be directly calling `physics.triplets`?

kaspernielsen96: I don't think so, but that is a question for Filippo.

mrmustard/physics/ansatze.py
```diff
@@ -395,9 +395,10 @@ def __and__(self, other: PolyExpAnsatz) -> PolyExpAnsatz:
         Returns:
             The tensor product of this ansatz and other.
         """
+
         As = [math.block_diag(a1, a2) for a1 in self.A for a2 in other.A]
         bs = [math.concat([b1, b2], axis=-1) for b1 in self.b for b2 in other.b]
-        cs = [math.outer(c1, c2) for c1 in self.c for c2 in other.c]
+        cs = [math.outer(c1, c2).reshape(-1) for c1 in self.c for c2 in other.c]
```
apchytr, line 401:
```suggestion
        cs = [math.reshape(math.outer(c1, c2), -1) for c1 in self.c for c2 in other.c]
```

mrmustard/physics/gaussian_integrals.py
```diff
@@ -149,9 +149,27 @@ def complex_gaussian_integral(
...
+    if c1.shape == (1,) and c2.shape == (1,):
+        c12 = math.outer(c1, c2).reshape(-1)
```
apchytr, line 160: These `.reshape` calls should be `math.reshape` (for tensorflow backends)

mrmustard/physics/gaussian_integrals.py
ziofil: surely some users will do that, yes

PR #429, opened 2024-07-02
Minor cleanup to some docstrings + miscellaneous 

apchytr:
**Context:** A small cleanup PR.

- Minor documentation updates/ formatting for consistency. 
- Removed `__getattr__` on `CCView` as it can cause undocumented behaviour.
- Added a `block` option on `draw` such that blocking plots do not display when running the testing suite. 
- Some unnecessary calls to `math.astensor` 
- Moved the batched array logic from `Fock` to `ArrayAnsatz`

### DIFF
--- a/mrmustard/lab_dev/circuit_components.py
+++ b/mrmustard/lab_dev/circuit_components.py
@@ -186,7 +186,8 @@ class CircuitComponent:
 
     @property
     def bargmann(self) -> tuple:
-        r"""The Bargmann parametrization of this component, if available.
+        r"""
+        The Bargmann parametrization of this component, if available.
         It returns a triple (A, b, c) such that the Bargmann function of this component is
         :math:`F(z) = c \exp\left(\frac{1}{2} z^T A z + b^T z\right)`
 
@@ -216,7 +217,8 @@ class CircuitComponent:
         phi: float = 0.0,
         name: Optional[str] = None,
     ) -> CircuitComponent:
-        r"""Returns a circuit component from the given triple (A,b,c) that parametrizes the
+        r"""
+        Returns a circuit component from the given triple (A,b,c) that parametrizes the
         quadrature wavefunction of this component in the form :math:`c * exp(1/2 x^T A x + b^T x)`.
 
         Args:
@@ -578,7 +580,8 @@ class CircuitComponent:
 
 
 class CCView(CircuitComponent):
-    r"""A base class for views of circuit components. It allows for a more efficient
+    r"""
+    A base class for views of circuit components. It allows for a more efficient
     use of components when we need the same component on different wires.
 
     Args:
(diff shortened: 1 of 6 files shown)


mrmustard/lab_dev/circuit_components.py
```diff
@@ -589,10 +592,6 @@ def __init__(self, component: CircuitComponent) -> None:
         self.__dict__ = component.__dict__.copy()
         self._component = component._light_copy()
 
-    def __getattr__(self, name):
```
apchytr, line 592: @ziofil Can you confirm this can be removed? It doesn't seem to break any tests but just want to make sure

ziofil: Yes, it should be fine to remove because it's not doing what it should ^^'
The need was to avoid creating copies of large arrays (which may exist when the representation is `Fock`), so we had created a view that lazily conjugates the representation, and the `__getattr__` was supposed to make it behave like a regular CircuitComponent. But we didn't implement it correctly: it shouldn't look for the method in `_component` (otherwise you get the bug you discovered that `.inverse()` keeps getting called on the original). It should look for the method in `super()` and super should be the same class where `.dual` or `.adjoint` was called. What do you think?

mrmustard/math/tensor_networks/networks.py
```diff
@@ -165,5 +174,5 @@ def draw(tensors: list[Tensor], layout: str = "spring_layout", figsize: tuple[in
     )
 
     plt.title("Mr Mustard Network")
-    plt.show()
+    plt.show(block=block)
```
ziofil, line 177: well spotted

mrmustard/physics/ansatze.py
```diff
@@ -335,7 +334,8 @@ def __call__(self, z: Batch[Vector]) -> Scalar:
         return val
 
     def __mul__(self, other: Union[Scalar, PolyExpAnsatz]) -> PolyExpAnsatz:
-        r"""Multiplies this ansatz by a scalar or another ansatz or a plain scalar.
+        r"""
+        Multiplies this ansatz by a scalar or another ansatz or a plain scalar.
```
ziofil, line 338:
```suggestion
        Multiplies this ansatz by a scalar or another ansatz.
```

mrmustard/physics/ansatze.py
```diff
@@ -358,7 +358,8 @@ def __mul__(self, other: Union[Scalar, PolyExpAnsatz]) -> PolyExpAnsatz:
                 raise TypeError(f"Cannot multiply {self.__class__} and {other.__class__}.") from e
 
     def __truediv__(self, other: Union[Scalar, PolyExpAnsatz]) -> PolyExpAnsatz:
-        r"""Divides this ansatz by a scalar or another ansatz or a plain scalar.
+        r"""
+        Divides this ansatz by a scalar or another ansatz or a plain scalar.
```
ziofil, line 362:
```suggestion
        Divides this ansatz by a scalar or another ansatz.
```

ziofil (commented):
Great! Let me know if you prefer to fix the remaining issue with views in a separate PR and I'll approve this one.

ziofil (approved):
👍

mrmustard/lab_dev/circuit_components.py
apchytr: Discussed offline -  this is to be investigated and done in a follow up PR

PR #430, opened 2024-07-09
Fixes to triples.py

apchytr:
**Context:** Some calls in triples.py were relying on a call to `tf.experimental.numpy.experimental_enable_numpy_behavior()` when using a tensorflow backend. This would only be triggered when calling `math.kron`. As a result, the full test suite would pass but individual tests would fail.

**Description of the Change:** Changed some `.reshape` calls to `math.reshape`.


ziofil (approved):
thanks!

PR #431, opened 2024-07-09
Alternative solution to representation caching 

apchytr:
Opening for discussion 

**Context:** This PR is meant as a proposed alternative to representation caching. By using 'lazy evaluation' in representation/ansatz we can avoid having potentially costly computations just to initialize the representation for calls such as `obj.representation.__class__`. 

**Description of the Change:** Ansatz no longer calls backend methods in the init. Instead, the original parameters are stored (`mat`, `vec`, `array` for `PolyExpAnsatz` or  `array` for `ArrayAnsatz`) and backend methods such as `astensor` and `at_least3d` are only called when the property itself is called (then subsequently stored to avoid repeat calls). In addition to this the representation/ansatz can now also accept a function and some kwargs in substitution of specifying an `(A, b, c)`. 

**Benefits:** Based on some initial profiling initializing a representation/ansatz is more efficient.

### DIFF
--- a/mrmustard/physics/ansatze.py
+++ b/mrmustard/physics/ansatze.py
@@ -142,6 +142,11 @@ class PolyExpBase(Ansatz):
         self._mat = mat
         self._vec = vec
         self._array = array
+
+        self._backend_mat = None
+        self._backend_vec = None
+        self._backend_array = None
+
         self._simplified = False
 
     @property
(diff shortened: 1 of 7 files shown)


mrmustard/physics/ansatze.py
```diff
@@ -284,12 +372,15 @@ def __init__(
...
+        fn: Optional[Callable] = None,
+        **kwargs: Any,
     ):
         self.name = name
 
-        if A is None and b is None:
-            raise ValueError("Please provide either A or b.")
-        super().__init__(mat=A, vec=b, array=c)
+        if A is None and b is None and fn is None:
+            print(fn)
```
apchytr, line 381:
```suggestion
```

ziofil (commented):
Beautiful concept, I really like it! Before we lock it in could we keep A,b,c in the init of `Bargmann`, and have a separate class method that takes `fn, **kwargs` instead? (it's a bit odd to have `A,b,c` *and* (e.g.) `cov, means` as arguments of the init)

apchytr:
> Beautiful concept, I really like it! Before we lock it in could we keep A,b,c in the init of `Bargmann`, and have a separate class method that takes `fn, **kwargs` instead? (it's a bit odd to have `A,b,c` _and_ (e.g.) `cov, means` as arguments of the init)

PR #432, opened 2024-07-10
New poly*exp ansatz

kaspernielsen96:
**Context:**
Implementing the new poly*exp ansatz.
For now just writing and testing the ansatz, but **_NOT_** yet changing the Bargmann representation or anything to use the ansatz.
**Benefits:**
Many once implemented


mrmustard/physics/ansatze.py
```diff
@@ -175,6 +181,18 @@ def degree(self) -> int:
             return 0
         return self.array.shape[-1] - 1
 
+    @property
+    def polynomial_degrees(self) -> tuple[int, tuple]:
```
ziofil, line 185: Why here and not in the `DiffOpPolyExpAnsatz` class?

mrmustard/physics/ansatze.py
```diff
@@ -241,6 +259,76 @@ def _order_batch(self):
...
+                            ],
+                            [
+                                self.mat[i, dim_alpha:, :dim_alpha],
+                                self.mat[i, dim_alpha:, dim_alpha:],
+                            ],
+                        ]
+                    )
+                    for i in range(batch_size)
+                ]
+            )
```
ziofil, line 290:
```suggestion
            A_bar = math.block(
                        [
                            [
                                math.zeros((batch_size, dim_alpha, dim_alpha), dtype=A.dtype),
                                self.mat[..., dim_alpha:, :dim_alpha].T,
                            ],
                            [
                                self.mat[..., dim_alpha:, :dim_alpha],
                                self.mat[..., dim_alpha:, dim_alpha:],
                            ],
                        ]
                    )
```

mrmustard/physics/ansatze.py
```diff
@@ -241,6 +259,76 @@ def _order_batch(self):
...
+
+            b_bar = np.array(
+                [
+                    np.concatenate((np.zeros(dim_alpha), self.vec[i, dim_alpha:]))
+                    for i in range(batch_size)
+                ]
+            )
+
```
ziofil, line 298:
```suggestion
            b_bar = math.block(
                [
                    math.zeros((dim_alpha, dim_beta), dtype=b.dtype), self.vec[..., dim_alpha:]
                ]
            )
```

mrmustard/physics/ansatze.py
```diff
@@ -241,6 +259,76 @@ def _order_batch(self):
...
+            poly_bar = math.hermite_renormalized_batch(
```
ziofil, line 299: Given that this is the only place in MM where this function is used, it may be worth modifying its definition to work with the batch at the beginning so that we avoid the moveaxis calls back and forth. The function is called `vanilla_batch`, it's super easy to modify. Also, we should use `math` rather than `np` whenever the function should be differentiable with respect to at least one argument (e.g. `np.sum`, `np.block`, etc... break the chain rule because they should be differentiable. Use `math.sum` `math.block` etc. (We don't support every possible numpy function and sometimes the signatures differ)

mrmustard/physics/ansatze.py
ziofil: there are more places below where you can do it in one step

mrmustard/physics/ansatze.py
kaspernielsen96: This one is the new `degree` method that is defined just above in the code. 
The difference is that the degree function assumes a single variable polynomial, whereas we can have different degrees of polynomials on different wires. I just added it to the `polyexpbase` because the other one is also there. We might want to make `degree` just return the first argument of `polynomial_degrees` once the new ansatz is up and running.

mrmustard/physics/ansatze.py
kaspernielsen96: Do you think we can rewrite it even further so that it can work with two different batch dimensions, one for the argument and one for the Abc? Then we can avoid the loop around it.

tests/test_physics/test_ansatz.py
```diff
@@ -355,3 +356,216 @@ def test_bargmann_Abc_to_phasespace_cov_means(self):
...
+
+    @pytest.mark.parametrize("triple", [Abc_n1, Abc_n2, Abc_n3])
+    def test_init(self, triple):
+        A, b, c = triple
+        ansatz = DiffOpPolyExpAnsatz(A, b, c)
+        assert np.allclose(ansatz.mat[0], A)
+        assert np.allclose(ansatz.vec[0], b)
+        assert np.allclose(ansatz.array[0], c)
+
+    def test_add(self):
```
elib20, line 378: Adding them together didn't work if the c's have different lengths. This returned an error.

```
A1 = np.ones((1,2,2))
b1 = np.ones((1,2))
c1 = np.array([[1,2,3,4,5,6]])

A2 = np.ones((1,2,2))
b2 = np.ones((1,2))
c2 = np.array([[1,2,3,4,5]])

ansatze1 = DiffOpPolyExpAnsatz(A1,b1,c1)
ansatze2 = DiffOpPolyExpAnsatz(A2,b2,c2)

ansatze1 + ansatze2
```

mrmustard/physics/ansatze.py
```diff
@@ -401,6 +497,372 @@ def __and__(self, other: PolyExpAnsatz) -> PolyExpAnsatz:
...
+        return self.vec
+
+    @property
+    def c(self) -> Batch[ComplexTensor]:
+        r"""
+        The array of coefficients for the polynomial terms in the ansatz.
+        """
+        return self.array
+
+    def __call__(self, z: Batch[Vector]) -> Scalar:
```
elib20, line 571: I can't really figure out what I'm doing wrong here. I thought I was initializing a Gaussian with two variables, and via c I thought I was telling it to take the 0th derivative on the last variable, then set it to zero (which should just be setting the second variable to zero). In which case, I would expect to just get exp(1/2) if I evaluate the ansatze at 1. But I'm getting that the ansatze evaluates to exp(1).

I have hbar=1 too.
```
A1 = np.eye(2)
b1 = np.zeros((2))
c1 = np.array([1])

ansatze = DiffOpPolyExpAnsatz(A1,b1,c1)
assert ansatze(np.array([1])) == np.exp(1)
assert ansatze(np.array([1])) == np.exp(0.5)
```

mrmustard/physics/ansatze.py
```diff
@@ -401,6 +497,372 @@ def __and__(self, other: PolyExpAnsatz) -> PolyExpAnsatz:
...
+
+    .. code-block::
+
+        >>> from mrmustard.physics.ansatze import DiffOpPolyExpAnsatz
+
+        >>> A = np.array([[1.0, 0.0], [0.0, 1.0]])
+        >>> b = np.array([1.0, 1.0])
+        >>> c = np.array([1.0,2.0,3.0])
+
+        >>> F = PolyExpAnsatz(A, b, c)
```
elib20, line 519: This is a different class

elib20: If I try this
```
A = np.array([[1.0, 0.0], [0.0, 1.0]])
b = np.array([1.0, 1.0])
c = np.array([1.0,2.0,3.0])

ansatze = DiffOpPolyExpAnsatz(A,b,c)
```
I trigger the error that the batch sizes need to be the same.

mrmustard/physics/ansatze.py
```diff
@@ -401,6 +497,375 @@ def __and__(self, other: PolyExpAnsatz) -> PolyExpAnsatz:
...
+class DiffOpPolyExpAnsatz(PolyExpBase):
+    r"""
+    The ansatz of the Fock-Bargmann representation.
+
+    Represents the ansatz function:
+
+        :math:`F(z) = \sum_i [\sum_k c^(i)_k \partial_y^k \textrm{exp}((z,y)^T A_i (z,y) / 2 + (z,y)^T b_i)|_{y=0}]`
+
+    with ``k`` being a multi-index. The matrices :math:`A_i` and vectors :math:`b_i` are
+    parameters of the exponential terms in the ansatz, and :math:`z` is a vector of variables.
```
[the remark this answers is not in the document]
elib20: please clarify where the batch dimension factors in here, and clarify the rules around the dimension of c, what variables the derivatives are assumed to act on, etc.

elib20 (commented):
I started trying to review, but I can't figure out how the initialization works, so I am blocked from further review until I understand that enough to be able to run small sanity checks.

You accompanying note is very comprehensive for all the methods, but it doesn't go into much detail about how the __init__ method works.

In particular, I cannot figure out what dimensions c should be taking, what the rules are around the batch dimension.

mrmustard/physics/ansatze.py
elib20: I got it now. I needed to add a batch dimension to both c and the argument for the call function

mrmustard/physics/ansatze.py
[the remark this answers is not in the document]
kaspernielsen96: A working example of your code is
`A = np.eye(2)
b = np.ones(2)
c = np.array([[[1,2,3,4,5,6],[1,2,3,4,5,0]]])

ansatze = DiffOpPolyExpAnsatz(A,b,c)
ansatze.polynomial_degrees`

But I think you meant the following, where an outer product has to be taken
`A = np.eye(2)
b = np.ones(2)
c = np.outer(np.array(([1,2,3,4,5,6])),np.array(([1,2,3,4,5,0]))).reshape(1,6,6)
ansatze = DiffOpPolyExpAnsatz(A,b,c)
ansatze.polynomial_degrees`

However, right now we don't remove trailing zeros and the degree returns one too many (starts from degree 1 rather than 0)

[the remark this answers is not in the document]
kaspernielsen96: Maybe it should be renamed to polynomial shape, since that is what we actually need for some other functions.

tests/test_physics/test_ansatz.py
kaspernielsen96: But would we expect that to work? One can also not add two ansatze together if the A and b are different.

mrmustard/physics/ansatze.py
kaspernielsen96: Thanks. This is updated in the new docs.

tests/test_physics/test_ansatz.py
kaspernielsen96: But would we expect this to work? We can also not add ansatze that have different size of A and b.

mrmustard/physics/ansatze.py
kaspernielsen96: I updated the call function so you don't need to add a batch to the argument yourself. Also note that the batch on the argument can be different than for the triple. This makes it work equivalent to the old ansatz

mrmustard/math/lattice/steps.py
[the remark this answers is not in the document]
kaspernielsen96: This will be a separate PR.

mrmustard/physics/ansatze.py
```diff
@@ -241,6 +259,84 @@ def _order_batch(self):
         self.vec = math.gather(self.vec, sorted_indices, axis=0)
         self.array = math.gather(self.array, sorted_indices, axis=0)
 
+    def decompose_ansatz(self) -> DiffOpPolyExpAnsatz:
+        r"""
+        This method decomposes a DiffOpPolyExpAnsatz. Given an ansatz of dimensions:
+        A=(batch,m+n,m+n), b=(batch,m+n), c = (batch,k_1,k_2,...,k_n),
+        it can be rewritten as an ansatz of dimensions
+        A=(batch,2m,2m), b=(batch,2m), c = (batch,l_1,l_2,...,l_m), with l_i = sum_j k_j
+        This decomposition is typically favourable if n>m, and will only run if that is the case.
```
elib20, line 268: It might be helpful of understanding the code in the future to clarify what are shape_alpha, dim_alpha, shape_beta, dim_beta in this explanation.

tests/test_physics/test_ansatz.py
```diff
@@ -355,3 +356,223 @@ def test_bargmann_Abc_to_phasespace_cov_means(self):
...
+    def test_decompose_ansatz(self):
+        A, b, _ = Abc_triple(4)
+        c = np.random.uniform(-10, 10, size=(1, 3, 3, 3))
+        ansatz = DiffOpPolyExpAnsatz(A, b, c)
+        decomp_ansatz = ansatz.decompose_ansatz()
+        z = np.random.uniform(-10, 10, size=(1, 1))
+        assert np.allclose(ansatz(z), decomp_ansatz(z))
+        assert np.allclose(decomp_ansatz.A.shape, (1, 2, 2))
+
+    def test_decompose_ansatz_batch(self):
```
elib20, line 555: I think you are testing both the batching of the A,b,c and the batching of z here right? Maybe just add a docstring to the test to explain that.

elib20 (approved):
Awesome work Kasper! 

All my questions/comments are essentially around code documentation since this is a cool, but non-trivial new ansatze, so I want to make sure it's easy to follow what is going on in the future. I also think more of the tests should include cases where c is a vector.

I don't think I found any major bugs, so from my side, I'll provide a conditional approval assuming the documentation is improved a bit, and the tests check a few more cases.

mrmustard/physics/ansatze.py
```diff
@@ -139,9 +145,9 @@ class PolyExpBase(Ansatz):
     """
 
     def __init__(self, mat: Batch[Matrix], vec: Batch[Vector], array: Batch[Tensor]):
-        self.mat = math.atleast_3d(math.astensor(mat))
-        self.vec = math.atleast_2d(math.astensor(vec))
-        self.array = math.atleast_1d(math.astensor(array))
+        self.mat = math.atleast_3d(mat)
+        self.vec = math.atleast_2d(vec)
+        self.array = math.atleast_1d(array)
```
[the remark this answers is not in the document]
kaspernielsen96: No, that is what I thought as well. But after discussions with Filippo we agreed that this is actually the best and also simplest way. No matter what we will always have to give an explicit batch dimension for c, since the polynomials can have any degree.

mrmustard/physics/ansatze.py
```diff
@@ -401,6 +497,366 @@ def __and__(self, other: PolyExpAnsatz) -> PolyExpAnsatz:
...
+
+        exp_sum = math.exp(1 / 2 * A_part + b_part)
+        if dim_beta == 0:
+            val = math.sum(exp_sum * self.c, axes=[-1])
+        else:
+            b_poly = math.astensor(
+                [
+                    math.sum(self.A[..., dim_alpha:, :dim_alpha] * z[i, None, :], axes=[-1])
+                    + self.b[..., dim_alpha:]
+                    for i in range(batch_size_arg)
```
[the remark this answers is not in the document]
kaspernielsen96: You are right, I have updated it. The loop around `hermite_renorm_batch` is however difficult to get around.

tests/test_physics/test_ansatz.py
```diff
@@ -355,3 +356,223 @@ def test_bargmann_Abc_to_phasespace_cov_means(self):
...
+        d = 0.1
+
+        ansatz = DiffOpPolyExpAnsatz(A, b, c)
+        ansatz2 = ansatz / d
+
+        assert np.allclose(ansatz2.mat[0], A)
+        assert np.allclose(ansatz2.vec[0], b)
+        assert np.allclose(ansatz2.array[0], c / d)
+
+    def test_call(self):
```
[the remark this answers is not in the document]
kaspernielsen96: What do you mean with analytically calculated one? When c is a vector the only way to calculate the value is by going through the call functions which also includes a call to `hermite_renorm`. I can create all that logic in the test, but it will look very similar to the `call` function.

kaspernielsen96:
Regarding `and` and different shapes of c. I add some (ugly) code that can handle that and will add padding zeros (any improvements are more than welcome).
Note that the number of not-wires with polynomials must be the same. Otherwise things will start to break down very quickly since this ties into also A and b.

tests/test_physics/test_ansatz.py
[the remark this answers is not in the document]
elib20: E.g. for the A,b,c corresponding to a two-mode squeezed state, we know the poly*exp ansatz for a c with all zeros and one 1 in it should should just be a Fock state, where the analytic expression is known.

mrmustard/physics/ansatze.py
kaspernielsen96: What we need for later is the number of (not) wires with polynomials on them and sometimes the degree of the polynomial on each wire. This is different from the sum of them, unless we do the decomposition - in that case the new polynomials will have the degree which is the sum of the degrees of the old ones. I also renamed this to polynomial shape.

mrmustard/math/backend_manager.py
```diff
@@ -800,6 +800,27 @@ def minimum(self, a: Tensor, b: Tensor) -> Tensor:
             ),
         )
 
+    def moveaxis(self, a: Tensor, old: Tensor, new: Tensor) -> Tensor:
```
apchytr, line 803: Can we rename `a` to something more descriptive like `array`

mrmustard/math/backend_manager.py
```diff
@@ -800,6 +800,27 @@ def minimum(self, a: Tensor, b: Tensor) -> Tensor:
...
+        r"""Move axes of an array to a new position.
+
+        Args:
+            a: The array to move axes off
```
apchytr, line 807:
```suggestion
            a: The array to move the axis of.
```

mrmustard/math/backend_manager.py
```diff
@@ -800,6 +800,27 @@ def minimum(self, a: Tensor, b: Tensor) -> Tensor:
             ),
         )
 
+    def moveaxis(self, a: Tensor, old: Tensor, new: Tensor) -> Tensor:
+        r"""Move axes of an array to a new position.
```
apchytr, line 804:
```suggestion
        r"""
        Moves the axes of an array to a new position.
```

mrmustard/math/backend_tensorflow.py
```diff
@@ -245,6 +245,9 @@ def maximum(self, a: tf.Tensor, b: tf.Tensor) -> tf.Tensor:
     def minimum(self, a: tf.Tensor, b: tf.Tensor) -> tf.Tensor:
         return tf.minimum(a, b)
 
+    def moveaxis(self, a: tf.Tensor, old: tf.Tensor, new: tf.Tensor) -> tf.Tensor:
```
apchytr, line 248: I believe the type hint for old and new would still be `Union[int, Sequence[int]]` or `int | Sequence[int]`

apchytr: In fact all of them could just be `a: Tensor, old: int | Sequence[int], new: int | Sequence[int]`

mrmustard/math/lattice/steps.py
```diff
@@ -80,25 +80,24 @@ def vanilla_step_batch(
     The necessary pivot and neighbours must have already been computed,
     as this step will read those values from G.
     Note that this function is different from vanilla_step with b is no longer a vector,
-    it becomes a bathced vector with the batch dimension on the last index.
+    it becomes a bathced vector with the batch dimension on the first index.
 
     Args:
         G (array or dict): fock amplitudes data store that supports getitem[tuple[int, ...]]
```
apchytr, line 86: Unrelated to this specific PR but types arent really necessary in the docstring as long as the args are typehinted so these should probably be removed eg `G (array or dict)` -> `G`

mrmustard/physics/ansatze.py
```diff
@@ -218,6 +251,18 @@ def degree(self) -> int:
...
         return self.array.shape[-1] - 1
 
+    @property
+    def polynomial_shape(self) -> tuple[int, tuple]:
+        r"""
+        This method finds the dimensionality of the polynomial, i.e. how many wires
+        have polynomials attached to them and what the degree of the polynomial is
+        on each of the wires.
+        """
+
```
apchytr, line 261:
```suggestion
```

mrmustard/physics/ansatze.py
```diff
@@ -515,6 +639,387 @@ def __and__(self, other: PolyExpAnsatz) -> PolyExpAnsatz:
...
+    @property
+    def c(self) -> Batch[ComplexTensor]:
+        r"""
+        The array of coefficients for the polynomial terms in the ansatz.
+        """
+        return self.array
+
+    @classmethod
+    def from_function(cls, fn: Callable, **kwargs: Any) -> DiffOpPolyExpAnsatz:
```
apchytr, line 715: I believe if you inherit from PolyExpAnsatz instead of PolyExpBase then you can avoid having to copy paste `A`, `b`, `c` and `from_function` which should clean things up

mrmustard/math/lattice/steps.py
kaspernielsen96: This seems to be a general thing for all functions in this file. I am not sure this is the PR to correct these things, but I agree with you.

mrmustard/physics/ansatze.py
kaspernielsen96: I am pretty sure that I inherent from PolyExpBase, but I can be wrong.

mrmustard/math/lattice/steps.py
apchytr: Yeah we probably just need a cleanup PR for these things

mrmustard/physics/ansatze.py
apchytr: Hmm even so I think it still makes sense to have `DiffOpPolyExpAnsatz` inherit from `PolyExpAnsatz` in this PR to avoid this (albeit temporary) copy pasting. When the time comes to do the replacement it should still be relatively easy I believe. Lmk what you think

kaspernielsen96: I agree from a code perspective, but the math of the two ansatze are quite different, if `PolyExpAnsatz` had supported c as an array and not just a scalar. The docstrings of the two show the differences between the two when c is not just a scalar. To me it would from a math/physics perspective be weird if one inherited from the other since they are in principle quite different once c is no longer a scalar.

apchytr: As discussed offline we agreed to keep as is

mrmustard/math/backend_manager.py
```diff
@@ -800,11 +800,12 @@ def minimum(self, a: Tensor, b: Tensor) -> Tensor:
...
 
-    def moveaxis(self, a: Tensor, old: Tensor, new: Tensor) -> Tensor:
-        r"""Move axes of an array to a new position.
+    def moveaxis(self, array: Tensor, old: Tensor, new: Tensor) -> Tensor:
+        r"""
+        Moves the axes of an array to a new position.
 
         Args:
-            a: The array to move axes off
+            a: The array to move the axes of.
```
elib20, line 808:
```suggestion
            array: The array to move the axes of.
```

tests/test_physics/test_ansatz.py
```diff
@@ -451,17 +476,57 @@ def test_call(self):
         ):
             ansatz(z)
 
+        A = np.array([[0, 1], [1, 0]])
+        b = np.zeros(2)
+        c = c = np.zeros(10, dtype=complex).reshape(1, -1)
```
elib20, line 481:
```suggestion
        c = np.zeros(10, dtype=complex).reshape(1, -1)
```

elib20 (approved):
Nice work!

elib20 (approved):
reapproved!

PR #434, opened 2024-07-11
New gaussian integral

kaspernielsen96:
**Context:**
A gaussian integral that takes two Abc triplets.
Adding an extra method for `complex_gaussian_integral`

### DIFF
--- a/mrmustard/physics/gaussian_integrals.py
+++ b/mrmustard/physics/gaussian_integrals.py
@@ -19,7 +19,7 @@ This module contains gaussian integral functions and related helper functions.
 from typing import Sequence, Tuple
 import numpy as np
 from mrmustard import math
-from mrmustard.utils.typing import ComplexMatrix, ComplexVector
+from mrmustard.utils.typing import ComplexMatrix, ComplexVector, ComplexTensor
 
 
 def real_gaussian_integral(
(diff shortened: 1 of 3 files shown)


mrmustard/physics/gaussian_integrals.py
```diff
@@ -286,3 +287,83 @@ def contract_two_Abc(
     return complex_gaussian_integral(
         Abc, idx1, tuple(n + Abc1[0].shape[-1] for n in idx2), measure=-1.0
     )
+
+
+def complex_gaussian_integral_2(
```
apchytr, line 292: On first glance this method looks very similar to the single `Abc` case. I was wondering if this could be generalized to just handle both cases (maybe even more)? E.g. `complex_gaussian_integral(Abc: list[tuple], idx: tuple[int, ...], idx_conj: tuple[int, ...], measure: float = -1)` where `len(Abc)` determines how many Abc triples you're working with

kaspernielsen96: It is essentially the same for now, but Filippo has ideas to update it later (how the outer products are handled). One could maybe make one function to handle both cases.

ziofil (approved):
👍

PR #436, opened 2024-07-15
Lazy evaluation in Representation/Ansatz

apchytr:
Context: This PR is meant as an alternative to representation caching. By using 'lazy evaluation' in representation/ansatz we can avoid having potentially costly computations just to initialize the representation for calls such as obj.representation.__class__.

Description of the Change: Ansatz no longer calls backend methods in the init. Instead, the original parameters are stored (mat, vec, array for PolyExpAnsatz or array for ArrayAnsatz) and backend methods such as astensor and at_least3d are only called when the property itself is called (then subsequently stored to avoid repeat calls). In addition to this the representation/ansatz can now also accept a function and some kwargs in substitution of specifying an (A, b, c) through a from_function class method. This `from_function` method accepts `Constant`/`Variable` params as the arguments thus avoiding the need to reinitialize a representation with every call of the CC. 

Benefits: Based on some initial profiling initializing a representation/ansatz is more efficient.


ziofil (commented):
very very cool. let's go through it together before I push the approve button

ziofil (approved):
super!

PR #441, opened 2024-07-23
Fix off by 1 error

ziofil:
The auto_shape was off by 1 (e.g. `Coherent([0], 0.0, 0.0)` had a shape of `(0,)` rather than `(1,)`).


mrmustard/math/lattice/strategies/vanilla.py
```diff
@@ -327,5 +327,5 @@ def autoshape_numba(A, b, c, max_prob, max_shape) -> int:  # pragma: no cover
             ) / SQRT[k + 2]
             norm += np.abs(buf3[(k + 1) % 2, 1])
             k += 1
-        shape[m] = k
+        shape[m] = k + 1
```
timmysilv, line 330: is this only wrong when `k` is still 0 (eg. the above loop never occurs)? if so, maybe it should be `shape[m] = k or 1` to avoid the extra `+ 1`

ziofil: genius

ziofil, line 330:
```suggestion
        shape[m] = k or 1
```

timmysilv (approved):

PR #443, opened 2024-07-24
Quadrature methods for circuit components and states

elib20:
**Context:** We would like convenience methods for evaluating the quadrature basis representation of circuit components and states.

- renames the `quadrature` method for circuit components to `quadrature_triple`
- adds a new `to_quadrature` method to convert to the quadrature representation from Bargmann
- adds a new `quadrature` method for circuit components to evaluate the quadrature basis function over an array of points. This works for Bargmann and Fock representations, batched or unbatched.
- adds a new `quadrature_distribution` method for Ket and DM to evaluate the quadrature basis probability distribution over an array of points. This works for Bargmann and Fock representations, batched or unbatched.

**Benefits:** 
- Convenient methods for users
- Quadrature methods now work for Fock representation

**Possible Drawbacks:**
None

**Related GitHub Issues:**


ziofil:
Maybe it's late for this suggestions, but wouldn't it be sufficient to apply `BtoQ` on all wires and return the ansatz? Then the ansatz itself can be called independently.

mrmustard/lab_dev/circuit_components.py
```diff
@@ -317,6 +319,60 @@ def from_quadrature(
...
+        from .circuit_components_utils.b_to_q import BtoQ
+
+        BtoQ_ob = BtoQ(self.wires.output.bra.modes, phi).adjoint
+        BtoQ_ib = BtoQ(self.wires.input.bra.modes, phi).adjoint.dual
+        BtoQ_ok = BtoQ(self.wires.output.ket.modes, phi)
+        BtoQ_ik = BtoQ(self.wires.input.ket.modes, phi).dual
+        QQQQ = BtoQ_ib @ (BtoQ_ik @ self @ BtoQ_ok) @ BtoQ_ob
+        return QQQQ
+
+    def quadrature_triple(self, phi: float = 0.0) -> tuple | ComplexTensor:
```
apchytr, line 345:
```suggestion
    def quadrature_triple(self, phi: float = 0.0) -> tuple[Batch[ComplexMatrix], Batch[ComplexVector], Batch[ComplexTensor]]:
```

mrmustard/lab_dev/circuit_components.py
```diff
@@ -317,6 +319,60 @@ def from_quadrature(
...
+                conjugates[i] if i in self.wires.ket.indices else True
+                for i in range(len(self.wires.indices))
+            ]
+            quad_basis = math.sum(
+                [quadrature_basis(array, quad, conjugates, phi) for array in fock_arrays], axes=[0]
+            )
+            return quad_basis
+
+        QQQQ = self.to_quadrature(phi=phi)
+        return QQQQ.representation.ansatz(quad)
```
apchytr, line 374: Should be able to do 
```suggestion
        return QQQQ.representation(quad)
```

mrmustard/lab_dev/circuit_components.py
```diff
@@ -317,6 +319,60 @@ def from_quadrature(
...
+        """
+
+        if isinstance(self.representation, Fock):
+            fock_arrays = self.representation.array
+            # Find where all the bras and kets are so they can be conjugated appropriately
+            conjugates = [False] * len(self.wires.indices)
+            conjugates = [
+                conjugates[i] if i in self.wires.ket.indices else True
+                for i in range(len(self.wires.indices))
+            ]
```
apchytr, line 367: I think you can simplify 363-367 by doing 
```suggestion
            conjugates = [
                False if i in self.wires.ket.indices else True
                for i in range(len(self.wires.indices))
            ]
```

mrmustard/lab_dev/circuit_components.py
```diff
@@ -317,6 +319,60 @@ def from_quadrature(
...
+        r"""
+        The quadrature representation triple A,b,c of this circuit component.
+        """
+        if isinstance(self.representation, Fock):
+            raise NotImplementedError("Not implemented with Fock representation.")
+
+        QQQQ = self.to_quadrature(phi=phi)
+        return QQQQ.representation.data
+
+    def quadrature(self, quad: Batch[Vector], phi: float = 0.0) -> tuple | ComplexTensor:
```
apchytr, line 355:
```suggestion
    def quadrature(self, quad: Batch[Vector], phi: float = 0.0) -> ComplexTensor:
```
I believe based on the type hint of call

mrmustard/lab_dev/circuit_components.py
```diff
@@ -317,6 +319,60 @@ def from_quadrature(
...
+        r"""
+        The (discretized) quadrature basis representation of the circuit component.
```
apchytr, line 357: Can you document the Args

mrmustard/lab_dev/states/base.py
```diff
@@ -330,6 +331,12 @@ def phase_space(self, s: float) -> tuple:
         new_state = self >> BtoPS(self.modes, s=s)
         return bargmann_Abc_to_phasespace_cov_means(*new_state.bargmann_triple(batched=True))
 
+    def quadrature_distribution(self, quad: Vector, phi: float = 0.0) -> tuple | ComplexTensor:
+        r"""
+        The (discretized) quadrature distribution of the circuit component.
```
apchytr, line 336:
```suggestion
        The (discretized) quadrature distribution of the State.
```
Unless we want to move this to CircuitComponent

apchytr: Also, can Args be included 🙇

mrmustard/lab_dev/states/base.py
```diff
@@ -719,6 +726,21 @@ def _purities(self) -> RealVector:
     def purity(self) -> float:
         return self.L2_norm
 
+    def quadrature_distribution(self, quad: Vector, phi: float = 0.0) -> tuple | ComplexTensor:
+        r"""
+        The (discretized) quadrature distribution of the circuit component.
```
apchytr, line 731: I don't think we need the docstrings for the overwritten methods

mrmustard/physics/fock.py
```diff
@@ -818,10 +818,62 @@ def estimate_quadrature_axis(cutoff, minimum=5, period_resolution=20):
...
+        phi (float): angle of the quadrature basis vector
+
+    Returns:
+        tuple(Tensor): quadrature basis representation at the points in quad
+    """
+    dims = len(fock_array.shape)
+
+    if quad.shape[-1] != dims:
+        raise ValueError(
+            "Input fock array has dimension {fock_array.shape}. Make sure quad has the same dimension."
```
apchytr, line 843:
```suggestion
            f"Input fock array has dimension {fock_array.shape}. Make sure quad has the same dimension."
```

mrmustard/physics/fock.py
```diff
@@ -818,10 +818,62 @@ def estimate_quadrature_axis(cutoff, minimum=5, period_resolution=20):
...
+        )
+
+    if type(conjugates) is bool:
+        conjugates = [conjugates] * dims
```
apchytr, line 847:
```suggestion
        conjugates = conjugates if isinstance(conjugates, Iterable) else [conjugates] * dims
```

mrmustard/lab_dev/circuit_components.py
```diff
@@ -317,6 +319,60 @@ def from_quadrature(
...
+        Returns a circuit component with the quadrature representation of this component
+        in terms of A,b,c.
+
+        Args:
+            phi: The quadrature angle. ``phi=0`` corresponds to the x quadrature,
+                    ``phi=pi/2`` to the p quadrature. The default value is ``0``.
+        Returns:
+            A circuit component with the given quadrature representation.
+        """
+        if isinstance(self.representation, Fock):
```
kaspernielsen96, line 333: We now have the option to convert a Fock object to Bargmann and then use BtoQ, but I am not sure if we want to do that?

apchytr: Maybe we could with a warning?

elib20: Why would we need a warning? I think it's fine without.

mrmustard/lab_dev/circuit_components.py
```diff
@@ -317,6 +319,76 @@ def from_quadrature(
...
+        The quadrature representation triple A,b,c of this circuit component.
+
+        Args:
+            phi: The quadrature angle. ``phi=0`` corresponds to the x quadrature,
+                    ``phi=pi/2`` to the p quadrature. The default value is ``0``.
+        Returns:
+            A,b,c triple of the quadrature representation
+        """
+        if isinstance(self.representation, Fock):
+            raise NotImplementedError("Not implemented with Fock representation.")
```
kaspernielsen96, line 360: Do we also want to convert this to bargmann?

kaspernielsen96 (approved):

mrmustard/lab_dev/circuit_components.py
```diff
@@ -317,6 +319,76 @@ def from_quadrature(
...
+            A circuit component with the given quadrature representation.
+        """
+
+        if isinstance(self.representation, Fock):
+            fock_arrays = self.representation.array
+            # Find where all the bras and kets are so they can be conjugated appropriately
+            conjugates = [
+                False if i in self.wires.ket.indices else True
+                for i in range(len(self.wires.indices))
+            ]
```
ziofil, line 383:
```suggestion
            conjugates = [i not in self.wires.ket.indices for i in self.wires.indices]
```

mrmustard/physics/fock.py
```diff
@@ -818,6 +818,57 @@ def estimate_quadrature_axis(cutoff, minimum=5, period_resolution=20):
...
+
+    if quad.shape[-1] != dims:
+        raise ValueError(
+            f"Input fock array has dimension {dims} whereas ``quad`` has {quad.shape[-1]}."
+        )
+
+    conjugates = conjugates if isinstance(conjugates, Iterable) else [conjugates] * dims
+
+    # construct quadrature basis vectors
+    cutoffs = fock_array.shape
```
ziofil, line 849: I think we need to be careful with how we mean "shape" and "cutoff". Usually cutoff refers to the largest photon number, which is 1 less than the shape.

mrmustard/physics/fock.py
```diff
@@ -818,6 +818,57 @@ def estimate_quadrature_axis(cutoff, minimum=5, period_resolution=20):
...
+            q_to_n = math.einsum("a,ab->ab", Ur, q_to_n)
+        if conjugates[dim]:
+            q_to_n = math.conj(q_to_n)
+        quad_basis_vecs += [math.cast(q_to_n, "complex128")]
+
+    # Convert each dimension to quadrature
+    subscripts = [chr(i) for i in range(98, 98 + dims)]
+    fock_string = "".join(subscripts[:dims])  #'bcd....'
+    q_string = "".join([fock_string[i] + "a," for i in range(dims - 1)] + [fock_string[-1] + "a"])
+    fock_array = math.einsum(
```
ziofil, line 865: maybe it's clearer if we don't call it `fock_array` after conversion

mrmustard/physics/fock.py
```diff
@@ -836,34 +887,17 @@ def quadrature_distribution(
...
-    pdf = (
-        math.einsum("nm,nj,mj->j", state, psi_x, psi_x)
-        if is_dm
-        else math.abs(math.einsum("n,nj->j", state, psi_x)) ** 2
-    )
+    dims = len(state.shape)
+    is_dm = dims == 2
+
+    quad = math.transpose(math.astensor([x] * dims))
+    conjugates = [False, True] if is_dm else [False]
```
ziofil, line 898: I already commented above, but the convention is that bras come before kets so this may be the opposite of what we want

tests/test_lab_dev/test_circuit_components.py
```diff
@@ -445,23 +445,23 @@ def test_fock_component_no_bargmann(self):
     def test_quadrature_ket(self):
         "tests that transforming to quadrature and back gives the same ket"
         ket = SqueezedVacuum([0], 0.4, 0.5) >> Dgate([0], 0.3, 0.2)
-        back = Ket.from_quadrature([0], ket.quadrature())
+        back = Ket.from_quadrature([0], ket.quadrature_triple())
```
ziofil, line 448: these tests are so cool :)

ziofil:
I approved it, but make sure those indices are in the right order

tests/test_lab_dev/test_circuit_components.py
elib20: it's because you wrote them!

mrmustard/lab_dev/circuit_components.py
elib20: We just need a list to say which dimensions of the tensor we need to complex conjugate. If I do:
`
state = SqueezedVacuum([0,1]).dm()
conjugates = [
                False if i in state.wires.ket.indices else True
                for i in range(len(state.wires.indices))
            ]
`
I get back [True,True,False,False], which is correct since we need to take the complex conjugate of the bras that come first.

mrmustard/physics/fock.py
elib20: Good catch!

mrmustard/lab_dev/circuit_components.py
ziofil: oh, right, I got confused by the line starting with `False` 😆

kaspernielsen96 (approved):

PR #444, opened 2024-07-29
Restructuring Mr Mustard states.py, transformations.py and circuit_components_utils.py + pylint misc

apchytr:
**Context:** Rather than keeping all our states in a `states.py` or transformations in a `transformations.py` for example we should separate out each class into an individual `class.py` file. 

**Description of the Change:** Individual classes per file (with some exceptions depending on if it makes sense or not) 

**Benefits:** Easier to see what is and isn't implemented from the file structure. Easier to see unit tests for classes. Ensures we don't run into the 'pylint: too complex' check. 

**Possible Drawbacks:** Trade off is that we will have more `.py` files in our repo.


mrmustard/lab_dev/circuit_components_utils/__init__.py
ziofil, line 1: I think we can put these utility gates in transformations too

ziofil (approved):
Thanks! This was very much needed.

mrmustard/lab_dev/circuit_components_utils/__init__.py
apchytr: Hmm on one hand they do inherit from the transformation base classes (at least `BtoQ` and `BtoPS` do) but on the other hand they behave differently enough from our gates to warrant separating them out. Not sure which would be the best what do you think?

PR #445, opened 2024-07-31
Turned bargmann property into a method and added `batched` argument

ziofil:
This PR turns the `.bargmann` property into a method and adds the argument `batched`, which now can be set to `False` and avoid having to extract the A,b,c triple from a batch of length 1.


mrmustard/lab_dev/circuit_components.py
```diff
@@ -184,27 +184,30 @@ def from_bargmann(
         wires = Wires(set(modes_out_bra), set(modes_in_bra), set(modes_out_ket), set(modes_in_ket))
         return cls._from_attributes(repr, wires, name)
 
-    @property
-    def bargmann(self) -> tuple:
+    def bargmann(self, batched=False) -> tuple:
```
apchytr, line 187:
```suggestion
    def bargmann(self, batched: bool=False) -> tuple[Batch[ComplexMatrix], Batch[ComplexVector], Batch[ComplexTensor]]:
```

mrmustard/lab_dev/circuit_components.py
```diff
@@ -183,27 +186,32 @@ def from_bargmann(
         wires = Wires(set(modes_out_bra), set(modes_in_bra), set(modes_out_ket), set(modes_in_ket))
         return cls._from_attributes(repr, wires, name)
 
-    @property
-    def bargmann(self) -> tuple:
+    def bargmann_triple(
+        self, batched: bool = False
+    ) -> tuple[Batch[ComplexMatrix], Batch[ComplexVector], Batch[ComplexTensor]]:
```
apchytr, line 191: I think `Batch`, `ComplexMatrix`, `ComplexVector` and `ComplexTensor` need to be imported based on codefactor

apchytr (approved):

PR #449, opened 2024-08-01
Implement new polyexp ansatz

kaspernielsen96:
**Context:**
Updating MM to use the new polyexp ansatz
Renaming the new ansatz to `PolyExpAnsatz` and the old one to `PolyExpAnsatzOld`.
Update `mat_mul`(and sub functions), `fock`,`to_fock`,  to work with polynomials


mrmustard/physics/gaussian_integrals.py
```diff
@@ -258,12 +258,18 @@ def reorder_abc(Abc: tuple, order: Sequence[int]):
...
+            if type(order) == list:
+                order.extend(np.arange(len(order), len(order) + dim_poly).tolist())
+            elif type(order) == tuple:
+                order = order + tuple(np.arange(len(order), len(order) + dim_poly))
         A = math.gather(math.gather(A, order, axis=-1), order, axis=-2)
         b = math.gather(b, order, axis=-1)
-        if len(c.shape) == len(order):
-            c = math.transpose(c, order)
+        # if len(c.shape) == len(order):
+        #     c = math.transpose(c, order)
```
apchytr, line 272: Do these lines need to be uncommented or removed?

mrmustard/physics/gaussian_integrals.py
```diff
@@ -374,3 +380,92 @@ def complex_gaussian_integral_2(
...
+    c1 = math.astensor(c1)
+    c2 = math.astensor(c2)
+
+    dim_n1 = len(c1.shape)
+    dim_n2 = len(c2.shape)
+
+    dim_m1 = A1.shape[-1] - dim_n1
+    dim_m2 = A2.shape[-1] - dim_n2
+
+    A12 = np.block(
```
apchytr, line 410:
```suggestion
    A12 = math.block(
```
as well as the `np` calls below

mrmustard/physics/gaussian_integrals.py
kaspernielsen96: I asked Filippo what the purpose of this is, and I don't think we came to a conclusion. Everything works is we remove them, and I cannot see why we would want to reorder the polynomials in the new ansatz. So they should probably be removed.

mrmustard/physics/gaussian_integrals.py
```diff
@@ -374,3 +378,92 @@ def complex_gaussian_integral_2(
...
+    dim_m1 = A1.shape[-1] - dim_n1
+    dim_m2 = A2.shape[-1] - dim_n2
+
+    A12 = math.block(
+        [
+            [
+                math.cast(A1, "complex128")[:dim_m1, :dim_m1],
+                np.zeros((dim_m1, dim_m2)),
+                math.cast(A1, "complex128")[:dim_m1, dim_m1:],
+                np.zeros((dim_m1, dim_n2)),
```
ziofil, line 414: A1 should already have dtype complex128 no? 🤔
And you can pass the dtype to the zero array initializer `math.zeros((dim_m1, dim_m2), dtype=A1.dtype)`

mrmustard/physics/gaussian_integrals.py
```diff
@@ -374,3 +378,92 @@ def complex_gaussian_integral_2(
...
+            [
+                np.zeros((dim_n2, dim_m1)),
+                math.cast(A2, "complex128")[dim_m2:, :dim_m2],
+                np.zeros((dim_n2, dim_n1)),
+                math.cast(A2, "complex128")[dim_m2:, dim_m2:],
+            ],
+        ]
+    )
+    b12 = math.concat((b1[:dim_m1], b2[:dim_m2], b1[dim_m1:], b2[dim_m2:]), axis=-1)
+    c12 = math.reshape(math.outer(c1, c2), c1.shape + c2.shape)
```
ziofil, line 437: For now let's leave it in, but we should find a way to avoid outer products. E.g. instead of defining the contraction as a tensor product followed by a gaussian integral, we should define the gaussian integral directly between two Abcs. For another PR :)

ziofil (approved):
minor details. great work!

ziofil (approved):
👍

PR #451, opened 2024-08-05
Serialization

timmysilv:
### Before submitting

      test directory!

      `pip install black`.

**Context:**

**Benefits:**

**Possible Drawbacks:**

**Related GitHub Issues:**

### DIFF
--- a/mrmustard/lab_dev/circuit_components.py
+++ b/mrmustard/lab_dev/circuit_components.py
@@ -22,12 +22,14 @@ from __future__ import annotations
 from typing import Optional, Sequence, Union
 import numbers
 from functools import cached_property
+from pathlib import Path
 
 import numpy as np
 import ipywidgets as widgets
 from IPython.display import display
 
 from mrmustard import settings, math, widgets as mmwidgets
+from mrmustard.utils import serialize
 from mrmustard.utils.typing import Scalar, ComplexTensor
 from mrmustard.physics.representations import Representation, Bargmann, Fock
 from mrmustard.math.parameter_set import ParameterSet
@@ -91,6 +93,24 @@ class CircuitComponent:
             )
             if self._representation:
                 self._representation = self._representation.reorder(tuple(perm))
+
+    def serialize(self) -> Path:
+        r"""Serialize a CircuitComponent object."""
+        return serialize.save(
+            type(self),
+            name=self.name,
+            representation=str(self.representation.serialize()),
+            wires=self.wires.serialize(),
+        )
+
+    @classmethod
+    def deserialize(cls, data) -> CircuitComponent:
+        r"""Deserialize a CircuitComponent object."""
+        return cls._from_attributes(
+            serialize.load(data["representation"]),
+            Wires.deserialize(data["wires"]),
+            name=data["name"],
+        )
 
     @classmethod
     def _from_attributes(
--- a/mrmustard/physics/representations.py
+++ b/mrmustard/physics/representations.py
@@ -810,12 +810,12 @@ class Fock(Representation):
 
     def serialize(self) -> Path:
         r"""Serialize a Fock instance."""
-        return serialize.save(type(self), array=self.array, batched=self.batched)
+        return serialize.save(type(self), array=self.array)
 
     @classmethod
     def deserialize(cls, data) -> Fock:
         r"""Deserialize a Fock instance."""
-        return cls(data["array"], batched=data["batched"])
+        return cls(data["array"], batched=True)
 
     def _ipython_display_(self):
         w = widgets.fock(self)
--- a/mrmustard/physics/representations.py
+++ b/mrmustard/physics/representations.py
@@ -21,6 +21,8 @@ from __future__ import annotations
 from warnings import warn
 from abc import ABC, abstractmethod
 from typing import Any, Callable, Iterable, Union
+from pathlib import Path
+
 from matplotlib import colors
 import matplotlib.pyplot as plt
 import numpy as np
@@ -34,6 +36,7 @@ from mrmustard.physics.gaussian_integrals import (
     complex_gaussian_integral,
 )
 from mrmustard.physics.ansatze import Ansatz, PolyExpAnsatz, ArrayAnsatz
+from mrmustard.utils import serialize
 from mrmustard.utils.typing import (
     Batch,
     ComplexMatrix,
@@ -97,6 +100,15 @@ class Representation(ABC):
         The scalar part of the representation.
         For now it's ``c`` for Bargmann and the array for Fock.
         """
+
+    @abstractmethod
+    def serialize(self) -> Path:
+        r"""Serialize a Representation."""
+
+    @classmethod
+    @abstractmethod
+    def deserialize(cls, data) -> Representation:
+        r"""Deserialize a Representation."""
 
     def __eq__(self, other: Representation) -> bool:
         r"""
@@ -516,6 +528,18 @@ class Bargmann(Representation):
         A, b, c = zip(*Abc)
         return Bargmann(A, b, c)
 
+    def serialize(self) -> Path:
+        r"""Serialize a Bargmann instance."""
+        return serialize.save(type(self), A=self.A, b=self.b, c=self.c)
+
+    @classmethod
+    def deserialize(cls, data) -> Bargmann:
+        r"""Deserialize a Bargmann instance."""
+        A = data["A"]
+        b = data["b"]
+        c = data["c"]
+        return cls(A, b, c)
+
     def _ipython_display_(self):
         display(widgets.bargmann(self))
 
@@ -783,6 +807,15 @@ class Fock(Representation):
 
         ret = self.array[(slice(0, None),) + tuple(slice(0, s) for s in shape)]
         return Fock(array=ret, batched=True)
+
+    def serialize(self) -> Path:
+        r"""Serialize a Fock instance."""
+        return serialize.save(type(self), array=self.array, batched=self.batched)
+
+    @classmethod
+    def deserialize(cls, data) -> Fock:
+        r"""Deserialize a Fock instance."""
+        return cls(data["array"], batched=data["batched"])
 
     def _ipython_display_(self):
         w = widgets.fock(self)
--- a/mrmustard/utils/serialize.py
+++ b/mrmustard/utils/serialize.py
@@ -28,7 +28,6 @@ CACHE = Path(__file__).parents[2].absolute() / ".serialize_cache"
 def save(cls, **data) -> Path:
     r"""Save a serialized set of data to file for later deserialization."""
     file = CACHE / f"{cls.__qualname__}_{uuid4()}.json"  # random filename
-    data["file"] = str(file)
     data["class"] = f"{cls.__module__}.{cls.__qualname__}"
 
     # use numpy serialization tools, save filename in json
--- a/mrmustard/utils/serialize.py
+++ b/mrmustard/utils/serialize.py
@@ -1,0 +1,83 @@
+# Copyright 2024 Xanadu Quantum Technologies Inc.
+
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+
+#     http://www.apache.org/licenses/LICENSE-2.0
+
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""A serialization library for MrMustard."""
+
+import json
+from pathlib import Path
+from pydoc import locate
+from uuid import uuid4
+
+import numpy as np
+
+
+CACHE = Path(__file__).parents[2].absolute() / ".serialize_cache"
+
+
+def save(cls, **data) -> Path:
+    r"""Save a serialized set of data to file for later deserialization."""
+    file = CACHE / f"{cls.__qualname__}_{uuid4()}.json"  # random filename
+    data["file"] = str(file)
+    data["class"] = f"{cls.__module__}.{cls.__qualname__}"
+
+    # use numpy serialization tools, save filename in json
+    arrays = {k: v for k, v in data.items() if isinstance(v, np.ndarray)}
+    if len(arrays) > 1:  # many arrays, use zipped serialization
+        # save to file
+        npz = file.with_suffix(".npz")
+        np.savez(npz, **arrays)
+        # update json
+        _ = list(map(data.pop, arrays))
+        data["npz"] = str(npz)
+    elif len(arrays) == 1:
+        # save to file
+        npy = file.with_suffix(".npy")
+        [(k, arr)] = list(arrays.items())
+        np.save(npy, arr)
+        # update json
+        del data[k]
+        data["npy"] = (k, str(npy))
+
+    with file.open("w") as f:
+        json.dump(data, f)
+
+    return file
+
+
+def load(file: Path, remove_after=True):
+    r"""
+    The deserializer entrypoint for MrMustard.
+
+    Args:
+        file (Path): The file to load from
+        remove_after (Optional[bool]): Once load is complete, delete the saved file
+    """
+    file = Path(file)
+    with file.open() as f:
+        data = json.load(f)
+    if remove_after:
+        file.unlink()
+
+    if "npy" in data:
+        (k, npy_file) = data.pop("npy")
+        data[k] = np.load(npy_file)
+        if remove_after:
+            Path(npy_file).unlink()
+    elif "npz" in data:
+        npz_file = data.pop("npz")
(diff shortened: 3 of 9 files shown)


apchytr:
Might have missed it but does the serialization keep track of a version? Could be useful even if we arent concerned about backward compatibility

timmysilv:
it doesn't. can definitely add it, though I'm not sure what it'll add. I'd be curious to know more on how people intend on using this before adding more features. I kept it pretty barebones because I honestly don't know what we want from this tool, and kinda need more info before adding more advanced features to the module. that said, I can add version when saving and pop it when loading if we want that.

apchytr:
> it doesn't. can definitely add it, though I'm not sure what it'll add. I'd be curious to know more on how people intend on using this before adding more features. I kept it pretty barebones because I honestly don't know what we want from this tool, and kinda need more info before adding more advanced features to the module. that said, I can add version when saving and pop it when loading if we want that.

I agree we should keep it barebones for now but I do think theres some benefit including it. At the very least it would make it clear what version the saved files are associated with if breaking changes happen. That is to say only if we don't rely on the static `0.7.3` version. I think either it could be the latest commit hash from develop (temporary until we have actual releases) or we could start updating the pyproject.toml with an alpha (e.g. `0.7.3a1`). What do you think?

mrmustard/physics/representations.py
```diff
@@ -516,6 +528,18 @@ def __matmul__(self, other: Bargmann) -> Bargmann:
...
+        r"""Serialize a Bargmann instance."""
+        return save(type(self), arrays={"A": self.A, "b": self.b, "c": self.c})
+
+    @classmethod
+    def deserialize(cls, data) -> Bargmann:
+        r"""Deserialize a Bargmann instance."""
+        A = data["A"]
+        b = data["b"]
+        c = data["c"]
+        return cls(A, b, c)
```
zeyueN, line 541: Without context, this just looks like a dummy function 😂

mrmustard/physics/representations.py
```diff
@@ -784,6 +808,15 @@ def reduce(self, shape: Union[int, Iterable[int]]) -> Fock:
...
         return Fock(array=ret, batched=True)
 
+    def serialize(self) -> Path:
+        r"""Serialize a Fock instance."""
+        return save(type(self), arrays={"array": self.array})
+
+    @classmethod
+    def deserialize(cls, data) -> Fock:
+        r"""Deserialize a Fock instance."""
+        return cls(data["array"], batched=True)
```
zeyueN, line 818: Not knowing how batching works, just noticing the `__init__` has `batched=False`. Would this be able to work for unbatched stuff?

mrmustard/utils/serialize.py
```diff
@@ -0,0 +1,121 @@
...
+@contextmanager
+def cache_subdir(name=None):
+    r"""Context manager to have calls to ``save`` write to a cache subdirectory."""
+    global CACHE  # pylint:disable=global-statement
+    if not name:
+        name = f"subcache_{uuid4().hex}"
+
+    old_cache = CACHE
+    CACHE /= name
+    CACHE.mkdir()
```
zeyueN, line 119:
```suggestion
    CACHE.mkdir(parents=True)
```
seems the `".serialize_cache"` won't be there if subdir is used the first time serialization is used.

mrmustard/utils/serialize.py
```diff
@@ -0,0 +1,121 @@
...
+
+@contextmanager
+def cache_subdir(name=None):
+    r"""Context manager to have calls to ``save`` write to a cache subdirectory."""
+    global CACHE  # pylint:disable=global-statement
+    if not name:
+        name = f"subcache_{uuid4().hex}"
+
+    old_cache = CACHE
+    CACHE /= name
```
zeyueN, line 118: never thought of using `/=` with paths, nice!

zeyueN (commented):
Thanks a bunch! I like how this is done.

One questions, for circuits the wiring information saved in the json file in some form? Or one has to load the circuit and then get the wire info. In other words, if I just want to look at the tensor network graph of a circuit without loading the data arrays, is there an easy way of doing that?

Could you also paste a small sample json file here?

mrmustard/physics/representations.py
timmysilv: if you create it with `batched=False`, it just [adds a batch dim later in construction](https://github.com/XanaduAI/MrMustard/blob/4085e252ff8353e7769133ef7e74fedd88221766/mrmustard/physics/ansatze.py#L1052), so the internal representation is always batched. This also means that the serialized data is always batched, so we deserialize accordingly.

mrmustard/utils/serialize.py
timmysilv: it will be there because I'm adding `/.serialize_cache/.gitkeep` in this PR 😄

mrmustard/utils/serialize.py
timmysilv: came to me in a dream ☁️

timmysilv:
### Serializing one gate
```pycon
>>> Dgate([0], x=1.1).serialize()
PosixPath('/home/ubuntu/src/github.com/XanaduAI/MrMustard/.serialize_cache/Dgate_fe40cc9a65e940ec83850ade04a5cac4.json')
```
<details><summary>Filesystem after serializing a single gate</summary>

```
(mr_mustard) ~/src/github.com/XanaduAI/MrMustard (serialization ✔) tree .serialize_cache 
.serialize_cache
├── Bargmann_70ba7fe5deef479e8677775e096ef9a0.json
├── Bargmann_70ba7fe5deef479e8677775e096ef9a0.npz
└── Dgate_fe40cc9a65e940ec83850ade04a5cac4.json
```
</details>

<details><summary>The Dgate JSON</summary>

```json
{
  "name": "Dgate",
  "representation": "/home/ubuntu/src/github.com/XanaduAI/MrMustard/.serialize_cache/Bargmann_70ba7fe5deef479e8677775e096ef9a0.json",
  "wires": "[[], [], [0], [0]]",
  "class": "mrmustard.lab_dev.transformations.dgate.Dgate"
}
```
</details>

<details><summary>The Bargmann JSON</summary>

```json
{
  "class": "mrmustard.physics.representations.Bargmann",
  "arrays": "/home/ubuntu/src/github.com/XanaduAI/MrMustard/.serialize_cache/Bargmann_70ba7fe5deef479e8677775e096ef9a0.npz",
  "backend": "numpy"
}
```
</details>

### Serializing a circuit

```pycon
>>> Circuit([Dgate([0], x=1.1)]).serialize()
PosixPath('/home/ubuntu/src/github.com/XanaduAI/MrMustard/.serialize_cache/Circuit_82b2f9e1484d42ed9feb0d9cca6299a6.json')
```
<details><summary>Filesystem after also serializing a circuit (new zip and json)</summary>

```
(mr_mustard) ~/src/github.com/XanaduAI/MrMustard (serialization ✔) tree .serialize_cache
.serialize_cache
├── Circuit_82b2f9e1484d42ed9feb0d9cca6299a6.json
└── collection_3591dab7fd1241bdaf2a45a7fce1134f.zip
```
</details>

<details><summary>The Circuit JSON</summary>

```json
{
  "order": [
    "Dgate_9ec072981e4a4a429c2c9e9bc4eb24f3.json"
  ],
  "zipfile": "/home/ubuntu/src/github.com/XanaduAI/MrMustard/.serialize_cache/collection_3591dab7fd1241bdaf2a45a7fce1134f.zip",
  "subdir": "/home/ubuntu/src/github.com/XanaduAI/MrMustard/.serialize_cache/subcache_ca97053f8d994beb99a74fff29f26d0a",
  "class": "mrmustard.lab_dev.circuits.Circuit"
}
```
</details>

timmysilv:
since the last review, I've made two notable changes:
- the cache directory is now in `settings`, so you can configure it easily (no more global hehe)
- the `remove_after` kwarg for `load` is now *False* by default

@apchytr I've added basic versioning just using `__version__`. I'm open to adding more informative versioning in the future but I find the alpha suffixes cumbersome to maintain, so I'd rather not do it unless we see clear value. Perhaps this can be motivation to make things more stable? So you may only break things between versions (even the most minor version change)

@zeyueN I've posted examples of all the things above 😄

timmysilv:
ok since that last comment things have really changed. Check out the updated PR description for details, but basically only `Circuit.serialize()` writes to file now.

timmysilv:
@zeyueN I just added the `path` property to the circuit serialization. Note that it's only evaluated if the user called `make_path`, so it'll be empty otherwise. I don't save the `_graph` property, because it depends on non-serialized/non-functional data (a random number tied to the wires object that isn't serialized either)

tests/test_utils/test_serialize.py
```diff
@@ -142,46 +162,48 @@ def test_backend_change_error(self, cache_dir, monkeypatch):
...
-    r"""
-    Test that serializing then deserializing a MrMustard object creates an equivalent instance.
-    """
-    obj = obj()
-    assert load(obj.serialize()) == obj
-    assert not list(cache_dir.glob("*"))
+        loaded = load(path)
+        assert loaded == circ
+        assert all(type(a) == type(b) for a, b in zip(circ.components, loaded.components))
+        assert list(path.parent.glob("*")) == [path]
```
zeyueN, line 209: nice roundtrip checks!

zeyueN (approved):
The changes make sense. Looking great! Thanks!

.serialize_cache/.gitkeep
apchytr, line 1: Sorry, maybe its mentioned somewhere but what is this file for?

mrmustard/lab_dev/circuit_components.py
```diff
@@ -95,6 +98,46 @@ def __init__(
                 if self._representation:
                     self._representation = self._representation.reorder(tuple(perm))
 
+    def serialize(self) -> Tuple[Dict[str, Any], List[Tuple[str, ArrayLike]]]:
```
apchytr, line 101: We could avoid typing imports if we just used `tuple`, `dict`, `any`, etc

mrmustard/physics/representations.py
```diff
@@ -774,6 +794,15 @@ def __matmul__(self, other: Fock) -> Fock:
                 batched_array.append(math.tensordot(reduced_s.array[i], reduced_o.array[j], axes))
         return self.from_ansatz(ArrayAnsatz(batched_array))
 
+    def serialize(self) -> list[tuple[str, ArrayLike]]:
+        """Serialize a Fock instance."""
+        return [("array", self.data)]
```
apchytr, line 799: I wonder if there is some way to set this up such that we have a generic serialize class that everything serializable inherits from and we define the attrs on each of these classes telling the serializer what to serialize (maybe through `__slots__`?). Something like that would streamline a lot of this I believe

.serialize_cache/.gitkeep
timmysilv: so we don't call `mkdir(exist_ok=True)` every time someone calls `save` :p
it's just a convention to keep a folder around because git doesn't track folders, only files

mrmustard/lab_dev/circuit_components.py
timmysilv: tbh I wasn't sure when to use the `typing` aliases, but I see that a lot of it is historical. I'll remove the ones that don't add much value

mrmustard/physics/representations.py
timmysilv: I didn't want to over-engineer this, so I felt that adhering to a simple standard was sufficient, the standard being that representations return a list that is ready to be turned into a dict. We could also do something like `__slots__`, but I'd prefer to leave room for some flexibility without making the code too ugly (and this feels quite compact to me 😄 )

I also considered having representations follow the same standard as circuit components (return two items: a serializable dict, and non-serializable data in this format), then have the circuit component merge its data with its representation's data. Decided against it because this is more compact and equally consistent. lmk if you have a strong preference any which way

mrmustard/physics/representations.py
```diff
@@ -97,6 +100,15 @@ def reorder(self, order: tuple[int, ...] | list[int]) -> Representation:
         Reorders the representation indices.
         """
 
+    @abstractmethod
+    def serialize(self) -> dict[str, ArrayLike]:
```
apchytr, line 104: Does it make sense to rename these to `to_dict` and `from_dict` instead?

mrmustard/physics/representations.py
```diff
@@ -510,6 +522,15 @@ def __matmul__(self, other: Bargmann) -> Bargmann:
         A, b, c = zip(*Abc)
         return Bargmann(A, b, c)
 
+    def serialize(self) -> dict[str, ArrayLike]:
+        """Serialize a Bargmann instance."""
+        return {"A": self.A, "b": self.b, "c": self.c}
```
apchytr, line 527: Do we want to keep track of the contraction idxs as well? I'm guessing no for a couple of reasons but wanted to make sure

mrmustard/physics/representations.py
apchytr: I think actually it would. It looks like the use case is to call `serialize` on the circuit itself. Renaming would made that workflow clearer I believe

timmysilv: agreed, done!

mrmustard/physics/representations.py
timmysilv: I'm open to it, just didn't think to store them before. what were the reasons against?

apchytr: I figured maybe storing something that isn't an argument in the init would mess things up. Also, if we expect to only be using on `Circuit` then I'm under the impression the contraction idxs aren't really used? If it's simple enough to include then sure just didnt want to introduce unnecessary overhead

mrmustard/lab_dev/circuits.py
```diff
@@ -427,6 +430,45 @@ def validate_path(self, path) -> None:
...
             remaining[i1] = (remaining[i1] @ remaining.pop(i2))[0]
 
+    def serialize(self, filestem=None):
+        r"""
+        Serialize a Circuit.
+
+        Keyword Args:
+            filestem (str): An optional name to give the resulting file saved to disk.
+
+        All components must implement the following in order to serialize correctly:
```
apchytr, line 440: Hmm not sure if this needs to be mentioned in the docstring? It almost reads like the docstring of a hidden method (not exposed to the user)

mrmustard/lab_dev/circuits.py
```diff
@@ -427,6 +430,45 @@ def validate_path(self, path) -> None:
 
             remaining[i1] = (remaining[i1] @ remaining.pop(i2))[0]
 
+    def serialize(self, filestem=None):
+        r"""
+        Serialize a Circuit.
+
+        Keyword Args:
```
apchytr, line 437:
```suggestion
        Args:
```
This should be moved to the bottom of the docstring for consistency

mrmustard/lab_dev/circuits.py
```diff
@@ -427,6 +430,45 @@ def validate_path(self, path) -> None:
 
             remaining[i1] = (remaining[i1] @ remaining.pop(i2))[0]
 
+    def serialize(self, filestem=None):
```
apchytr, line 433:
```suggestion
    def serialize(self, filestem: str=None):
```
Maybe even optionally a Path?

mrmustard/lab_dev/circuits.py
```diff
@@ -427,6 +430,45 @@ def validate_path(self, path) -> None:
...
+        r"""
+        Serialize a Circuit.
+
+        Keyword Args:
+            filestem (str): An optional name to give the resulting file saved to disk.
```
apchytr, line 438:
```suggestion
            filestem: An optional name to give the resulting file saved to disk.
```

mrmustard/lab_dev/circuits.py
apchytr: Maybe move to the docstring of the CC serialize?

mrmustard/lab_dev/circuit_components.py
```diff
@@ -104,6 +107,45 @@ def __init__(
                 if self._representation:
                     self._representation = self._representation.reorder(tuple(perm))
 
+    def serialize(self) -> tuple[dict[str, Any], dict[str, ArrayLike]]:
```
apchytr, line 110: I think we should rename these as well but not sure if `to_dict` still works here?

timmysilv: I think the use-case is too bespoke here for a general name - the function creates a tuple of dicts, one serializable and one not. I believe its only use is for serialization, so I couldn't think of a better name personally

apchytr: `to_serialize`? lol. Maybe just make it a hidden method since we dont expect users to call it 👍

mrmustard/physics/representations.py
timmysilv: I think I will leave it out for now, especially because we almost never serialize the representation directly (only if the circuitcomponent isn't a known type with modes defined)

mrmustard/lab_dev/circuit_components.py
timmysilv: done!

apchytr (approved):
Looks great thanks!

PR #452, opened 2024-08-06
QuadratureEigenstate Fixes 

apchytr:
**Context:** QuadratureEigenstate `__repr__` currently fails due to an L2 norm calculation. Solution is to override the L2_norm. Additionally, set manual shape on QuadratureEigenstate for visualization purposes. 

**Description of the Change:** L2_norm on QuadratureEigenstate returns np.inf, manual_shape is set to (50,), some additional cleanup.


ziofil (approved):
👍

PR #453, opened 2024-08-07
Move reduce to ArrayAnsatz and include calls in dunder methods

apchytr:
**Context:** Dunder methods on ArrayAnsatz such as `__add__` currently fail if the shapes of the arrays do not match (e.g. adding a `(1, 2, 2)` and a `(1, 3, 3)` array). This PR includes calls to reduce in these dunder methods exploiting the fact we can just pad the smaller array with zeros making them compatible.

**Description of the Change:** `reduce` has been moved to `ArrayAnsatz` and `__add__`, `__mul__` and `__truediv__` make use of calls to `reduce`. Additionally, some cleanup such as reordering methods alphabetically. 

**Benefits:** Interactions between ArrayAnsatz of different fock cutoffs are more flexible

### DIFF
--- a/mrmustard/physics/ansatze.py
+++ b/mrmustard/physics/ansatze.py
@@ -1247,7 +1247,15 @@ class ArrayAnsatz(Ansatz):
         """
         if isinstance(other, ArrayAnsatz):
             try:
-                new_array = [a / b for a in self.array for b in other.array]
+                diff = sum(self.array.shape[1:]) - sum(other.array.shape[1:])
+                if diff < 0:
+                    new_array = [
+                        a / b for a in self.reduce(other.array.shape[1:]).array for b in other.array
+                    ]
+                else:
+                    new_array = [
+                        a / b for a in self.array for b in other.reduce(self.array.shape[1:]).array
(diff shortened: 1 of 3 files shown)


mrmustard/physics/ansatze.py
```diff
@@ -1092,6 +1100,35 @@ def from_function(cls, fn: Callable, **kwargs: Any) -> ArrayAnsatz:
...
+            return self
+        length = self.num_vars
+        shape = (shape,) * length if isinstance(shape, int) else shape
+        if len(shape) != length:
+            msg = f"Expected shape of length {length}, "
+            msg += f"given shape has length {len(shape)}."
+            raise ValueError(msg)
+
+        if any(s > t for s, t in zip(shape, self.array.shape[1:])):
+            warn(
```
apchytr, line 1120: Do we still want this warning? Or should the recommendation just be included in the docstring

ziofil: perhaps it is a bit too annoying indeed

ziofil (approved):
good!

PR #454, opened 2024-08-07
only bound thelwarus by major version

timmysilv:
https://open.spotify.com/track/0qGQXQdUNwPYQadGuELgEI?si=a00585ac1b99477e


apchytr (approved):
Thanks!

PR #455, opened 2024-08-09
try to reduce complexity of circuits.py

timmysilv:
Codefactor always complains of complex code due to circuits.py - I've made 3 changes:
- inline the `_ixdict` builder, it's only used in that once place
- inline the `propagate_shapes` helper, again, only used in that one place
- move `lookup_path` to a function that takes a Circuit, instead of it being an instance method

### DIFF
--- a/mrmustard/lab_dev/circuits.py
+++ b/mrmustard/lab_dev/circuits.py
@@ -94,29 +94,31 @@ class Circuit:
         This dictionary is used to propagate the shapes of the components in the circuit.
         """
         if not hasattr(self, "_idxdict"):
-            self._idxdict = self._indices_dict()
+            self._idxdict = {}
+            for i, opA in enumerate(self.components):
+                out_idx = set(opA.wires.output.indices)
+                indices: dict[int, dict[int, int]] = {}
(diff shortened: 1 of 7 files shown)


mrmustard/lab_dev/circuits.py
```diff
@@ -624,3 +496,120 @@ def component_to_str(comp: CircuitComponent) -> list[str]:
             ret += "\n\n"
 
         return ret
+
+
+def lookup_path(circuit: Circuit) -> None:
```
apchytr, line 501: I'm not sure how I feel about it being a standalone method but maybe this should be in a `circuits_utils` folder? We could even include `_indices_dict` there

apchytr: Maybe even `component_to_str` should be in circuit_components_utils

apchytr: The point is that we have a lot of fairly large class methods that maybe make sense to be standalone util methods significantly cleaning up our files

timmysilv:
better ways to go about this

PR #456, opened 2024-08-12
Adding classical wires

apchytr:
**Context:** It is apparent that Mr Mustard requires some notion of 'classical wire' to support ~~both the new poly exp ansatz as well as~~ measurements. Here, I introduce `classical_output` and `classical_input` (or some other name we can agree upon) into `Wires` that is meant to fill this role. The idea is that `Wires` will keep track of the modes that contain this classical information (associating it with an index and unique ID) and the representation will do the rest (in a followup PR). 

**Description of the Change:** `Wires` now accepts `classical_output` and `classical_input` meant to fill the role of non-quantum wires. The behaviour (within Wires) is identical to bra and ket.

**Benefits:** We will now be able to link classical information between circuit components. This approach avoids the need to introduce several new `Wires` classes. 

**Possible Drawbacks:** It's possible there are some use cases that may require us to reconsider this approach? ~~Maybe it turns out we really do need separate classes (i.e `ClassicalWire`, `QuantumWire`, `PolyWire`).~~


mrmustard/lab_dev/wires.py
```diff
@@ -176,6 +187,42 @@ def __init__(
...
+        return Wires(
+            self.args[2], self.args[3], self.args[0], self.args[1], self.args[4], self.args[5]
+        )
+
+    @cached_property
+    def bra(self) -> Wires:
+        r"""
+        New ``Wires`` object with only bra wires.
+        """
+        ret = Wires(self.args[0], self.args[1], set(), set())
```
kaspernielsen96, line 204: Do you need two more empty sets here so the total number of sets is 6?

mrmustard/lab_dev/wires.py
```diff
@@ -240,108 +285,112 @@ def index_dicts(self) -> list[dict[int, int]]:
...
         ret._original = self.original or self  # pylint: disable=protected-access
         return ret
 
     @cached_property
     def ket(self) -> Wires:
-        r"New ``Wires`` object without bra wires."
+        r"""
+        New ``Wires`` object with only ket wires.
+        """
         ret = Wires(set(), set(), self.args[2], self.args[3])
```
kaspernielsen96, line 318: Also 6 sets here?

mrmustard/lab_dev/wires.py
apchytr: Not really, tbh the last two empty sets aren't even needed. The reason is that it will default to `None` which is already treated as `set()` in the `__init__`.

kaspernielsen96 (commented):
Looks great. Thanks for clarifying.

kaspernielsen96 (approved):

timmysilv:
I'm a bit late to the game here, but `widgets.wires()` should probably be updated to draw the classical wires as well - it can be done in another PR, not a big deal

PR #457, opened 2024-08-12
Fock to bargmann

kaspernielsen96:
**Context:**
Adding a method to convert circuit components from Fock to Bargmann
When contracting a bargmann and a fock object with each other, the fock object is converted to bargmann. This is opposite of what is currently done. If two fock objects meet each other they stay in fock.
**Benefits:**
- Allows all the benefits from the new ansatz without changing the API. (i.e. `Number.dual` as a measurement device). 
- All objects in fock will automatically be converted to bargmann, which is more efficient for contracting with other objects down the line. It will create a "identity" Bargmann object with the Fock coefficients attached as the array (c)

**Possible Drawbacks:**

**Related GitHub Issues:**


mrmustard/lab_dev/circuit_components.py
```diff
@@ -409,21 +410,22 @@ def fock(self, shape: Optional[int | Sequence[int]] = None, batched=False) -> Co
...
-                f"Expected Fock shape of length {len(auto_shape)}, got length {len(shape)}"
-            )
-
+            shape = (shape,) * num_vars
         try:
             As, bs, cs = self.bargmann_triple(batched=True)
             if self.representation.ansatz.polynomial_shape[0] == 0:
+                auto_shape = self.auto_shape()
+                shape = shape or auto_shape
+                if len(shape) != len(auto_shape):
```
ziofil, line 421:
```suggestion
                shape = shape or self.auto_shape()
                if len(shape) != num_vars:
```
this way we don't call `auto_shape` when it's not needed

mrmustard/lab_dev/circuit_components.py
```diff
@@ -409,21 +410,22 @@ def fock(self, shape: Optional[int | Sequence[int]] = None, batched=False) -> Co
...
+                    raise ValueError(
+                        f"Expected Fock shape of length {len(auto_shape)}, got length {len(shape)}"
```
ziofil, line 423:
```suggestion
                        f"Expected Fock shape of length {num_vars}, got length {len(shape)}"
```

mrmustard/lab_dev/circuit_components.py
```diff
@@ -434,6 +436,12 @@ def fock(self, shape: Optional[int | Sequence[int]] = None, batched=False) -> Co
                     for A, b, c in zip(As, bs, cs)
                 ]
         except AttributeError:
+            auto_shape = self.auto_shape()
+            shape = shape or auto_shape
+            if len(shape) != len(auto_shape):
+                raise ValueError(
+                    f"Expected Fock shape of length {len(auto_shape)}, got length {len(shape)}"
+                )
```
ziofil, line 444:
```suggestion
            shape = shape or self.auto_shape()
            if len(shape) != num_vars:
                raise ValueError(
                    f"Expected Fock shape of length {num_vars}, got length {len(shape)}"
                )
```

mrmustard/lab_dev/circuit_components.py
```diff
@@ -531,6 +539,33 @@ def to_fock(self, shape: int | Sequence[int] | None = None) -> CircuitComponent:
...
+
+            >>> assert d_bargmann.name == d.name
+            >>> assert d_bargmann.wires == d.wires
+            >>> assert isinstance(d_bargmann.representation, Bargmann)
+        """
+        # try:
+        #     bargmann = Bargmann(*self.bargmann_triple(batched=True))
+        #     bargmann._original_fock_data = self.representation.data
+        #     return self._from_attributes(bargmann, self.wires, self.name)
+        # except AttributeError:
```
ziofil, line 563:
```suggestion
```

mrmustard/physics/representations.py
```diff
@@ -439,10 +439,10 @@ def trace(self, idx_z: tuple[int, ...], idx_zconj: tuple[int, ...]) -> Bargmann:
...
             Bargmann: the ansatz with the given indices traced over
         """
-        if self.ansatz.degree > 0:
-            raise NotImplementedError(
-                "Partial trace is only supported for ansatze with polynomial of degree ``0``."
-            )
+        # if self.ansatz.degree > 0:
+        #     raise NotImplementedError(
+        #         "Partial trace is only supported for ansatze with polynomial of degree ``0``."
+        #     )
```
ziofil, line 445:
```suggestion
```

ziofil (approved):
🚀

mrmustard/lab_dev/circuit_components.py
```diff
@@ -409,21 +410,22 @@ def fock(self, shape: Optional[int | Sequence[int]] = None, batched=False) -> Co
...
+                shape = shape or auto_shape
+                if len(shape) != len(auto_shape):
+                    raise ValueError(
+                        f"Expected Fock shape of length {len(auto_shape)}, got length {len(shape)}"
+                    )
                 arrays = [math.hermite_renormalized(A, b, c, shape) for A, b, c in zip(As, bs, cs)]
             elif self.representation.ansatz.polynomial_shape[0] > 0:
-                num_vars = self.representation.ansatz.num_vars
+                if not shape:
+                    shape = (settings.AUTOCUTOFF_MAX_CUTOFF,) * num_vars
```
apchytr, line 428: Can you do the same trick as before
```suggestion
                    shape = shape or (settings.AUTOCUTOFF_MAX_CUTOFF,) * num_vars
```

mrmustard/lab_dev/circuit_components.py
```diff
@@ -531,6 +539,33 @@ def to_fock(self, shape: int | Sequence[int] | None = None) -> CircuitComponent:
...
+        """
+        # try:
+        #     bargmann = Bargmann(*self.bargmann_triple(batched=True))
+        #     bargmann._original_fock_data = self.representation.data
+        #     return self._from_attributes(bargmann, self.wires, self.name)
+        # except AttributeError:
+        A, b, _ = identity_Abc(len(self.wires))
+        bargmann = Bargmann(A, b, self.representation.data)
+        bargmann._original_fock_data = self.representation.data
+        return self._from_attributes(bargmann, self.wires, self.name)
```
apchytr, line 567: Should we have a 
```
if isinstance(self.representation, Bargmann):
     return self
```
somewhere here?

kaspernielsen96: I had that originally, but when everything can now be converted to bargmann I ended up having to use `to_fock()` way too much (so much it is no longer intuitive) to get mm to do what I want when contracting circuits. 

If we want the original bargmann data to be used, it should only be used in cases where the `c` in the triple is a scalar.

kaspernielsen96: I added this as a feature that we only use original bargmann data if c is a scalar.

kaspernielsen96: We ended up with a solution where we do not use the original bargmann data.

mrmustard/lab_dev/circuit_components.py
```diff
@@ -637,26 +671,18 @@ def __matmul__(self, other: CircuitComponent | Scalar) -> CircuitComponent:
...
-            self_rep = self.representation.reduce(self_shape)
-        else:
-            self_rep = self.to_fock(self_shape).representation
-        if isinstance(other.representation, Fock):
-            other_rep = other.representation.reduce(other_shape)
-        else:
-            other_rep = other.to_fock(other_shape).representation
+            self_rep = self.representation
+            other_rep = other.representation
+        elif isinstance(self.representation, Fock) and isinstance(other.representation, Fock):
```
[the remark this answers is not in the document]
apchytr: Then I think the bottom two you could call to_bargmann on both reps reducing it further

mrmustard/lab_dev/wires.py
```diff
@@ -214,6 +214,20 @@ def classical(self) -> Wires:
         ret._original = self.original or self  # pylint: disable=protected-access
         return ret
 
+    @cached_property
+    def quantum(self) -> Wires:
```
apchytr, line 218: Can we get a test for this

mrmustard/lab_dev/circuit_components.py
```diff
@@ -594,7 +622,13 @@ def _rshift_return(
         "internal convenience method for right-shift, to return the right type of object"
         if len(ret.wires) > 0:
             return ret
-        scalar = ret.representation.scalar
+        if len(ret.representation.data) > 1:
```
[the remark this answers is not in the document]
kaspernielsen96: No this is when a bargmann object has no wires but has polynomials (and thus also an A and b still). Before we would just return a scalar here, but now we have to evaluate the A,b and polynomial and return that value.

mrmustard/lab_dev/circuit_components.py
```diff
@@ -531,6 +537,28 @@ def to_fock(self, shape: int | Sequence[int] | None = None) -> CircuitComponent:
...
+
+
+            >>> assert d_bargmann.name == d.name
+            >>> assert d_bargmann.wires == d.wires
+            >>> assert isinstance(d_bargmann.representation, Bargmann)
+        """
+        A, b, _ = identity_Abc(len(self.wires.quantum))
+        bargmann = Bargmann(A, b, self.representation.data)
+        bargmann._original_fock_data = self.representation.data
+        return self._from_attributes(bargmann, self.wires, self.name)
```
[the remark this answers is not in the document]
kaspernielsen96: I may be doing something wrong, but it is not happy if I try to do it this way.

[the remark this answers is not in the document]
apchytr: Which part is it unhappy with? In principal this should just be regenerating the object and replacing the representation (this time with Bargmann as opposed to Fock)

mrmustard/lab_dev/circuit_components.py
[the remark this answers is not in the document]
kaspernielsen96: Right, the first check is if it is bargmann and the next is if it has a polynomial.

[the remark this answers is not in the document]
apchytr: I think this first check breaks down no? The Fock array can have len > 1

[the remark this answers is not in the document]
apchytr: Might make more sense to instead do a try except and use .triples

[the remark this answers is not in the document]
kaspernielsen96: No the first one would not break down (the fock array is `.data[0]`). But I changed the implementation to something more clear.

mrmustard/lab_dev/circuit_components.py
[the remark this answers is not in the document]
kaspernielsen96: Managed to get it to work.

ziofil (approved):
let's gooooo

ziofil (approved):
let's gooooo

mrmustard/lab_dev/circuit_components.py
```diff
@@ -594,7 +632,13 @@ def _rshift_return(
         "internal convenience method for right-shift, to return the right type of object"
         if len(ret.wires) > 0:
             return ret
-        scalar = ret.representation.scalar
+        if isinstance(ret.representation, Bargmann):
```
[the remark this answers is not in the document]
kaspernielsen96: The `if len(ret.wires) > 0:` may also have to change with the notion of classical and quantum wires. But I guess that is for a different PR.

apchytr (approved):

PR #460, opened 2024-08-13
better defaults

ziofil:
Now `settings.HBAR` defaults to 1.0 and `settings.AUTOSHAPE_PROBABILITY` defaults to 0.99999 (99.999%).


mrmustard/lab_dev/transformations/amplifier.py
```diff
@@ -41,7 +41,7 @@ class Amplifier(Channel):
         >>> from mrmustard.lab_dev import Amplifier, Coherent
 
         >>> amp = Amplifier([0], gain=4)
-        >>> coh = Coherent([0], x=1.0, y=2.0)  # units of hbar=2 (default)
+        >>> coh = Coherent([0], x=1.0 * np.sqrt(2), y=2.0 * np.sqrt(2))  # units of hbar=1 (default)
```
apchytr, line 44: Thinking about user convenience should we try to make x and y hbar independent?

mrmustard/utils/settings.py
```diff
@@ -15,7 +15,7 @@
 """A module containing global settings."""
 
 from __future__ import annotations
-
+from typing import Optional
```
apchytr, line 18:
```suggestion
```

mrmustard/utils/settings.py
```diff
@@ -153,15 +149,15 @@ def HBAR(self, value: float):
...
+    def SEED(self) -> int:
         r"""Returns the seed value if set, otherwise returns a random seed."""
         if self._seed is None:
             self._seed = np.random.randint(0, 2**31 - 1)
             self.rng = np.random.default_rng(self._seed)
         return self._seed
 
     @SEED.setter
-    def SEED(self, value: int):
+    def SEED(self, value: Optional[int]):
```
apchytr, line 160:
```suggestion
    def SEED(self, value: int | None):
```

mrmustard/lab_dev/transformations/amplifier.py
ziofil: The Bargmann representation is hbar-independent but coordinates in phase space are hbar-dependent by definition. I either had to put a sqrt(2) here or change the allclose statement below where I check the value of mu (which is hbar dependent). I'll change the value below to make it clear.

mrmustard/lab_dev/transformations/amplifier.py
```diff
@@ -41,7 +41,7 @@ class Amplifier(Channel):
...
         >>> _, mu, _ = (coh >> amp).phase_space(0)
         >>> assert np.allclose(mu[0], np.array([4.0, 8.0]))
```
ziofil, line 46:
```suggestion
        >>> coh = Coherent([0], x=1.0, y=2.0)
        >>> _, mu, _ = (coh >> amp).phase_space(0)
        >>> assert np.allclose(mu[0]*np.sqrt(2/settings.HBAR), np.array([4.0, 8.0]))
```

apchytr (approved):

PR #461, opened 2024-08-13
Potential workaround for _from_attributes

apchytr:
**Context:** Atm calling e.g `DGate.to_fock` returns a `Unitary` instead of a `DGate` due to `_from_attributes`. Here is a proposed alternative where we attempt to initialize the original class using the `ParameterSet` and default to `_from_attributes` if that fails.

**Description of the Change:** `to_fock` will attempt to initialize `self.__class__ ` using `_getitem_builtin` (moved from `State`) and if it fails with a `TypeError` will default to `_from_attributes`. `Number` and `Sauron` now populate their parameter sets. `ParameterSet` now has a `to_dict` method. Lots of code cleanup/ appeasing codefactor/code coverage.

**Benefits:** `to_fock` returns more user facing classes. This helps us avoid the need to overwrite `_from_attributes` in some cases.


mrmustard/lab_dev/circuit_components.py
apchytr, line 1: Large diff here because I reordered things (properties -> classmethods -> methods -> dunder methods) in alphabetical order.

mrmustard/lab_dev/circuit_components.py
```diff
@@ -489,65 +470,70 @@ def to_fock(self, shape: int | Sequence[int] | None = None) -> CircuitComponent:
...
-        return self._from_attributes(rep, self.wires, name)
+        if parameter.value.shape != ():
+            length = sum(parameter.value.shape)
+            if length != 1 and length != len(self.modes):
+                raise ValueError(f"Length of ``{parameter.name}`` must be 1 or {len(self.modes)}.")
+        self.parameter_set.add_parameter(parameter)
+        self.__dict__[parameter.name] = parameter
 
-    def __mul__(self, other: Scalar) -> CircuitComponent:
+    def _from_param_set(
```
timmysilv, line 499: I can't quite put my finger on it, but something feels weird about this. It's effectively just a copy, then the representation gets swapped out. we probably don't need `name` or `wires`, since those are handled by `self.__class__` and `self.modes` inherently.

would it make more sense to repurpose this work, and instead implement `__copy__` for `CircuitComponent`?

apchytr: Agreed. The main reason for separating it out this way was the hope that serialization might make use of it. Tbh the better way I think is to just move `_getitem_builtin_state` (the method I copied the logic from) to `CircuitComponent` as `_getitem_builtin`.

mrmustard/lab_dev/circuit_components.py
apchytr: Yes, including the change to `to_fock`

mrmustard/lab_dev/circuit_components.py
timmysilv: I didn't know this existed!! that sounds pretty good to me if it can be generalized, but the problem I hit was that `self.modes` isn't _necessarily_ defined. In practice, it is. All concrete classes have it. Only intermediate abstract classes (`Operation`, `Map`, `Unitary` and `Channel`) are constructed differently (using `modes_in` and `modes_out`), as well as the base types (`CircuitComponent`, `Transformation` and `State`) which are constructed using `wires`. Evidently, we assume in `State` that `self.modes` is defined, so we can keep that assumption.

apchytr: Yeah, I agree this was something that bothered me also. I think since it is a hidden method if we just document that this is meant to only work on user facing classes it should be fine? Otherwise then I guess this is a case where we just have to have some duplicate code and I can keep it in `to_fock`

timmysilv (approved):
I like it! I don't wanna drag this out much so I'll give my approval, but I do have an alternate proposal. Perhaps instead of moving `_getitem_builtin` up, we should just add the functionality to `ParameterSet` - something like `todict()`? then the `modes` quirk can have its own handling in one place, and the parameter kwargs generation can be in another place. thoughts? if you do choose to make the change, I promise to come back with another quick approval 😇

apchytr:
> I like it! I don't wanna drag this out much so I'll give my approval, but I do have an alternate proposal. Perhaps instead of moving `_getitem_builtin` up, we should just add the functionality to `ParameterSet` - something like `todict()`? then the `modes` quirk can have its own handling in one place, and the parameter kwargs generation can be in another place. thoughts? if you do choose to make the change, I promise to come back with another quick approval 😇

@timmysilv pushed a `to_dict` method on `ParameterSet`. It's mostly just taking the kwargs logic from the previous `_getitem_builtin` 👍

timmysilv (approved):
nice!

PR #463, opened 2024-08-15
increase min cutoff for visualize2d/3d to 50

ziofil:
Visualize2d/3d can show a wrong image when plotting non-DM wigner functions, such as the parity operator. This PR guarantees a minimum fock cutoff of 50 for the wigner plot, which solves most of these issues.

Before:

Screenshot 2024-08-15 at 11 56 02 AM

After:

Screenshot 2024-08-15 at 11 56 32 AM

Also 3d (after):
Screenshot 2024-08-15 at 11 57 57 AM


mrmustard/lab_dev/circuit_components.py
```diff
@@ -471,12 +471,16 @@ def to_fock(self, shape: int | Sequence[int] | None = None) -> CircuitComponent:
...
+        try:
+            fock._original_bargmann_data = self.representation.triple
+        except AttributeError:
+            fock._original_bargmann_data = None
         try:
             ret = self._getitem_builtin(self.modes)
             ret._representation = fock
         except TypeError:
             ret = self._from_attributes(fock, self.wires, self.name)
+        ret.manual_shape = ret.representation.array.shape[1:]
```
timmysilv, line 483: this should be handled by the cached_property right?

ziofil: yes you're right, what's the right way to do it?

timmysilv: I believe you just delete it (so that the next time it's called, it gets re-computed):
```suggestion
        del ret.manual_shape
```

from [the docs](https://docs.python.org/3/library/functools.html#functools.cached_property)
> The cached value can be cleared by deleting the attribute. This allows the cached_property method to run again.

ziofil: hmm, but I want to update it, not delete it. Maybe I should do it entry-wise?

timmysilv: the next time it's invoked, it'll [re-evaluate](https://github.com/XanaduAI/MrMustard/blob/4ac64c1be84ee98ba7290755e52c8c3e5f11a0db/mrmustard/lab_dev/circuit_components.py#L202) (update) it, and it should be correct. Example:
```python
from functools import cached_property

class X:
    def __init__(self, x):
        self.x = x

    @cached_property
    def xsum(self):
        return sum(self.x)
```

```pycon
>>> foo = X([1,2,3])
>>> foo.xsum
6
>>> foo.x = [4,5,6]
>>> foo.xsum
6  # BAD! didn't update
>>> del foo.xsum
>>> foo.xsum
15 
```

ziofil: oooh silly me. I had forgotten it already takes the shape of the array if it exists.

timmysilv:
i can't speak to whether there are cases where this minimum doesn't really make sense, but I am guessing you can - do any instances come to mind where we don't want this minimum, and if so, can we use the new kwarg in those cases? otherwise, this looks good to me

ziofil:
> i can't speak to whether there are cases where this minimum doesn't really make sense, but I am guessing you can - do any instances come to mind where we don't want this minimum, and if so, can we use the new kwarg in those cases? otherwise, this looks good to me

We'll replace the discretized method with calling the new ansatz as soon as it's available. In the meantime this is  good enough to not incur in odd results like you'd have with too small cutoff.

timmysilv (approved):

PR #464, opened 2024-08-16
fix type hint for python 3.9

timmysilv:
you can't type-hint with `|` in python 3.9, so I'm righting my wrong

### DIFF
--- a/mrmustard/utils/settings.py
+++ b/mrmustard/utils/settings.py
@@ -14,9 +14,9 @@
 
 """A module containing global settings."""
 
-from pathlib import Path
-from typing import Union
+from __future__ import annotations
 
+from pathlib import Path
 from rich import print
 import rich.table
 import numpy as np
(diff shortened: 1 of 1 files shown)


kaspernielsen96 (approved):
tests works again. Thanks!

apchytr:
Can you try 'from __future__ import annotations' instead? I could swear there's a way to make it work with 3.9

mrmustard/utils/settings.py
```diff
@@ -14,6 +14,8 @@
 
 """A module containing global settings."""
 
+from __future__ import annotations
```
apchytr, line 17: Once @kaspernielsen96 verifies that this works we should probably convert all of our `Optional` type hints to `|` and include this import. I can open

kaspernielsen96: It works with these updates on 3.9.

kaspernielsen96 (approved):
Also works with latest updates on 3.9

apchytr (approved):

PR #466, opened 2024-08-19
Updating type hints

apchytr:
Reopening to avoid unwanted commits 

**Context:** The use of `Optional` and `Union` should be replaced with `|` according to [PEP standards](https://peps.python.org/pep-0604/). 

**Description of the Change:** `Optional` -> ` | None`, `Union` -> ` | `, `Tuple` -> `tuple`, etc

### DIFF
--- a/mrmustard/physics/ansatze.py
+++ b/mrmustard/physics/ansatze.py
@@ -20,7 +20,7 @@ from __future__ import annotations
 
 import itertools
 from abc import ABC, abstractmethod
-from typing import Any, Callable, Union, Optional, Sequence
+from typing import Callable, Sequence
 from warnings import warn
 
 import numpy as np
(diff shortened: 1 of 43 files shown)


ziofil (approved):
thanks!

mrmustard/physics/ansatze.py
```diff
@@ -66,7 +66,7 @@ def __init__(self) -> None:
         self._kwargs = {}
 
     @abstractmethod
-    def from_function(cls, fn: Callable, **kwargs: Any) -> Ansatz:
+    def from_function(cls, fn: Callable, **kwargs: any) -> Ansatz:
```
ziofil, line 69: shouldn't `any` be `Any`?

apchytr: Actually yes! You're right. It's not until now that I found out `any` is referring to the function and not the type

ziofil (approved):
great 👍

PR #472, opened 2024-08-22
Sampler

apchytr:
**Context:** As part of measurement devices in lab_dev we require a sampler class. The sampler can be though of as a black box that takes in a state and outputs either a probability distribution or a collection of samples. Here I implement said class along with subclasses for PNR samples and Homodyne samples where the sampler is designed in way to easily allow for performance shortcuts. Note: some implementation details are subject to change as we continue to develop measurement devices.

**Description of the Change:** `Sampler`, `PNRSampler`, `HomodyneSampler`.


mrmustard/lab_dev/sampler.py
```diff
@@ -0,0 +1,171 @@
...
+        """
+        self._validate_state(state)
+        if state is not None:
+            dm_state = state.dm()
+            states = [dm_state >> meas_op.dual for meas_op in self.meas_ops]
+            probs = [
+                state.probability if isinstance(state, State) else math.real(state)
+                for state in states
+            ]
+            return probs / sum(probs)
```
aplund, line 91: This will silently re-normalise situations where there is little overlap with the state.  Should a warning or some other message be emitted when sum(probs) isn't close to 1.0?

mrmustard/lab_dev/sampler.py
```diff
@@ -0,0 +1,171 @@
...
+
+    def probabilities(self, state: State | None = None):
+        self._validate_state(state)
+        if state is not None:
+            disjoint_modes = [mode for mode in state.modes if mode not in self.meas_ops.modes]
+            dm_state = state.dm() >> TraceOut(disjoint_modes) if disjoint_modes else state.dm()
+            q_state = dm_state >> self.meas_ops
+            z = [[x] * q_state.representation.ansatz.num_vars for x in self.meas_outcomes]
+            probs = math.real(q_state.representation(z)) * math.sqrt(settings.HBAR)
+            return probs / sum(probs)
```
aplund, line 170: Like above it could be good to validate a valid probability distribution.  This would require here multiplying the probability density by the differential (dx), but that should be possible as you have the whole list of points and the differential could be estimated by the difference between points.

Also, is `math.real` required here?  Shouldn't that be part of the `representation` code?

aplund (commented):
Just looking at this code in isolation, I'm not sure how it's all meant to plug into the existing structure.  So It's not really an issue here, but I'm just a bit unclear where this is meant to live.  Is it meant to be attached to measurements? Will the sampler class be closely attached to measurements?

mrmustard/lab_dev/sampler.py
```diff
@@ -0,0 +1,171 @@
...
+    Args:
+        meas_outcomes: The measurement outcomes for this sampler.
+        meas_ops: The optional measurement operators of this sampler.
+        prob_dist: An optional probability distribution for this sampler.
+    """
+
+    def __init__(
+        self,
+        meas_outcomes: list[any],
+        meas_ops: CircuitComponent | list[CircuitComponent] | None = None,
```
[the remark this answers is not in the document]
apchytr: Can definitely be a Sequence 👍

apchytr:
> Just looking at this code in isolation, I'm not sure how it's all meant to plug into the existing structure. So It's not really an issue here, but I'm just a bit unclear where this is meant to live. Is it meant to be attached to measurements? Will the sampler class be closely attached to measurements?

Yes, exactly. Within the bigger picture of Mr Mustard the sampler lives inside measurement devices. I have a prototype branch [here](https://github.com/XanaduAI/MrMustard/compare/prototypeMeas) if you want to get a sense of what it might look like.

mrmustard/lab_dev/sampler.py
apchytr: Yeah I agree that there should be something for validation. I had some form of it before but was getting annoying rounding errors so I ended up just doing the lazy `/ sum(probs)`. I'll play around with it some more to see if I can get something working .

Unfortunately, the representation gives complex values so I think the `math.real` is required (or at least nice to have).

aplund: Perhaps this is a case of having to specify that you want this re-normalisation though an argument.  Lots of weird behaviour can be expected from probability distributions that are conditional (i.e. here conditional on the events that you keep).  Looking at the numpy source for `choice`, the total probability requirements are very strict.  The probabilities must sum to 1.0 within the sqrt of machine epsilon for the float type being used.  If you do adopt an argument, it might be good to accept a custom absolute tolerance as we generally accept distributions that are within 1e-4 of 1.0.

apchytr: I've managed to get a probability validation method on the Sampler. Turns out some of the probability computations I was doing was neglecting terms in the multimode case causing it to fail. I've updated PNRSampler working on the Homodyne now

mrmustard/lab_dev/sampler.py
```diff
@@ -0,0 +1,231 @@
...
+
+    def __init__(
+        self,
+        meas_outcomes: Sequence[any],
+        meas_ops: CircuitComponent | Sequence[CircuitComponent],
+        prob_dist: Sequence[float] | None = None,
+    ):
+        self._meas_ops = meas_ops
+        self._meas_outcomes = meas_outcomes
+        self._prob_dist = prob_dist
```
aplund, line 53: I'm just a little unclear to the role of the `prob_dist` attribute.  I thought it might be used as a cache, but it seems to never be set again after this point.  Shouldn't it be updated when `probabilities()` is called?  Again, I'm just guessing here as to it's role.  Should it only be set on construction?

aplund (commented):
This seems fine.  Can you just clarify the role of `prob_dist`?

mrmustard/lab_dev/sampler.py
apchytr: Hmm I dont believe caching the computed probability distribution makes sense as its state dependent so it would only really save us resources if youre repeatedly calling on the same state. 

The idea was if you have a prob dist already you could just set it (e.g for debugging purposes). I think really for this PR I'll remove it in the name of YAGNI and if theres a use case we can reintroduce it

zeyueN:
@CodiumAI-Agent /update_changelog

zeyueN:
@CodiumAI-Agent /describe
--pr_description.extra_instructions="
For the title, use the format [type]: [summary]
"
--pr_description.publish_description_as_comment=true
--pr_description.generate_ai_title=true
--pr_description.final_update_message=false

mrmustard/lab_dev/sampler.py
```diff
@@ -0,0 +1,190 @@
...
+__all__ = ["Sampler", "PNRSampler", "HomodyneSampler"]
+
+
+class Sampler(ABC):
+    r"""
+    A sampler for measurements of quantum circuits.
+
+    Args:
+        meas_outcomes: The measurement outcomes for this sampler.
+        povms: The POVMs of this sampler.
```
elib20, line 43: `meas_outcomes` are just labels for the `povms` right? But since `povms` are instances of `CircuitComponent`, they already each have a `name` attribute. In the future, should we just rely on that attribute to label the outcomes?

mrmustard/lab_dev/sampler.py
```diff
@@ -0,0 +1,190 @@
...
+        """
+        return self._meas_outcomes
+
+    @abstractmethod
+    def probabilities(self, state: State, atol: float = 1e-4) -> Sequence[float]:
+        r"""
+        Returns the probability distribution of a state w.r.t. measurement outcomes.
+
+        Args:
+            state: The state to generate the probability distribution of.
```
elib20, line 74: Does the input state need to be normalized? If so, good to specify

mrmustard/lab_dev/sampler.py
```diff
@@ -0,0 +1,190 @@
...
+            atol: The absolute tolerance used for validating the computed probability
+                distribution.
```
elib20, line 76: validating = probabilities sum to 1? Perhaps specify probabilities need to sum to 1 p/m atol. 

This is also related to my question about the state needing to be normalized.

mrmustard/lab_dev/sampler.py
```diff
@@ -0,0 +1,190 @@
...
+            seed: An optional seed for random sampling.
+        """
+        initial_mode = state.modes[0]
+        initial_samples = self.sample_prob_dist(state[initial_mode], n_samples, seed)
+
+        if len(state.modes) == 1:
+            return initial_samples
+
+        unique_samples, counts = np.unique(initial_samples, return_counts=True)
+        ret = []
```
elib20, line 95: more descriptive variable name

mrmustard/lab_dev/sampler.py
```diff
@@ -0,0 +1,190 @@
...
+            samples = self.sample(reduced_state, counts)
+            for sample in samples:
+                ret.append(np.append([unique_sample], sample))
+        return np.array(ret)
+
+    def sample_prob_dist(
+        self, state: State, n_samples: int = 1000, seed: int | None = None
+    ) -> np.ndarray:
+        r"""
+        Samples a a state by computing the probability distribution.
```
elib20, line 108:
```suggestion
        Samples a state by computing the probability distribution.
```

mrmustard/lab_dev/sampler.py
```diff
@@ -0,0 +1,190 @@
...
+    Args:
+        cutoff: The photon number cutoff.
+    """
+
+    def __init__(self, cutoff: int) -> None:
+        super().__init__(list(range(cutoff)), [Number([0], n) for n in range(cutoff)])
+        self._cutoff = cutoff
+
+    def probabilities(self, state, atol=1e-4):
+        fock_state = state.dm().to_fock(self._cutoff)
```
elib20, line 154: If the state is pure, we should leverage that. I.e. compute the ket in the fock basis then take the mod squared of the probability amplitudes.

mrmustard/lab_dev/sampler.py
```diff
@@ -0,0 +1,190 @@
...
+
+    def _validate_probs(self, probs: Sequence[float], dx: float, atol: float) -> Sequence[float]:
+        r"""
+        Validates that the given probability distribution sums to `1.0` within some
+        tolerance and returns a renormalized probability distribution to account for
+        small numerical errors.
+
+        Args:
+            probs: The probability distribution to validate.
+            dx: The uniform differential for the probability distribution.
```
elib20, line 130: I think this should be part of the probabilities method. This method should only accept probabilities, i.e. after the probability density has been multiplied by the differential element

mrmustard/lab_dev/sampler.py
```diff
@@ -0,0 +1,190 @@
...
+            atol: The absolute tolerance to validate with.
+        """
+        atol = atol or settings.ATOL
+        probs_dx = probs * dx
+        prob_sum = sum(probs_dx)
```
elib20, line 135:
```suggestion
        prob_sum = sum(probs)
```

mrmustard/lab_dev/sampler.py
elib20:
```suggestion
```

mrmustard/lab_dev/sampler.py
```diff
@@ -0,0 +1,190 @@
...
+        super().__init__(
+            list(meas_outcomes), [QuadratureEigenstate([0], x=x, phi=phi) for x in meas_outcomes]
+        )
+        self._step = step
+
+    def probabilities(self, state, atol=1e-4):
+        q_state = state.dm() >> BtoQ(state.modes, phi=self.povms[0].phi.value[0])
+        z = [x * 2 for x in product(self.meas_outcomes, repeat=len(state.modes))]
+        probs = q_state.representation(z) * math.sqrt(settings.HBAR)
+        return self._validate_probs(probs, self._step ** len(state.modes), atol)
```
elib20, line 190:
```suggestion
        probs = q_state.representation(z) * math.sqrt(settings.HBAR) * self._step ** len(state.modes)
        return self._validate_probs(probs, atol)
```

mrmustard/lab_dev/sampler.py
```diff
@@ -0,0 +1,190 @@
...
+            seed: An optional seed for random sampling.
+        """
+        rng = np.random.default_rng(seed) if seed else settings.rng
+        return rng.choice(
+            a=list(product(self.meas_outcomes, repeat=len(state.modes))),
+            p=self.probabilities(state),
+            size=n_samples,
+        )
+
+    def _validate_probs(self, probs: Sequence[float], dx: float, atol: float) -> Sequence[float]:
```
elib20, line 122:
```suggestion
    def _validate_probs(self, probs: Sequence[float], atol: float) -> Sequence[float]:
```

mrmustard/lab_dev/sampler.py
```diff
@@ -0,0 +1,190 @@
...
+
+    def probabilities(self, state, atol=1e-4):
+        fock_state = state.dm().to_fock(self._cutoff)
+        probs = math.astensor(
+            [
+                fock_state.representation.data[0][(ns * 2)]
+                for ns in product(self.meas_outcomes, repeat=len(state.modes))
+            ]
+        )
+        return self._validate_probs(probs, 1, atol)
```
elib20, line 161:
```suggestion
        return self._validate_probs(probs, atol)
```

mrmustard/lab_dev/sampler.py
```diff
@@ -0,0 +1,190 @@
...
+        num: int = 1000,
+    ) -> None:
+        meas_outcomes, step = np.linspace(*bounds, num, retstep=True)
+        super().__init__(
+            list(meas_outcomes), [QuadratureEigenstate([0], x=x, phi=phi) for x in meas_outcomes]
+        )
+        self._step = step
+
+    def probabilities(self, state, atol=1e-4):
+        q_state = state.dm() >> BtoQ(state.modes, phi=self.povms[0].phi.value[0])
```
elib20, line 187: states now have a convenience method called `quadrature_distribution` that should make this slightly more compact

mrmustard/lab_dev/sampler.py
elib20: We should consider making a convenience method called `fock_distribution` for states, analogous to `quadrature_distribution`

tests/test_lab_dev/test_sampler.py
```diff
@@ -0,0 +1,116 @@
...
+        state = Coherent([0], x=[0.1])
+        samples = sampler.sample(state, n_samples)
+
+        count = np.zeros_like(sampler.meas_outcomes)
+        for sample in samples:
+            idx = sampler.meas_outcomes.index(sample)
+            count[idx] += 1
+        probs = count / n_samples
+
+        assert np.allclose(probs, sampler.probabilities(state), atol=1e-2)
```
elib20, line 69: Could a test like this be flaky in the future? How was atol=1e-2 chosen?

tests/test_lab_dev/test_sampler.py
```diff
@@ -0,0 +1,116 @@
...
+        assert math.allclose(sampler.meas_outcomes, list(np.linspace(-5, 5, 100)))
+
+    def test_probabilties(self):
+        sampler = HomodyneSampler()
+
+        state = Coherent([0], x=[0.1])
+        exp_probs = [
+            (state.dm() >> QuadratureEigenstate([0], x).dual)
+            * sampler._step  # pylint: disable=protected-access
+            for x in sampler.meas_outcomes
```
elib20, line 91: I would compare against the convenience method `quadrature_distribution`

tests/test_lab_dev/test_sampler.py
```diff
@@ -0,0 +1,116 @@
...
+        state = Coherent([0], x=[0.1])
+        samples = sampler.sample(state, n_samples)
+
+        count = np.zeros_like(sampler.meas_outcomes)
+        for sample in samples:
+            idx = sampler.meas_outcomes.index(sample)
+            count[idx] += 1
+        probs = count / n_samples
+
+        assert np.allclose(probs, sampler.probabilities(state), atol=1e-2)
```
elib20, line 116: Feel free to draw some inspiration from the old strawberryfields tests for homodyne, like checking the mean and variance of the probability distribution match what we expect for Gaussian states. https://github.com/XanaduAI/strawberryfields/blob/master/tests/backend/test_homodyne.py

elib20 (changes requested):
Nice work Anthony! I have left mostly minor comments and suggestions. Your code was very clean and easy to follow :)

mrmustard/lab_dev/sampler.py
apchytr: I would prefer not to rely on name particularly because relying on strings can be quite error prone. I also make a lot of use of this structure in the follow up measurement devices PR so it actually ends up being quite useful.

mrmustard/lab_dev/sampler.py
apchytr: My understanding is that `ret` is pretty commonly used to refer to `returned`. We actually have a lot of instances of it throughout Mr Mustard but I could rename to `returned` or `result` if that helps?

tests/test_lab_dev/test_sampler.py
apchytr: Yeah I just selected the highest passing tolerance so I think this test needs to be updated to be a bit more robust 👍

mrmustard/lab_dev/sampler.py
elib20: ok in that case, maybe provide a slightly more descriptive docstring for what is returned, e.g. what shape will the array have etc.

mrmustard/lab_dev/sampler.py
```diff
@@ -0,0 +1,192 @@
...
+    def __init__(
+        self,
+        meas_outcomes: Sequence[Any],
+        povms: CircuitComponent | Sequence[CircuitComponent],
+    ):
+        self._povms = povms
+        self._meas_outcomes = meas_outcomes
+
+    @property
+    def povms(self) -> CircuitComponent | Sequence[CircuitComponent]:
```
ziofil, line 55: Not for now, but we could make a type class for a POVM element which is the dual of a DM.

mrmustard/lab_dev/sampler.py
```diff
@@ -0,0 +1,192 @@
...
+    Args:
+        meas_outcomes: The measurement outcomes for this sampler.
+        povms: The POVMs of this sampler.
+    """
+
+    def __init__(
+        self,
+        meas_outcomes: Sequence[Any],
+        povms: CircuitComponent | Sequence[CircuitComponent],
+    ):
```
ziofil, line 50: Perhaps we should check that `len(meas_outcomes)` is the same as `len(povms)` (or `len(povms) - 1` in case we let the Sampler complete the POVM).

mrmustard/lab_dev/sampler.py
```diff
@@ -0,0 +1,192 @@
...
+
+        if len(state.modes) == 1:
+            return initial_samples
+
+        unique_samples, counts = np.unique(initial_samples, return_counts=True)
+        ret = []
+        for unique_sample, counts in zip(unique_samples, counts):
+            meas_op = self.povms[self.meas_outcomes.index(unique_sample)].on([initial_mode])
+            reduced_state = (state >> meas_op.dual).normalize()
+            samples = self.sample(reduced_state, counts)
```
ziofil, line 102: Nice recursive implementation :)

mrmustard/lab_dev/sampler.py
```diff
@@ -0,0 +1,192 @@
...
+class PNRSampler(Sampler):
+    r"""
+    A sampler for photon-number resolving (PNR) detectors.
+
+    Args:
+        cutoff: The photon number cutoff.
+    """
+
+    def __init__(self, cutoff: int) -> None:
+        super().__init__(list(range(cutoff)), [Number([0], n) for n in range(cutoff)])
```
ziofil, line 151: shouldn't it be `Number([0], n).dual`?

ziofil: In any case, this seems inefficient (albeit correct). There's no need to create the POVM elements and later take inner products with them. Of course it gives the correct results, but since we know the measurement basis is just Fock states can't we bypass this mechanism?

mrmustard/lab_dev/sampler.py
```diff
@@ -0,0 +1,192 @@
...
+        phi: The quadrature angle where ``0`` corresponds to ``x`` and ``\pi/2`` to ``p``.
+        bounds: The range of values to discretize over.
+        num: The number of points to discretize over.
+    """
+
+    def __init__(
+        self,
+        phi: float = 0,
+        bounds: tuple[float, float] = (-10, 10),
+        num: int = 1000,
```
ziofil, line 179: This too seems unnecessary. As for the PNR, it will work, but this should perhaps be set up as a continuous sampler. Maybe it's worth splitting the base sampler into a continuous and a discrete one. The discrete can be used for PNRs (and whatever discrete POVM a user wants to instantiate) and the continuous can be used for Homodyne Heterodyne etc..

tests/test_lab_dev/test_sampler.py
apchytr: Ported those tests 😄

mrmustard/lab_dev/sampler.py
apchytr: Yeah I was thinking the same. Just couldnt think of a use case yet for a dedicated class

mrmustard/lab_dev/samplers.py
```diff
@@ -0,0 +1,207 @@
...
+        initial_samples = self.sample_prob_dist(state[initial_mode], n_samples, seed)
+
+        if len(state.modes) == 1:
+            return initial_samples
+
+        unique_samples, counts = np.unique(initial_samples, return_counts=True)
+        ret = []
+        for unique_sample, counts in zip(unique_samples, counts):
+            meas_op = self._get_povm(unique_sample, initial_mode).dual
+            reduced_state = (state >> meas_op).normalize()
```
ziofil, line 103: I think this line can be more efficient in subclasses. Here it computes an inner product between the state and the POVM element and then calls `normalize` which computes more inner products. For a Homodyne measurement this is equivalent to calling the Ansatz in the Q basis with the variable value equal to the measurement outcome, and then dividing by the probability of the outcome (both of which are already known at this point). For a PNR measurement we have to go though the polynomial, but still it must be more efficient than creating several `Number` states and computing inner products against them.

mrmustard/lab_dev/samplers.py
```diff
@@ -0,0 +1,229 @@
...
+    def probabilities(self, state, atol=1e-4):
+        return self._validate_probs(state.fock_distribution(self._cutoff), atol)
+
+
+class HomodyneSampler(Sampler):
+    r"""
+    A sampler for homodyne measurements.
+
+    Args:
+        phi: The quadrature angle where ``0`` corresponds to ``x`` and ``\pi/2`` to ``p``.
```
ziofil, line 185: Perhaps we should leave this at 0 by default (homodyne always on x quadrature) and let people put an Rgate before to reach any angle that way.

mrmustard/lab_dev/samplers.py
```diff
@@ -0,0 +1,229 @@
...
+            list(meas_outcomes),
+            QuadratureEigenstate([0], x=0, phi=phi),
+        )
+        self._step = step
+        self._outcome_arg = "x"
+
+    def probabilities(self, state, atol=1e-4):
+        probs = state.quadrature_distribution(
+            self.meas_outcomes, self.povms.phi.value[0]
+        ) * self._step ** len(state.modes)
```
ziofil, line 207: I see here these are the probabilities related to measuring all modes, but since we are measuring one mode at a time, it may be simpler than this.

apchytr: Yes, our sample method does for one mode but I think this is a convenient way for users to compute all modes if they want to

mrmustard/lab_dev/samplers.py
```diff
@@ -0,0 +1,229 @@
...
+    def __init__(
+        self,
+        phi: float = 0,
+        bounds: tuple[float, float] = (-10, 10),
+        num: int = 1000,
+    ) -> None:
+        meas_outcomes, step = np.linspace(*bounds, num, retstep=True)
+        super().__init__(
+            list(meas_outcomes),
+            QuadratureEigenstate([0], x=0, phi=phi),
```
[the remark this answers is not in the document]
apchytr: We dont just left over from before 😅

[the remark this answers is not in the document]
apchytr: Correction because I forgot: Its a convenient way of keeping track of phi. An alternate is to just have a `_phi` attr

mrmustard/lab_dev/samplers.py
ziofil: yeah I take it back, it's probably simpler this way.

PR #473, opened 2024-08-23
Autoshape poly

kaspernielsen96:
**Context:**
Update autoshape for `ket` and `dm` if they have a polynomial.
Adding a check to see if the bargmann representation has a polynomial and set the `autoshape_max` as shape if that is the case.


ziofil (approved):
Makes sense. Do you think there's a way to guesstimate a better shape than falling back to the default max?

PR #474, opened 2024-08-27
DM probabilities bug fix

apchytr:
**Context:** Currently on develop we see the following
```
(mrml.Vacuum([0,1]) >> mrml.Number([0], 1).dual).probability
>> 0.0

(mrml.Vacuum([0,1]) >> mrml.Number([0], 1).dual).dm().probability
>> 1.0
```
thanks to @kaspernielsen96 for pointing out that it's due to the polynomial wires.

**Description of the Change:** `Bargmann.scalar` now handles the logic for polynomial wires. As a result, the above no longer occurs. Additionally, `.dm()` on `DM` and cleanup reordering `states/base.py`.


kaspernielsen96 (approved):
Great!

PR #475, opened 2024-08-27
Attempt at nuking tensor networks and lab

ziofil:
This is just for me to see what breaks and where. You can safely ignore.


ziofil:
too early

PR #478, opened 2024-08-30
freeze settings to avoid typos

timmysilv:
**Context:**
In the past, people have accidentally and unknowingly set non-existent settings then expected their changes to be respected. This is often a simple typo, but MrMustard wouldn't correct you.

- Freeze the set of available settings to be set once the `Settings` class has been set in stone (initialized).
- `settings.HBAR` no longer locks, but you get a warning _every time_ you change it
- Improve the settings context manager so all keys are allowed

**Benefits:**
No more typos or setting outdated settings by name. Eg. if we one day replace `HBAR` with `H_BAR`, and you try to set `settings.HBAR = 1.0`, you will get an attribute error (this is good because your `settings.HBAR` value wouldn't be used anywhere).

**Possible Drawbacks:**
N/A


mrmustard/utils/settings.py
```diff
@@ -122,6 +124,14 @@ def __init__(self):
         self.ATOL: float = 1e-8
         "The absolute tolerance when comparing two values or arrays. Default is 1e-8."
 
+        self._frozen = True
```
apchytr, line 127: When would we have frozen be false? And does that mean then the user could populate with their own settings?

mrmustard/utils/settings.py
```diff
@@ -122,6 +124,14 @@ def __init__(self):
...
+
+    def __setattr__(self, name, value):
+        """Once the class is initialized, do not allow the addition of new settings."""
+        if self._frozen and name != "HBAR" and not hasattr(self, name):
```
apchytr, line 131: Gah maybe instead of locking HBAR we just warn users that when they change it unexpected behaviour can occur?

mrmustard/utils/settings.py
timmysilv: it's defined as `False` at the class level, then while `__init__` happens and sets all the default values, there's no issue. but at the very end of `__init__` (here), we set it to True forevermore. I've seen this before, I think it's the semi-standard way of locking a class

mrmustard/utils/settings.py
timmysilv: my goal was to not mess with how things were in this PR, but I'm open to discussing new plans for HBAR. I'm guessing the lock was added because people were accidentally doing sketchy math

elib20: I think one reason was that some of functions for plotting states use cached arrays that depend on the value of hbar

apchytr: Yeah, the way the HBAR lock is setup right now kind of conflicts with the settings lock hence requiring this attribute specific logic (which I'd like to avoid as much as possible). I don't see an easy way to fix this which is why I suggested a warning instead. I'd like to say that just telling a user "warning: changing this setting may conflict with any prior computations" is sufficient but I'm not too sure

timmysilv: I'm fine with a warning or an error, but this explicit HBAR check is required either way to ensure the warning/error doesn't appear on access, right?

apchytr: Not necessarily? Like if we do 
```suggestion
    @property
    def HBAR(self):
        r"""The value of the Planck constant. Default is ``2``.
        Cannot be changed after its value is queried for the first time.
        """
        return self._hbar
```
```suggestion
    @HBAR.setter
    def HBAR(self, value: float):
        warnings.warn("Changing HBAR can conflict with prior computations.") 
        self._hbar = value
```
then we only get warnings when someone changes HBAR. (Tested locally)

timmysilv: oh I see. I was assuming we want to avoid the warning if it's their first time changing it, but I kinda agree that this warning is fine if @elib20 is ok with it

timmysilv: ok i've updated it to just raise a warning whenever set

tests/test_lab_dev/test_states/test_quadrature_eigenstate.py
```diff
@@ -60,19 +60,16 @@ def test_init_error(self):
...
 
-        settings._hbar_locked = False
         settings.HBAR = hbar
         q1 = QuadratureEigenstate([0], x=0, phi=0)
         assert np.allclose(q0.bargmann_triple()[0], q1.bargmann_triple()[0])
         assert np.allclose(q0.bargmann_triple()[1], q1.bargmann_triple()[1])
         assert np.allclose(q0.bargmann_triple()[2], q1.bargmann_triple()[2])
 
-        settings._hbar_locked = False
         settings.HBAR = 2.0
```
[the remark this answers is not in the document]
timmysilv: this test certainly was bound to cause problems. I updated the context manager to always work regardless of custom keys, and use it to modify HBAR temporarily now

apchytr (approved):
Thanks looks great!

tests/test_lab_dev/test_states/test_quadrature_eigenstate.py
```diff
@@ -60,20 +60,15 @@ def test_init_error(self):
 
     @pytest.mark.parametrize("hbar", hbar)
     def test_probability_hbar(self, hbar):
-        settings._hbar_locked = False
         settings.HBAR = 2.0
```
timmysilv, line 63:
```suggestion
```

apchytr (approved):

PR #486, opened 2024-09-06
to_bargmann() original bargmann data bug

apchytr:
**Context:** currently on develop `to_bargmann` is not checking if original bargmann data exists leading to issues such as 
```
state = SqueezedVacuum([0], r=1.0)
dgate = Dgate([0], x=10.0).to_fock(150)
state >> dgate
```
running out of memory. 

**Description of the Change:** `to_bargmann` now checks for original bargmann data. Removed a single line that has `original_fock_data` as that is not used anywhere. Moved some things to ansatz. `__truediv__` and `__mul__` can update the original bargmann data as long as other is a scalar. `_original_bargmann_data` is renamed to `_original_abc_data`. Some `Representation` cleanup removing duplicate code.

### DIFF
--- a/mrmustard/lab_dev/circuit_components.py
+++ b/mrmustard/lab_dev/circuit_components.py
@@ -607,9 +607,12 @@ class CircuitComponent:
         if isinstance(self.representation, Bargmann):
             return self
         else:
-            A, b, _ = identity_Abc(len(self.wires.quantum))
-            bargmann = Bargmann(A, b, self.representation.data)
-            bargmann._original_fock_data = self.representation.data
+            if self.representation._original_bargmann_data:
+                A, b, c = self.representation._original_bargmann_data
+            else:
+                A, b, _ = identity_Abc(len(self.wires.quantum))
+                c = self.representation.data
+            bargmann = Bargmann(A, b, c)
             try:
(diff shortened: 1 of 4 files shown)


timmysilv:
looks good! I'm asking this more for personal clarification than anything else: what value do we gain from this separation of representations and ansatze? Is it in anticipation for future things? I'm wondering this because I see that this separation is the reason you lowered the `_original` data to the ansatz itself.

apchytr:
> looks good! I'm asking this more for personal clarification than anything else: what value do we gain from this separation of representations and ansatze? Is it in anticipation for future things? I'm wondering this because I see that this separation is the reason you lowered the `_original` data to the ansatz itself.

So I think the original design was that we would have different types of ansatz but I believe as we continue to develop lab_dev its becoming more and more evident that this separation is unnecessary. The only thing I could think of is that maybe codefactor will give us issues if we merge the two.

timmysilv (approved):
looks good to me! I think someone who has more informed feelings about the separation should confirm the lowering of `triple` not being a problem but this gets the job done imo

elib20 (approved):
We should decide if we need both representation and ansatz. The line is getting blurred here.

PR #500, opened 2024-10-10
c of BtoPS

arsalan-motamedi:
**Context:**
We would like `BtoPS` to compute the characteristic function, but it currently computes it with a `1/2pi` prefactor. 

Changing the `c` of `BtoPS` from `1/(2*pi)` back to `1` so that we can readily compute characteristic function (and propagating to tests).

**Benefits:**
Minor (we can compute characteristic function directly by BtoPS)

**Possible Drawbacks:**
Need to be ware of this change when computing Wigner function (the 2pi factor would be placed there -- look at `test_cft.py` to see how to do this).

**Related GitHub Issues:**
None.


ziofil (approved):
Thanks! I can't recall why I put those there... Let's hope it was just a brain fart 😂

apchytr (approved):

PR #512, opened 2024-10-18
Bump `rich` dependency to version 13

aplund:
**Context:** Old dependencies can hold back things which depend on the MrMustard, which can be very deep in the dependency tree.  The `rich` package had a very old major version dependency (10.*.*) and this could interact with other seemingly random packages when depending on MrMustard.

**Description of the Change:** Bump the major version of `rich` to 13. The only place I can see to test this version change is in the ProgressBar class, and this seemed to work fine for me on testing using the example code in the docs for using the optimizer.

**Benefits:** Other packages that depend on MrMustard won't get into a tricky dependency situation with a very old package.

**Possible Drawbacks:** Possible subtle bugs as this dependency doesn't appear to have unit tests.


apchytr (approved):

timmysilv (approved):
scipy & sphinx bumps scare me a bit but i think it's ok 🙂

aplund:
The issue with the code coverage checks seems to be changes in how docstrings and decorators are handled between versions of the "coverage" package.  This changes the lock file and moves from coverage version 7.4.4 to 7.6.3.  In the earlier version, docstrings and decorator executions are included, and the later version they are not.  So there's a heap of stuff that appears to be less coverage, but these things aren't actual code in the codebase.

(I can't merge this without all the checks being successful, including the code coverage ones.)

timmysilv:
i can't even find where/if this is configured. maybe there's a default value somewhere saying the max coverage delta must be <=0.15% or something like that?

apchytr:
@timmysilv I think it was using code coverage's default values. I've added a .yml file that we can configure to (hopefully) fix this.

apchytr (approved):

PR #516, opened 2024-10-30
skip progressbar if needed

ziofil:
**Context:**
when launching jobs on the command line, using `settings.PROGRESSBAR = False` is not enough because it's just invisible, but some output lines are still skipped.

Entirely skips creating the `ProgressBar` object

**Benefits:**
Slightly faster optimizations with no progress bar, cleaner console output

### DIFF
--- a/mrmustard/training/optimizer.py
+++ b/mrmustard/training/optimizer.py
@@ -18,7 +18,7 @@ used within Mr Mustard.
 
 from itertools import chain, groupby
 from typing import List, Callable, Sequence, Union, Mapping, Dict
-from mrmustard import math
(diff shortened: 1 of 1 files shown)


mrmustard/training/optimizer.py
```diff
@@ -98,35 +98,58 @@ def minimize(
     def _minimize(self, cost_fn, by_optimizing, max_steps, callbacks):
         # finding out which parameters are trainable from the ops
         trainable_params = self._get_trainable_params(by_optimizing)
+        bar = ProgressBar(max_steps)
```
apchytr, line 101: Should this be under line 102?

PR #523, opened 2024-11-14
Rich wires

ziofil:
Wires have been challenging to maintain due to evolving requirements.

This PR introduces a better abstraction for Wire that stores a collection of `QuantumWire` and `ClassicalWire` dataclasses. Metadata can be added to these dataclasses without modifying the Wires object’s handling of indices and dictionaries. Many methods, like `__matmul__`, are greatly simplified.


mrmustard/lab_dev/circuit_components.py
```diff
@@ -16,7 +16,7 @@
 A base class for the components of quantum circuits.
 """
 
-# pylint: disable=super-init-not-called, import-outside-toplevel
+# pylint: disable=super-init-not-called, import-outside-toplevel, too-many-positional-arguments
```
apchytr, line 19: Out of curiosity which line is raising this pylint issue?

ziofil: pylint started to complain about any method that needs more than 5 args, and it counts also `self` as one, so the init raises this issue

mrmustard/lab_dev/states/dm.py
```diff
@@ -130,22 +130,26 @@ def from_fock(
     @classmethod
     def from_ansatz(
         cls,
-        modes: Sequence[int],
+        modes: Iterable[int],
```
apchytr, line 133: Shouldn't it be Sequence? Based on [this stack overflow post](https://stackoverflow.com/questions/72157296/what-is-the-difference-between-type-hinting-a-variable-as-an-iterable-versus-a-s) a sequence is "any object that defines `__getitem__` and `__len__`" and we directly depend on `len` below.

mrmustard/lab_dev/states/dm.py
```diff
@@ -130,22 +130,26 @@ def from_fock(
...
         ansatz: PolyExpAnsatz | ArrayAnsatz | None = None,
         name: str | None = None,
     ) -> State:
+        if not isinstance(modes, set) and sorted(modes) != list(modes):
+            raise ValueError(f"Modes must be sorted. got {modes}")
```
apchytr, line 138: Seeing as we are already sorting the modes to do the check does it make sense to just use that?

mrmustard/lab_dev/states/ket.py
```diff
@@ -113,12 +113,16 @@ def from_ansatz(
...
+        if not isinstance(modes, set) and sorted(modes) != list(modes):
+            raise ValueError(f"Modes must be sorted. Got {modes}")
         modes = set(modes)
         if ansatz and ansatz.num_vars != len(modes):
             raise ValueError(
                 f"Expected an ansatz with {len(modes)} variables, found {ansatz.num_vars}."
             )
         wires = Wires(modes_out_ket=modes)
+        if isinstance(ansatz, ArrayAnsatz):
+            wires = wires.to_fock(ansatz.array.shape)
```
apchytr, line 125: Not sure how I feel about the idea of passing information related to the ansatz (i.e fock shape) into the wires. What is the motivation for it?

mrmustard/lab_dev/states/dm.py
ziofil: I thought iterable was more appropriate given that modes are assumed to be sorted. E.g. iterables like a set don't necessarily have an order.

mrmustard/lab_dev/states/dm.py
ziofil: it was to avoid users inputting modes with a particular order and expecting DM to retain it

mrmustard/lab_dev/states/ket.py
ziofil: It was so that when you need to change representation you have the arguments for BtoFoo.inverse()

mrmustard/lab_dev/states/dm.py
ziofil: Anyway, i don't feel strongly about it. It can be a sequence.

mrmustard/lab_dev/states/dm.py
apchytr: We could either or alternatively we could throw a warning. We could even just document "modes will be sorted in order".

mrmustard/lab_dev/states/dm.py
apchytr: Actually I think `Collection` is the most appropriate. You're right `set` means it's not necessarily a `Sequence` as it does not implement `__getitem__` but `Collection` covers all bases

mrmustard/physics/wires.py
```diff
@@ -169,242 +310,255 @@ def __init__(
...
+            modes_in_bra=self.output.bra.modes,
+            modes_out_ket=self.input.ket.modes,
+            modes_in_ket=self.output.ket.modes,
+            classical_out=self.input.classical.modes,
+            classical_in=self.output.classical.modes,
+        )
 
-    @cached_property
-    def classical(self) -> Wires:
+    def to_fock(self, shape: tuple[int, ...]) -> Wires:
```
apchytr, line 409: A point in favour of keeping the representation information in `Representation` rather that `Wires`: with this we now need to initialize a new instance of `Wires` every time we do a representation change. Instead, by separating out the responsibility into the `Representation` class a representation change is simply updating the value in a dictionary. For this reason (as well as conceptually) I think we should keep that logic in the `Representation` (and maybe even introduce a dataclass to wrap `repr`, `repr_params`)

mrmustard/physics/wires.py
```diff
@@ -169,242 +310,255 @@ def __init__(
...
-    ) -> None:
-        self.args: tuple[set, ...] = (
-            modes_out_bra or set(),
-            modes_in_bra or set(),
-            modes_out_ket or set(),
-            modes_in_ket or set(),
-            classical_out or set(),
-            classical_in or set(),
+    ):
+        if any(
```
apchytr, line 314: I don't think we need this check since the type hints already make it explicit that's what we expect

mrmustard/lab_dev/circuits.py
```diff
@@ -287,7 +287,7 @@ def serialize(self, filestem: str = None):
     @classmethod
     def deserialize(cls, data: dict) -> Circuit:
         r"""Deserialize a Circuit."""
-        comps, path = map(data.pop, ("components", "path"))
+        comps, path = data.pop("components"), data.pop("path")
```
aplund, line 290: I don't understand why this change in needed.  Aren't the two equivalent.  The changed one is probably more readable.

mrmustard/physics/wires.py
```diff
@@ -1,31 +1,169 @@
...
+class LegibleEnum(Enum):
+    """Enum class that provides a more legible string representation."""
 
-from mrmustard import widgets
+    def __str__(self) -> str:
+        return self.name
 
-__all__ = ["Wires"]
+    def __repr__(self) -> str:
+        return self.name
```
aplund, line 29: I'm not sure about the need for this in the repr.  The default str is probably not too bad either as it's just the enum with the instance name.

mrmustard/physics/wires.py
```diff
@@ -1,31 +1,169 @@
...
+            repr_params=self.repr_params,
+            id=self.id if not new_id else randint(0, 2**32 - 1),
+        )
+
+    def _order(self) -> int:
+        """
+        Artificial ordering for sorting quantum wires.
+        Order achieved is by bra/ket, then out/in, then mode.
+        """
+        return self.mode + 10_000 * (1 - 2 * self.is_out) - 100_000 * (1 - 2 * self.is_ket)
```
aplund, line 115: This seems to be used as the sorting key.  Perhaps using lex ordering of tuples would be more readable?

mrmustard/physics/wires.py
```diff
@@ -1,31 +1,169 @@
...
+    repr_params: Any = None
+    id: int = field(default_factory=lambda: randint(0, 2**32 - 1), compare=False)
+
+    def __hash__(self) -> int:
+        return hash((self.mode, self.is_out, self.is_ket, self.repr))
+
+    def __repr__(self) -> str:
+        return f"QuantumWire(mode={self.mode}, {'out' if self.is_out else 'in'}, {'ket' if self.is_ket else 'bra'}, repr={self.repr}, index={self.index})"
+
+    def __eq__(self, other: QuantumWire) -> bool:
```
aplund, line 83: Dataclasses have built-in equality and hashing generation.  Could this not be achieved by flagging the appropriate fields?

mrmustard/physics/wires.py
```diff
@@ -1,31 +1,169 @@
...
+        """
+        Artificial ordering for sorting quantum wires.
+        Order achieved is by bra/ket, then out/in, then mode.
+        """
+        return self.mode + 10_000 * (1 - 2 * self.is_out) - 100_000 * (1 - 2 * self.is_ket)
 
 
-class Wires:
+@dataclass(slots=True)
+class ClassicalWire:
```
aplund, line 119: The above comments on QuantumWires applies here too.

mrmustard/physics/wires.py
```diff
@@ -169,242 +310,255 @@ def __init__(
...
+        for i, m in enumerate(sorted(modes_in_ket)):
+            self.quantum_wires.add(QuantumWire(mode=m, is_out=False, is_ket=True, index=n + i))
+        n += len(modes_in_ket)
+        for i, m in enumerate(sorted(classical_out)):
+            self.classical_wires.add(ClassicalWire(mode=m, is_out=True, index=n + i))
+        n += len(classical_out)
+        for i, m in enumerate(sorted(classical_in)):
+            self.classical_wires.add(ClassicalWire(mode=m, is_out=False, index=n + i))
+
+    @classmethod
```
aplund, line 356: Is this a classmethod or a staticmethod?  I can't seem to see a reference to 'cls' in the body.  I think it's static, as I can't think of a case where a subclass would need this.  But if it is, then the first line should probably be 'w = cls()'.

mrmustard/physics/wires.py
ziofil: I agree that it is not necessary to pass the shape for Fock, but I want to push back on where we put the information. I hear you, that we need to rebuild the wires, but that is hardly the bottleneck given all the other things we do in a representation change. And the wires are part of the representation anyway, so afaik it makes more sense to centralize the wire information into the wires. I'll review the way I pass this info into the wires though.

mrmustard/physics/wires.py
```diff
@@ -169,242 +310,255 @@ def __init__(
...
-        return ret
+        return Wires.from_wires(
+            quantum={q for q in self.quantum_wires if q.is_out},
+            classical={c for c in self.classical_wires if c.is_out},
+        )
+
+    ###### PROPERTIES ######
 
     @cached_property
+    def id(self) -> int:
```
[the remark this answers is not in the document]
ziofil: This is for the entire Wires object, but I'll think whether we need it.

mrmustard/lab_dev/circuit_components.py
```diff
@@ -698,9 +699,9 @@ def __rshift__(self, other: CircuitComponent | numbers.Number) -> CircuitCompone
         if only_ket or only_bra or both_sides:
             ret = self @ other
         elif self_needs_bra or self_needs_ket:
-            ret = (self.adjoint @ self) @ other
+            ret = self.adjoint @ (self @ other)
```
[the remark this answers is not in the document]
ziofil: It's not a bug, I just noticed it was more efficient to not do the outer product

mrmustard/lab_dev/circuits.py
ziofil: yep, readability

mrmustard/physics/wires.py
ziofil: I just dislike the mess that precedes the name

mrmustard/physics/wires.py
ziofil: good idea, thanks

mrmustard/physics/wires.py
ziofil: What do you mean? Can you suggest a change?

mrmustard/physics/wires.py
ziofil: I think we need it because otherwise one can pass an unsorted tuple, or list. Like (3,2,1) and then the modes are screwed up. To guard against that we would have to cast to a set, but I'd rather not do that by default because then implicit stuff happens.

mrmustard/physics/wires.py
```diff
@@ -474,76 +657,44 @@ def __matmul__(self, other: Wires) -> tuple[Wires, list[int]]:
...
-
-        This way it is possible to write:
+        Using the permutations, it is possible to write:
 
         .. code-block::
 
             ansatz = ansatz1[idx1] @ ansatz2[idx2]  # not in standard order
-            wires, perm = wires1 @ wires2  # matmul the wires of each component
-            ansatz = ansatz.reorder(perm)  # now in standard order
+            wires, perm_CV, perm_DV = wires1 @ wires2  # matmul the wires
```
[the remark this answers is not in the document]
ziofil: yes, good catch!

mrmustard/physics/wires.py
apchytr: But we type hint it as a set does that mean we should update the type hint?

mrmustard/physics/wires.py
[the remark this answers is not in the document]
ziofil: ok I remved it.

mrmustard/physics/wires.py
ziofil: What change do you have in mind? Here the hint says `set[int] | None` and the method raises if it isn't that.

mrmustard/physics/wires.py
ziofil: Ah (just looked it up) but we also need to set frozen=True, which we don't want.

mrmustard/physics/wires.py
apchytr: Right the type hint says its `set | None` so we shouldn't be expecting something like an unsorted tuple or list to be passed in. If we want to support passing in an unsorted tuple or list then the type hint should be updated to reflect that. If not, then the check shouldn't be necessary as the type hint is already implying it is `set | None`.

ziofil: As I'm sure you know, the type hint does not enforce anything. Raising an exception in this case guards against running into bugs because of an ordered data structure is passed in when out of order. If this is your only concern about this PR can you accept this check and approve it?

apchytr (approved):
Awesome job Filippo!

PR #528, opened 2024-11-27
Python 3.12 version bump

apchytr:
### **User description**
**Context:** Support for python 3.12. This includes bumping the tensorflow version.

**Description of the Change:** Updated the pyproject.toml to support python 3.12 and bumped the tensorflow version. Removed seemingly unnecessary dependencies. Removed macOS warning about switching to tensorflow 2.15.0. Default to using the legacy Adam optimizer.

Note: `test_training` and `test_opt` fail with `tf > 2.15.0`. For now these tests are skipped and are to be fixed in the future. 
___

### **PR Type**
enhancement, dependencies

___

### **Description**
- Added support for Python 3.12 by updating the `classifiers` and `python` version range in `pyproject.toml`.
- Bumped the TensorFlow version to 2.18.0 to ensure compatibility with Python 3.12.
- Simplified TensorFlow dependencies by removing platform-specific configurations, likely due to improved compatibility in newer TensorFlow releases.

___

### **Changes walkthrough** 📝
<table><thead><tr><th></th><th align="left">Relevant files</th></tr></thead><tbody><tr><td><strong>Dependencies</strong></td><td><table>
<tr>
  <td>
    <details>
      <summary><strong>pyproject.toml</strong><dd><code>Update Python and TensorFlow version support in pyproject.toml</code></dd></summary>
<hr>

pyproject.toml

<li>Added support for Python 3.12 in classifiers.<br> <li> Updated Python version range to include 3.12.<br> <li> Bumped TensorFlow version to 2.18.0.<br> <li> Removed detailed TensorFlow platform-specific configurations.<br>

</details>

  </td>
  <td><a href="https://github.com/XanaduAI/MrMustard/pull/528/files#diff-50c86b7ed8ac2cf95bd48334961bf0530cdc77b5a56f852c5c61b89d735fd711">+3/-22</a>&nbsp; &nbsp; </td>

</tr>
</table></td></tr></tr></tbody></table>

___

> 💡 **PR-Agent usage**: Comment `/help "your question"` on any pull request to receive relevant information


tests/conftest.py
```diff
@@ -71,9 +73,21 @@ def set_backend(backend):
...
     if math.backend_name == "numpy":
         pytest.skip("numpy")
 
 
+def skip_tf():
+    r"""
+    Skips tests if the backend is tensorflow and the version is ``>2.15.0``.
+    """
+    if math.backend_name == "tensorflow":
+        if Version(metadata.distribution("tensorflow").version) > Version("2.15.0"):
```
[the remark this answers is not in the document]
apchytr: Right now yes. I'm still debating on having 2.15.0 be the lowest but then you have to jump through some hoops to have `tensorflow-macos` install when `tf == 2.15.0` but not install if `tf>2.15.0`.

mrmustard/math/backend_tensorflow.py
```diff
@@ -413,15 +409,7 @@ def sqrtm(self, tensor: tf.Tensor, dtype, rtol=1e-05, atol=1e-08) -> Tensor:
     # ~~~~~~~~~~~~~~~~~
 
     def DefaultEuclideanOptimizer(self) -> tf.keras.optimizers.legacy.Optimizer:
-        use_legacy = Version(metadata.distribution("tensorflow").version) < Version("2.16.0")
```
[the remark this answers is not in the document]
apchytr: Based on running the tests that work on `>=2.15.0` (meaning actual optimization tests are skipped) they are roughly the same. `0:03:02` for `2.15.0` and `0:03:38` for `2.18.0`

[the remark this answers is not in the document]
apchytr: That is to say, preliminary profiling suggests they're similar but we really wouldn't know until the optimization itself is fixed. Considering `tensorflow-macos` is no longer the suggested installation for macos I think it should be fine

tests/conftest.py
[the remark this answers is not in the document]
timmysilv: hmm. we haven't released since March - @elib20 is there a plan to make another official MrMustard version? I'm wondering how bad it would be to just say "if using MrMustard 7.4+, then you need TF 2.16+"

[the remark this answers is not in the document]
apchytr: I think the next planned release is once lab dev is complete. Seeing as that would be a massive update from the previous release it's completely fine to mention in the CHANGELOG that we just bumped the version

mrmustard/math/backend_tensorflow.py
```diff
@@ -413,14 +410,11 @@ def sqrtm(self, tensor: tf.Tensor, dtype, rtol=1e-05, atol=1e-08) -> Tensor:
...
-        AdamOpt = tf.keras.optimizers.legacy.Adam if use_legacy else tf.keras.optimizers.Adam
-        if not use_legacy and platform.system() == "Darwin" and platform.processor() == "arm":
-            warn(
-                "Mac ARM processor detected - MrMustard always trains using the latest Keras Adam "
-                "optimizer with TensorFlow 2.16+, but it is known to be slow on Mac+ARM. To use "
-                "the legacy optimizer, please downgrade TensorFlow to 2.15."
-            )
+        if platform.system() == "Darwin" and platform.processor() == "arm":
+            os.environ["TF_USE_LEGACY_KERAS"] = "True"
```
timmysilv, line 414: this works with keras 3?

mrmustard/training/optimizer.py
```diff
@@ -87,6 +87,7 @@ def minimize(
                 Learn more about how to use callbacks to have finer control of the optimization
                 process in the :mod:`.callbacks` module.
         """
+        math._euclidean_opt = None  # temporary workaround
```
timmysilv, line 90: how do we ensure this is only temporary?

tests/conftest.py
```diff
@@ -15,9 +15,11 @@
 import os
 from pathlib import Path
 
+from importlib import metadata
 import pytest
 from hypothesis import Verbosity
 from hypothesis import settings as hyp_settings
+from semantic_version import Version
```
timmysilv, line 22:
```suggestion
import pytest
from hypothesis import Verbosity
from hypothesis import settings as hyp_settings
```

mrmustard/math/backend_tensorflow.py
apchytr: It does!

mrmustard/training/optimizer.py
apchytr: I say temporary with the intent of coming back to this to support the non-legacy Adam optimizer. For now it's there to have our pipeline working

apchytr, line 90:
```suggestion
        math._euclidean_opt = None  # TODO: fix this temporary workaround
```

pyproject.toml
```diff
@@ -43,27 +44,9 @@ nbformat = "^5.10"
...
-#     { version = "^2.15.0", platform = "darwin", markers = "platform_machine!='arm64' and platform_machine!='aarch64'" },]
-tensorflow-cpu-aws = { version = "^2.15.0", platform = "linux", markers = "platform_machine=='arm64' or platform_machine=='aarch64'" }
-# https://github.com/tensorflow/tensorflow/blob/adb39b04e9cb116df4659a7e2de9eea27e62f25c/tensorflow/tools/pip_package/setup.py#L107-L108
-# https://github.com/python-poetry/poetry/issues/8271#issuecomment-1697740447
-tensorflow-io-gcs-filesystem = [
-    { version = ">= 0.23.1", markers = "platform_machine!='arm64' or platform_system!='Darwin'" },
-    { version = "< 0.32.0", markers = "platform_system == 'Windows'" }
-]
-tensorflow-probability = { version = ">=0.22.0,<1.0", extras = ["tf"] }
+tensorflow = {version = "^2.15.0, <2.19.0" }
```
timmysilv, line 47: is 2.15 actually supported now with all these changes? wondering if we should add a test suite that manually downgrades to 2.15 and confirms this. i know it's not _Poetic_, but I suspect it's the most likely lower bound dependency that will suddenly fail.

tests/test_training/test_opt.py
```diff
@@ -47,6 +48,9 @@ def test_S2gate_coincidence_prob(n):
     skip_np()
 
     settings.SEED = 40
+    rng = tf.random.get_global_generator()
+    rng.reset_from_seed(settings.SEED)
```
timmysilv, line 52: should this logic go into the `SEED` setter?

timmysilv (commented):
looks good! just 2 minor questions

tests/test_training/test_opt.py
timmysilv: agreed, but that's probably out of scope for this PR so I suppose we can leave it for now.

pyproject.toml
timmysilv: For python versions, it's fairly easy. Just use the `strategy.matrix` property like we do here:
https://github.com/XanaduAI/MrMustard/blob/2a1d7f97526bb36fbbff90b22ee686b001c122c0/.github/workflows/builds.yml#L15-L18
Then pass the `${{ matrix.python-version }}` value to `actions/setup-python`.

Poetry won't help with testing various tensorflow versions 😢 We'd have to work against the grain and manually pip install versions other than the one specific in `poetry.lock`. It's hacky, but can be done.

If we think either/both of these things are valuable, I can add it.

apchytr: I think it's worthwhile to have. My only concern is slowing down our CICD so the parameterizations should run in parallel. For now lets just stick to running over the different versions of python 👍

timmysilv (dismissed):

timmysilv (approved):

PR #521, opened 2024-11-08
Batch Refactor

apchytr:
### **User description**
**Context:** Currently in develop, there is no interface explicitly defined for the handling of batch dims. This makes things such as adding new batch dimensions (e.g. for measurement outcomes) difficult. Additionally, we wish to introduce the concept of "batch labels" where ideally some class is responsible for the handling of batch labels. 

**Description of the Change:** Introduce the `Batch` class. `__getitem__` and `__matmul__` functionality of ansatz has been replaced with `contract`. `__getitem__` on ansatz now slices batch dimensions. Removed `batched` argument on `from_fock` and the init of `ArrayAnsatz` in favour of passing a `Batch` class directly.

**Benefits:** Having a concrete interface means a cleaner, safer implementation of batching where we can now define arbitrary batch dims and attribute batch labels to them (for mrm einsum). 

TODO: 
	- Everything working w arbitrary batch dims
	- Handling of batch labels
	- Improve ufunc as much as possible
___

### **PR Type**
enhancement, tests

___

### **Description**
- Introduced the `Batch` class to handle batch dimensions across various modules.
- Replaced the `@` operator with the `contract` method for ansatz operations.
- Removed the `batched` parameter from methods and classes, simplifying their interfaces.
- Updated tests to accommodate the new `Batch` class and `contract` method.
- Added new tests for the `Batch` class and enhanced existing tests.
- Updated pylint configuration to disable `too-many-positional-arguments` check.

___

### **Changes walkthrough** 📝
<table><thead><tr><th></th><th align="left">Relevant files</th></tr></thead><tbody><tr><td><strong>Enhancement</strong></td><td><details><summary>12 files</summary><table>
<tr>
  <td>
    <details>
      <summary><strong>trace_out.py</strong><dd><code>Use `contract` method for ansatz operations</code>&nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; </dd></summary>
<hr>

mrmustard/lab_dev/circuit_components_utils/trace_out.py

- Replaced `@` operator with `contract` method for ansatz.

</details>

  </td>
  <td><a href="https://github.com/XanaduAI/MrMustard/pull/521/files#diff-8102ef64b48fba0fce2e567a777086aadde18006cf456fc45581af5b89685e59">+1/-1</a>&nbsp; &nbsp; &nbsp; </td>

</tr>

<tr>
[...]

PR #532, opened 2024-12-10
Ansatz contract

apchytr:
**Context:** Separating out from [Batch Refactor](https://github.com/XanaduAI/MrMustard/pull/521). 

**Description of the Change:** Removed `__getitem__` and `__matmul__` from Ansatz in favor of `contract`. Cleanup `BackendNumpy`'s `sum` and renamed `axes` to `axis` to match tf and np.  

**Benefits:** Clearer implementation of contractions between Ansatz


ziofil (approved):
yep! 👍

ziofil (dismissed):
thanks for this! make sure you remove any leftover `_contract_idxs` since they are not needed any longer

ziofil (approved):

PR #535, opened 2024-12-23
Light cleanup of class methods

apchytr:
**Context:** Noticed we can cleanup some of our class methods on `states` and `transformations`. I believe a lot of this is just remnants of the Representation refactor where now we can rely on `from_ansatz` and have `cls` handle the proper return types. 

**Description of the Change:** Some abstract methods are now defined and removed unnecessary code.


ziofil (dismissed):
Nice! Just a minor comment about a return type. Good to go!

ziofil (approved):

PR #536, opened 2025-01-10
Removing tensorflow-io-gcs-filesystem dependency

apchytr:
**Context:** `tensorflow-io-gcs-filesystem` doesn't seem to be used by MrM (tested locally) and causes issues with Windows

**Description of the Change:** removed `tensorflow-io-gcs-filesystem` from our list of dependencies

**Benefits:** Fixed installation issues with Windows. Cleaner list of dependencies. 

**Possible Drawbacks:** Needs to be tested with other systems in case it is required somewhere


ziofil (approved):
👍

PR #537, opened 2025-01-14
Ability to choose DM ordering of wires

arsalan-motamedi:
**Context:**
Currently, MM uses the non-standard ordering of wires when outputting DMs in Fock basis.

Added `MM_order` to the `from_fock` functionality of DMs so that by `MM_order = False` we get the standard convention back.

**Benefits:**
Easier for Fock manipulations. 

**Possible Drawbacks:**
None.

**Related GitHub Issues:**
None.

### DIFF
--- a/mrmustard/lab_dev/states/dm.py
+++ b/mrmustard/lab_dev/states/dm.py
@@ -303,7 +303,7 @@ class DM(State):
         Returns:
             array: The Fock representation of this component.
         """
-        array = super()._representation.fock_array(shape or self.auto_shape(), batched)
+        array = super().fock_array(shape or self.auto_shape(), batched)
         if standard_order:
             m = self.n_modes
             axes = tuple(range(m, 2 * m)) + tuple(
--- a/mrmustard/lab_dev/states/dm.py
+++ b/mrmustard/lab_dev/states/dm.py
@@ -288,21 +288,21 @@ class DM(State):
         self, shape: int | Sequence[int] | None = None, batched=False, standard_order: bool = False
     ) -> ComplexTensor:
         r"""
-        Returns an array representation of this component in the Fock basis with the given shape.
-        If the shape is not given, it defaults to the ``auto_shape`` of the component if it is
-        available, otherwise it defaults to the value of ``AUTOSHAPE_MAX`` in the settings.
-The ``standard_order`` boolean argument lets one choose the standard convention for the index ordering of the density matrix. For a single mode, if ``standard_order=True`` the returned 2D array :math:`rho_{ij}` has a first index corresponding to the "left" (ket) side of the matrix and the second index to the "right" (bra) side. Otherwise, MrMustard's convention is that the bra index comes before the ket index. In other words, for a single mode, the array returned by ``fock_array`` with ``standard_order=False`` (false by default) is the transpose of the standard density matrix. For multiple modes, the same applies to each pair of indices of each mode.
+                Returns an array representation of this component in the Fock basis with the given shape.
+                If the shape is not given, it defaults to the ``auto_shape`` of the component if it is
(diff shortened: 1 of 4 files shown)


mrmustard/lab_dev/states/dm.py
```diff
@@ -284,6 +284,34 @@ def expectation(self, operator: CircuitComponent):
 
         return result
 
+    def fock_array(
+        self, shape: int | Sequence[int] | None = None, batched=False, MM_order: bool = True
```
apchytr, line 288: Maybe we should rename to `standard_order=False` or `bra_ket_order=True`? Or even `transposed=False`?

mrmustard/lab_dev/states/dm.py
```diff
@@ -284,6 +284,34 @@ def expectation(self, operator: CircuitComponent):
...
+            batched: Whether the returned representation is batched or not. If ``False`` (default)
+                it will squeeze the batch dimension if it is 1.
+            MM_order: The ordering of the wires. If ``MM_order = True``, then the conventional ordering
+            of bra-ket is chosen. However, if one wants to get the actual matrix representation in the
+            standard conventions of linear algebra, then ``MM_order=False`` must be chosen.
+        Returns:
+            array: The Fock representation of this component.
+        """
+        m = self.n_modes
+        axes = tuple(range(m, 2 * m)) + tuple(
```
apchytr, line 307: I think it makes sense to move this under `if not MM_order:`

apchytr (commented):
Thanks Arsalan!

mrmustard/lab_dev/states/dm.py
```diff
@@ -284,6 +284,34 @@ def expectation(self, operator: CircuitComponent):
...
+            range(m)
+        )  # to take care of multi-mode case, otherwise, for a single mode we could just use a simple transpose method
+        array = self._representation.fock_array(shape or self.auto_shape(), batched)
```
apchytr, line 310:
```suggestion
        array = super().fock_array(shape or self.auto_shape(), batched)
```
also works!

mrmustard/lab_dev/states/dm.py
```diff
@@ -284,6 +284,34 @@ def expectation(self, operator: CircuitComponent):
...
         return result
 
+    def fock_array(
+        self, shape: int | Sequence[int] | None = None, batched=False, standard_order: bool = False
+    ) -> ComplexTensor:
+        r"""
+        Returns an array representation of this component in the Fock basis with the given shape.
+        If the shape is not given, it defaults to the ``auto_shape`` of the component if it is
+        available, otherwise it defaults to the value of ``AUTOSHAPE_MAX`` in the settings.
+
```
ziofil, line 294:
```suggestion
The ``standard_order`` boolean argument lets one choose the standard convention for the index ordering of the density matrix. For a single mode, if ``standard_order=True`` the returned 2D array :math:`rho_{ij}` has a first index corresponding to the "left" (ket) side of the matrix and the second index to the "right" (bra) side. Otherwise, MrMustard's convention is that the bra index comes before the ket index. In other words, for a single mode, the array returned by ``fock_array`` with ``standard_order=False`` (false by default) is the transpose of the standard density matrix. For multiple modes, the same applies to each pair of indices of each mode.

```

tests/test_lab_dev/test_states/test_dm.py
```diff
@@ -395,3 +395,12 @@ def test_is_physical(self, modes):
         rho = 2 * rho
         assert not rho.is_physical
         assert Ket.random(modes).dm().is_physical
+
+    def test_from_fock(self):
```
ziofil, line 399:
```suggestion
    def test_fock_array_ordering(self):
```

ziofil (approved):
thanks!

apchytr (approved):

PR #538, opened 2025-01-15
Windows fixes

apchytr:
**Context:** Attempting to install MrMustard on a Windows system revealed some bugs

**Description of the Change:** Including a dependency for tf-intel. Fixing an issue with hermite_renormalized where `shape` could be of type `Tuple(Int32, Int64, Int64, Int64,...)` causing numba issues


timmysilv:
why win32 specifically? also, did you run `poetry.lock` on a windows machine? not sure what all those python-version lockfile changes are for

apchytr:
> why win32 specifically? also, did you run `poetry.lock` on a windows machine? not sure what all those python-version lockfile changes are for

win32 is the one python supports (see [Windows Platform](https://docs.python.org/3/library/platform.html)). I believe its referring to the API and not necessarily a 32 bit architecture. 

The lockfile changes seem to be coming from running `poetry lock` on `poetry>2.0.0`. Must be additional information they introduced in higher versions

ziofil (approved):
thanks!

PR #539, opened 2025-01-16
Renaming and moving things around to match architecture 

apchytr:
**Context:** Offline discussion has made it apparent that certain objects within lab_dev (such as `Circuit`) should be renamed and grouped together to reflect the architectural decisions we're making. Circuits now should refer to a future 'Circuit' layer and what is currently in lab_dev is the 'Computational Graph' layer.

**Description of the Change:** Renamed `circuits.py` to `computational_graph.py`, `circuit_components.py` to `graph_component.py`. Moved `graph_component.py`, `branch_and_bound.py` and `computational_graph.py` to a `computational_graphs` folder. Renamed `Circuit` to `ComputationalGraph`, `GraphComponent` to `LightGraphComponent`, `CircuitComponent` to `GraphComponent`. Updated docs to reflect these changes. 

**Benefits:** Makes things clearer 

**Possible Drawbacks:** Since lab_dev has been around for so long people have most likely gotten used to terms such as `CircuitComponent`, `Circuit`, etc so there will be an adjustment period.


apchytr:
Would love to hear people's opinion on what we should actually name things. Perhaps `computational_graph` is better?

PR #540, opened 2025-01-17
Fixing Ansatz contract signature

apchytr:
**Context:** The Ansatz contract method was changing behaviour with it defaulting idxs to `None`. Thanks to @timmysilv for pointing out the issue.

**Description of the Change:** `Ansatz.contract`'s `idx1` and `idx2` no longer default to `None` but rather a `tuple`


timmysilv (approved):
awesome! just confirming - users shouldn't explicitly pass None to this method (as the typehint now suggests), right?

PR #541, opened 2025-01-20
Remove `julia` code and dependencies

apchytr:
Opening on behalf of @aplund 

**Context:** Julia code was introduced previously to try and work around algorithmic instabilities in the computation of Fock coefficients from Gaussian Bargmann functions.  This introduced a hard dependency on `julia`, which made installation tricker.  In the end, it was found that the more stable approach was achieved by changing the algorithm to average over the possible lattice sites.

**Description of the Change:** This change removes the `julia` code and the dependencies of that code.  The PRECISION_BITS_HERMITE_POLY setting is removed, as that was the switch to turn on the `julia` implementation. The `julia` code has been pushed into the branch
<https://github.com/XanaduAI/GaussianBargmannFock> which stores implementations of this function.

**Benefits:** There is no hard dependency on `juliacall` and the `julia` language.  This should ease installation, particularly on Windows platforms.

**Possible Drawbacks:** The ability switch to using the `julia` implementation is gone without a setting to bring it back quickly.


timmysilv:
happy to see the hard dependency go away, though I have some questions:
- have we tried installing on Windows and confirmed that `juliacall` doesn't work nicely there?
- can you undo the changes to `poetry.lock` and re-run `poetry lock` with the `--no-update` option to keep things stable?

apchytr:
> OverrideNeeded

@aplund what version of poetry do you have installed? Is the environment a fresh one?

aplund:
> @aplund what version of poetry do you have installed? Is the environment a fresh one?

It's not a fresh env, but I can rebuild it and try again:

```
$ poetry --version
Poetry (version 1.8.4)
```

apchytr:
> > @aplund what version of poetry do you have installed? Is the environment a fresh one?
> 
> It's not a fresh env, but I can rebuild it and try again:
> 
> ```
> $ poetry --version
> Poetry (version 1.8.4)
> ```

Yeah can you try with poetry 1.8.5 and a fresh env? If that doesnt work Im not sure what's happening since I cant reproduce the error locally

aplund:
Looks like poetry is super-picky for minor version numbers.  I used 1.8.5 and all good.

timmysilv:
re: trying it. I did, and I hit this issue: https://github.com/JuliaLang/julia/issues/52205

I tried the simple solutions, and they didn't work. Ultimately, downloading Julia manually and following [this comment](https://github.com/MilesCranmer/PySR/issues/566#issuecomment-2174021336) worked, but this is unfortunately involved for Windows users.

timmysilv (approved):

PR #547, opened 2025-01-30
Converting tests that use lab to lab_dev 

apchytr:
**Context:** We still have a lot of test files that depend on lab. Here we switch to using lab_dev. 

**Description of the Change:** Updated tests


ziofil:
lab was removed in other PR

PR #549, opened 2025-02-03
Fixing Mr Mustard's build

apchytr:
**Context:** Slight issue with `pyproject.toml` means when installing Mr Mustard via a .whl or .tar.gz file causes a `ModuleNotFound` issue. This is fixed by updating `setuptools` package discovery. Note: some of this is caused by the fact we have a "flat-layout" structure rather than an "src-layout".

**Description of the Change:** Updated `pyproject.toml`. Also fixed our `black` issues by setting it to `22.12.0`


ziofil (approved):
I trust you! ^^'

PR #550, opened 2025-02-05
new fast diagonal strategy for conditional density matrix

ziofil:
**Context:**
The diagonal algorithm with one leftover mode could use a speedup since we are seeking 20x by EOY.

The `hermite_renormalized_1leftoverMode` has been hooked up to a new algorithm, which is faster than the current one by about 5x.


mrmustard/math/lattice/strategies/fast_diagonal.py
Ganar-lab, line 1: I also have a potential stupid idea to stabilize the algorithm (at the cost of quite some slowdown): could it be possible to somewhat do a periodic back-propagation to recompute the previous values from the new values?
Example: dimension=2, at weight=3 of the algorithm: compute weight 3 as usual, then weights 4 and 5, then use weights 4 as pivot to back compute weight 3. Then go again for weight 4 (+ recycle computations?).
I think this way, the algorithm would be super slow, but the fundamental question is: could numerical imprecisions reduce if you average the computations from backward pivots (usual way) with those from forward pivots? I am under the impression that even though the additional computations are derived from the entry they try to compute, it would have the effect of averaging the "noise" of numerical imprecisions.

Ganar-lab, line 1: If my last question on using forward pivots is not totally stupid, I would then suggest to merge it with the previous comment (=use 2 pivots whenever you can instead of 1): if that last comment doesn't stabilize the algorithm, I would use forward pivot computation just for the entries that can be computed with only one pivot instead of two (with backward pivots). It would ensure that all entries are computed with averaging two computations (removing "weaknesses"), which is what is done in vanilla no?
Note: I reasoned in dimension=2. I don't know if it makes sense in higher dimensions.

mrmustard/math/backend_numpy.py
Ganar-lab, line 1: As the algorithm is still instable, shouldn't we not make fast_diagonal the computation by default ? I might be unfamiliar with how default is chosen, but as it is not obvious for me, I ask the question.

mrmustard/math/lattice/strategies/fast_diagonal.py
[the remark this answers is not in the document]
ziofil: Yes, stability slightly improves, but I wasn't happy with the slowdown given that the improvement is very tiny. It seems like paying a very large price for very little gains.

ziofil: I had never thought of this. I don't know if it would work, because if the higher values are "more wrong" than the lower ones, recomputing the lower ones using the upper ones might make them worse rather than improving them. Also, the recurrence relations may need to be modified to be able to use upper pivots, which complicates things a little. But I can think about it.

mrmustard/math/backend_numpy.py
ziofil: It's just as unstable as the previous implementation, but faster. So I made this one the default.

mrmustard/math/lattice/strategies/fast_diagonal.py
```diff
@@ -0,0 +1,176 @@
...
+"""
+
+from itertools import product
+from functools import lru_cache
+import numpy as np
+from mrmustard.math.lattice.strategies.vanilla import vanilla_stable
+
+__all__ = ["fast_diagonal"]
+
+SQRT = np.sqrt(np.arange(10000))
```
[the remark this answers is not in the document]
ziofil: why? we would have to recompute the square roots every time we call that function

mrmustard/math/lattice/strategies/fast_diagonal.py
```diff
@@ -0,0 +1,176 @@
...
+
+
+def fast_diagonal(A, b, c, output_cutoff: int, pnr_cutoffs: tuple[int, ...]):
+    r"""
+    Computes an array of conditional density matrices using the fast diagonal strategy.
+    """
+    output_shape = (output_cutoff, output_cutoff)
+    L = len(pnr_cutoffs) + 1  # total number of modes
+    perm = [i for m in range(L) for i in (m, m + L)]
+    A = np.array(A)
```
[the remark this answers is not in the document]
ziofil: I have tried to convert it to numba, but it's not faster because we are working with dictionaries, not arrays.

mrmustard/math/backend_numpy.py
Ganar-lab: Thanks for clarifying!

mrmustard/math/lattice/strategies/fast_diagonal.py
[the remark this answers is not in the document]
Ganar-lab: I see, thanks for the answer!

mrmustard/math/backend_jax.py
```diff
@@ -748,7 +750,7 @@ def hermite_renormalized_1leftoverMode_reorderedAB(
         function = partial(hermite_multidimensional_1leftoverMode, cutoffs=tuple(cutoffs))
```
apchytr, line 750: Shouldn't we also update the signature of `hermite_renormalized_1leftoverMode_reorderedAB` to match `hermite_renormalized_1leftoverMode`? 

Also, doesn't the partial call to function need to be updated here? I.e. 

```suggestion
        function = partial(hermite_multidimensional_1leftoverMode, output_cutoff = cutoffs[0], pnr_cutoffs=cutoffs[1:])
```

ziofil: Good catch for partial, it must be something the tests are not covering. Regarding `hermite_renormalized_1leftoverMode_reorderedAB`, as soon as I'll write the backward pass of the new diagonal algorithm we won't need it anymore (and we can delete the compactFock stuff entirely), so I didn't want to spend time on it.

apchytr (approved):

PR #554, opened 2025-02-11
Switching Mr Mustard to numpy 2.0+ 

apchytr:
**Context:** Switching to `numpy 2.0+` allows us to make use of additional features and bug fixes 

**Description of the Change:** Update pyproject.toml to use `numpy 2.0+ ` as well as any additional changes necessary


timmysilv:
do we want this as part of the 1.0 release?

timmysilv:
the reason would be if we have users that want new features on their projects, but those projects depend on numpy < 2

mrmustard/math/backend_numpy.py
```diff
@@ -45,6 +45,7 @@
     hermite_multidimensional_diagonal_batch,
 )
 
+np.set_printoptions(legacy="1.25")
```
ziofil, line 48: what does this do? do we need it?

ziofil (approved):
just a small question, thanks!

mrmustard/math/backend_numpy.py
apchytr: One change between the versions of numpy is how numeric scalars are printed. E.g. `3.0` becomes `np.float64(3.0)`. This breaks the tests that we have related to the `Circuit` repr where the simplest fix is to just use [np.set_printoptions](https://numpy.org/doc/2.2/reference/generated/numpy.set_printoptions.html) and keep the legacy display.

apchytr:
> the reason would be if we have users that want new features on their projects, but those projects depend on numpy < 2

I know there's some projects internal to Xanadu that have Mr Mustard as a dep. Do we know if any of those are going to have issues with this change?

timmysilv (approved):

PR #557, opened 2025-02-22
ExpAnsatz

ziofil:
**Context:**
This is part of a simplification aimed at speeding up the next steps for MM. It's an ansatz with just the Exponential part, and the same functionality as the PolyExp ansatz.

Adds ExpAnsatz and related tests

**Benefits:**
Simpler code (no polynomial part to worry about).


ziofil:
will redo when the new ansatz is in and only if needed

Issue #481, opened 2024-09-03
MemoryError


JacobHast:
### Before posting a bug report

- [X] I have searched exisisting GitHub issues to make sure the issue does not already exist.

### Expected behavior

I expect to be able to handle reasonable states within my computers memory

### Actual behavior

For some states, mrmustard throws a MemoryError:

MemoryError: Unable to allocate 564. GiB for an array with shape (194481, 194481) and data type complex128

### Reproduces how often

For me it happens if I do 2 rounds of breeding with 2-mode GBS states produced by detecting 20 photons, using autoshape max cutoff of 40

### System information

```shell
Mr Mustard: a differentiable bridge between phase space and Fock space.
Copyright 2021 Xanadu Quantum Technologies Inc.

Python version:            3.10.14
Platform info:             Windows-10-10.0.19045-SP0
Installation path:         
c:\Users\jacob\miniconda3\envs\mrmustard_dev\lib\site-packages\mrmustard
Mr Mustard version:        0.7.3
Numpy version:             1.23.5
Numba version:             0.59.1
Scipy version:             1.14.1
The Walrus version:        0.21.0
TensorFlow version:        2.17.0
```

### Source code

```shell
import mrmustard.lab_dev as mm
import numpy as np

# GBS state
state = (
    mm.SqueezedVacuum(modes=[0, 1], r=[0.75, -0.75])
    >> mm.BSgate(modes=[0, 1], theta=0.9)
    >> mm.Number(modes=[0], n=20).dual
)
state = state.normalize()

# Breed 1st round
state2 = (
    (state.on([0]) >> state.on([1]))
    >> mm.BSgate(modes=[0, 1], theta=np.pi / 4)
    >> mm.QuadratureEigenstate(modes=[1], phi=np.pi / 2).dual
)

# Breed 2nd round
state3 = (
    (state2.on([0]) >> state2.on([1]))
    >> mm.BSgate(modes=[0, 1], theta=np.pi / 4)
    >> mm.QuadratureEigenstate(modes=[1], phi=np.pi / 2).dual
)

# Do something on output state
state3.normalize()
```

### Tracebacks

```shell
---------------------------------------------------------------------------
MemoryError                               Traceback (most recent call last)
Cell In[13], line 27
     20 state3 = (
     21     (state2.on([0]) >> state2.on([1]))
     22     >> mm.BSgate(modes=[0, 1], theta=np.pi / 4)
[...]

PR #559, opened 2025-02-24
Forcing different representations to use Fock 

apchytr:
**Context:** Issues like [MemoryError #481](https://github.com/XanaduAI/MrMustard/issues/481) are currently present in develop. These occur when attempting to compute the L2_norm of a state with a non scalar c. The solution presented here is to force different representation contractions to occur in Fock and to force the L2 norm to compute in Fock when the number of derived variables is > 0. 

**Description of the Change:**  Representation now raises when two different ansatz are met. Circuit component now handles changing the representation and uses `to_fock` instead. Computing the L2 is done in Fock when c is not scalar. Added #481 as a test.

**Benefits:** normalization and purity calculation issues resolved


mrmustard/lab_dev/samplers.py
```diff
@@ -169,13 +169,14 @@ def _validate_probs(self, probs: Sequence[float], atol: float) -> Sequence[float
             atol: The absolute tolerance to validate with.
         """
         atol = atol or settings.ATOL
+        probs = math.abs(probs)
```
ziofil, line 172: is this to go from complex to float?

ziofil (approved):
great!

tests/test_lab_dev/test_states/test_ket.py
```diff
@@ -122,6 +125,29 @@ def test_normalize(self, coeff):
         normalized = state.normalize()
         assert np.isclose(normalized.probability, 1.0)
 
+    def test_normalize_poly_dim(self):
+        # https://github.com/XanaduAI/MrMustard/issues/481
+        with settings(AUTOSHAPE_PROBABILITY=0.99999999):
```
apchytr, line 130: @ziofil If this becomes a consistent enough issue should we just set the global `AUTOSHAPE_PROBABILITY` higher?

ziofil (approved):
let's merge :+1:

ziofil (approved):
Reminder for the future:
1. we need to restrict mixing linear superpositions and batch dimensions
2. L2 norm and related functionality needs to be aware

This goes in now and there will be a follow-up PR

PR #564, opened 2025-03-03
Remove parallel initialization of built in CircuitComponents in Mr Mustard 

apchytr:
**Context:** There are a number of reasons to remove the parallel initialization functionality in lab_dev. Most importantly is that it conflicts with our batching epic. Here, we remove said functionality as well as include a significant amount of cleanup. 

As an example, 
```
coherent_state = Coherent(modes=[0], x=[0.1])
```
is no longer the supported syntax. The preferred syntax now is
```
coherent_state = Coherent(mode=0, x=0.1)
``` 
where each mode must be independently initialized (either via the standard init or via `on`). 

As a result we have two groups of built-in `CircuitComponents`. Ones that are single mode and initialize with `mode: int` and ones that are multi-mode initializing with `mode: tuple[int, ...]`.  E.g. a couple of multi-mode components:
```
vacuum_state = Vacuum((0,1,2)) # can also be initialized via Vacuum(0) 
bs_gate = BSgate((0,1), theta, phi)
```

For reference, batching will adhere to numpy broadcast rules and syntactically look like
```
batched_x = math.astensor([[0.1], [0.2], [0.3], [0.4], [0.5]]) # batch_size (5,1)
batched_y = math.astensor([[0.1, 0.2, 0.3, 0.4, 0.5, 0.6, 0.7]]) # batch_size (1,7)
coherent_state = Coherent(mode=0, x=batched_x, y=batched_y) # batch_size (5,7)
```

**Description of the Change:** Removed parallel initialization and misc changes as a result. Lots of documentation cleanup. Standardized modes to be a `tuple` of modes for robustness and nicer syntax. Removed `_getitem_builtin`. Removed `test_opt_lab.py` as that was mostly kept as a reference for timing but really just slowed down builds.

**Benefits:** Will enable us to introduce batching. Removes some weird functionality we had w.r.t. multi-mode components.

**Possible Drawbacks:** Removes the convenience of parallel initialization but not any core functionality


mrmustard/lab_dev/states/dm.py
```diff
@@ -366,32 +366,15 @@ def _ipython_display_(self):  # pragma: no cover
...
-            raise ValueError(msg)
-
-        if self._parameters:
-            # if ``self`` has a parameter set it means it is a built-in state,
-            # in which case we slice the parameters
-            return self._getitem_builtin(modes)
-
-        # if ``self`` has no parameter set it is not a built-in state,
-        # in which case we trace the representation
+    def __getitem__(self, idx: int | Collection[int]) -> State:
```
ziofil, line 369: should we leave the docstring?

mrmustard/lab_dev/states/gaussian_state.py
```diff
@@ -76,22 +74,22 @@ def __init__(
...
-
         self._representation = self.from_ansatz(
             modes=modes,
             ansatz=PolyExpAnsatz.from_function(
                 fn=triples.gket_state_Abc, symplectic=self.parameters.symplectic
             ),
         ).representation
 
-    def _getitem_builtin(self, modes: set[int] | Sequence[int]):
+    def __getitem__(self, idx: int | Collection[int]):
```
ziofil, line 84: I think we should use `int | Sequence` because `Collection` is not ordered. I think it was wrong before your change to type hint it with `set`.

mrmustard/lab_dev/states/gaussian_state.py
```diff
@@ -76,22 +74,22 @@ def __init__(
...
+        Override the default __getitem__ method to handle symplectic slicing.
         """
-
-        remaining_modes = [mode for mode in self.modes if mode not in modes]
-
-        return self >> TraceOut(remaining_modes)
+        idx = (idx,) if isinstance(idx, int) else idx
+        if not set(idx).issubset(set(self.modes)):
+            raise ValueError(f"Expected a subset of ``{self.modes}``, found ``{idx}``.")
+        trace_out_modes = set(self.modes) ^ set(idx)
```
ziofil, line 91: this is a bit risky because you might get a result with scrambled modes. `[mode for mode in self.modes if mode not in modes]` preserves the order.

ziofil, line 91: this is a bit risky because you might get a result with scrambled modes. `[mode for mode in self.modes if mode not in modes]` preserves the order.

mrmustard/lab_dev/states/gaussian_state.py
```diff
@@ -155,15 +148,17 @@ def __init__(
...
-        The slicing method for a GDM state.
+        Override the default __getitem__ method to handle symplectic slicing.
         """
-        remaining_modes = [mode for mode in self.modes if mode not in modes]
-
-        return self >> TraceOut(remaining_modes)
+        idx = (idx,) if isinstance(idx, int) else idx
+        if not set(idx).issubset(set(self.modes)):
+            raise ValueError(f"Expected a subset of ``{self.modes}``, found ``{idx}``.")
+        trace_out_modes = set(self.modes) ^ set(idx)
```
ziofil, line 163: same as above

mrmustard/physics/triples.py
```diff
@@ -252,19 +252,19 @@ def gket_state_Abc(symplectic: RealMatrix):
     return A[:m, :m], b, c
 
 
-def gdm_state_Abc(betas: Vector, symplectic: RealMatrix):
+def gdm_state_Abc(beta: float, symplectic: RealMatrix):
```
ziofil, line 255: I don't think it makes sense to restrict beta to be a float here. The generic gaussian density matrix can have different temperature in different modes. This is not a parallelizable state, it's usually entangled (unless the symplectic acts unitarily in each mode).

ziofil (commented):
wow this was a lot of work. thanks for fixing a lot of docstrings too. I left a few comments

mrmustard/lab_dev/utils.py
```diff
@@ -51,7 +51,7 @@ def make_parameter(
     return Variable(value=value, name=name, bounds=bounds, update_fn=update_fn, dtype=dtype)
 
 
-def reshape_params(n_modes: str, **kwargs) -> Generator:
+def reshape_params(n_modes: int, **kwargs) -> Generator:
```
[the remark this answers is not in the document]
apchytr: Yeah, it's required for `GDM` once I revert those changes

ziofil (approved):
thanks!

PR #571, opened 2025-04-01
mm_einsum 

ziofil:
`mm_einsum` is a low-level function that contracts a network of ansatze with the possibility to fully customize the contraction order and output shapes.

This PR:
- introduces `mm_einsum`
- implements a fully batched hermite_renormalized
- adds `reorder_batch` to the ansatze


mrmustard/math/backend_manager.py
```diff
@@ -717,6 +718,29 @@ def hermite_renormalized_batch(
         """
         return self._apply("hermite_renormalized_batch", (A, B, C, shape))
 
+    def hermite_renormalized_full_batch(
```
apchytr, line 721: This is awesome thank you! We should probably adapt this so that there is a single `hermite_renormalized` that can handle the batched and unbatched cases

ziofil: Good idea. Let's open a PR after this one to do that.

mrmustard/physics/ansatz/base.py
```diff
@@ -145,6 +145,20 @@ def reorder(self, order: tuple[int, ...] | list[int]) -> Ansatz:
         Reorders the ansatz indices.
         """
 
+    @abstractmethod
+    def reorder_batch(self, order: Sequence[int]):
```
apchytr, line 149: Can we add dedicated tests for this method in test_polyexp_ansatz.py and test_array_ansatz.py?

mrmustard/physics/mm_einsum.py
```diff
@@ -0,0 +1,169 @@
...
+        PolyExpAnsatz: The converted PolyExpAnsatz.
+    """
+    if isinstance(ansatz, PolyExpAnsatz):
+        return ansatz
+    try:
+        A, b, c = ansatz._original_abc_data
+    except (TypeError, AttributeError):
+        # TODO: update identity_Abc when it supports batching
+        A, b, _ = identity_Abc(ansatz.core_dims)
+        A = math.broadcast_to(A, ansatz.batch_shape + (2 * ansatz.core_dims, 2 * ansatz.core_dims))
```
apchytr, line 166: Why do we need to broadcast to `2 * core_dims`? And would it simplify things to do `A, b, _ = identity_Abc(2 * ansatz.core_dims)` above?

mrmustard/physics/mm_einsum.py
```diff
@@ -0,0 +1,169 @@
...
+    Args:
+        ansatz: The ansatz to convert.
+        shape: The shape of the ArrayAnsatz.
+
+    Returns:
+        ArrayAnsatz: The converted ArrayAnsatz.
+    """
+    if isinstance(ansatz, ArrayAnsatz):
+        return ansatz.reduce(shape)
+    array = math.hermite_renormalized_full_batch(*ansatz.triple, shape)
```
apchytr, line 144: Is this assuming the ansatz is batched? What happens if it isn't? Do we append a batch dim above?

mrmustard/physics/mm_einsum.py
```diff
@@ -0,0 +1,169 @@
...
+from mrmustard.physics.ansatz import ArrayAnsatz, PolyExpAnsatz
+from mrmustard.physics.ansatz.base import Ansatz
+from mrmustard.physics.triples import identity_Abc
+
+
+def _ints(seq: list[int | str]) -> list[int]:
+    return [i for i in seq if isinstance(i, int)]
+
+
+def _strings(seq: list[int | str]) -> list[str]:
```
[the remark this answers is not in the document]
ziofil: I find this more readable

mrmustard/physics/mm_einsum.py
ziofil: Because an M-mode unitary has 2M wires. No that would be wrong, we need the identity on core_dims modes

.github/workflows/tests_numpy.yml
```diff
@@ -41,7 +41,7 @@ jobs:
       - name: Run tests
         run: |
           set -o pipefail
-          uv run --all-extras pytest tests --backend=numpy -p no:warnings --tb=native ${{ env.UPLOAD_TIMINGS == 'true' && '--durations=0 -vv ' || '' }}| tee durations.txt
+          uv run --all-extras pytest -x tests --backend=numpy -p no:warnings --tb=native ${{ env.UPLOAD_TIMINGS == 'true' && '--durations=0 -vv ' || '' }}| tee durations.txt
```
ziofil, line 44:
```suggestion
          uv run --all-extras pytest tests --backend=numpy -p no:warnings --tb=native ${{ env.UPLOAD_TIMINGS == 'true' && '--durations=0 -vv ' || '' }}| tee durations.txt
```

mrmustard/physics/mm_einsum.py
Ganar-lab, line 1: From https://xanaduhq.slack.com/archives/C0618L1LJAZ/p1743722727856009 we discussed that it would be a good idea to have actual contraction evaluation when the full graph is contracted. Is that a higher level functionality that will be added later on?

mrmustard/physics/mm_einsum.py
```diff
@@ -0,0 +1,203 @@
...
+        >>> result = mm_einsum(
+        ...     ket0.ansatz, [0],
+        ...     ket1.ansatz, [1],
+        ...     bs.ansatz, [2, 3, 0, 1],
+        ...     output=[2, 3],
+        ...     contraction_order=[(0, 2), (1, 2)],
+        ...     fock_dims={0: 20, 1: 20, 2: 10, 3: 10},
+        ... )
+        >>> assert isinstance(result, ArrayAnsatz)
+        >>> assert result.shape == (10, 10)
```
Ganar-lab, line 75: The example does not work:
```
from mrmustard.lab_dev import Ket, BSgate
from mrmustard.physics.mm_einsum import mm_einsum
from mrmustard.physics.ansatz import ArrayAnsatz

ket0 = Ket.random([0])
ket1 = Ket.random([1])
bs = BSgate((0, 1), theta=0.5, phi=0.3)
result = mm_einsum(
        ket0.ansatz, [0],
        ket1.ansatz, [1],
        bs.ansatz, [2, 3, 0, 1],
        output=[2, 3],
        contraction_order=[(0, 2), (1, 2)],
        fock_dims={0: 20, 1: 20, 2: 10, 3: 10},
)
print(isinstance(result, ArrayAnsatz))
```
It prints False. It's a PolyExpAnsatz.

mrmustard/physics/mm_einsum.py
```diff
@@ -0,0 +1,203 @@
...
+    operators, allowing for flexible contraction of complex tensor networks in quantum optics.
+
+    Args:
+        *args: Alternating sequence of Ansatz objects and their associated index lists.
+            Each Ansatz is followed by a list of indices (strings for batch axes, integers for Hilbert space indices).
+        output (list[int | str]): The indices (batch names and Hilbert space indices) to keep in the output.
+        contraction_order (list[tuple[int, int]]): The order in which to contract the Ansatze,
+            specified as pairs of their positions in the input.
+        fock_dims (dict[int, int]): Mapping from Hilbert space indices (int) to Fock space dimensions.
+            If a dimension is 0, the corresponding Ansatz is converted to Bargmann (PolyExpAnsatz) form.
```
Ganar-lab, line 53: Do I understand correctly that these fock_dims are not used if the concerned wires of the inputs are all in Bargman? Similarly, they are ignored if they are all in AnsatzArray (and not used as a cutoff of some sort).
I don't understand if that's the intended use for it. If I give a entry for a dimension with a number, like 10, it should enforce it as a cutoff on the Bargman object, and convert it even if there is a Bargman object to contract with, no? Else, it seems a bit weird to set something that is ignored.
Also for the output wires, shouldn't it require some value? To know if the user wants things in Bragman or Fock.

mrmustard/physics/mm_einsum.py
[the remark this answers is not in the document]
ziofil: Ah I had missed the point about sparse tensors being auto-converted to dense (I thought it wasn't going to be a problem to wrap them in an ArrayAnsatz). Ok, I can allow for passing arrays directly.

ziofil: I'm not sure I understand what you mean. Let's chat on slack.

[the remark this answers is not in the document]
ziofil: done

mrmustard/physics/mm_einsum.py
```diff
@@ -0,0 +1,203 @@
...
+        >>> ket0 = Ket.random([0])
+        >>> ket1 = Ket.random([1])
+        >>> bs = BSgate((0, 1), theta=0.5, phi=0.3)
+        >>> # Contract the network: (ket0 & ket1) >> BS
+        >>> result = mm_einsum(
+        ...     ket0.ansatz, [0],
+        ...     ket1.ansatz, [1],
+        ...     bs.ansatz, [2, 3, 0, 1],
+        ...     output=[2, 3],
+        ...     contraction_order=[(0, 2), (1, 2)],
```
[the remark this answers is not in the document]
ziofil: Yes, that is called Linear assignment. The other standard (the one I implemented) is called SSA (static single assignment). opt_einsum has [helper functions](https://github.com/dgasmith/opt_einsum/blob/c826bb7df16f470a69f7bf90598fc27586209d11/opt_einsum/paths.py#L63) to convert between the two. I find SSA is more intuitive than Linear. But sure, if you prefer I can switch to Linear.

mrmustard/physics/mm_einsum.py
ziofil: Works now, thanks

RyosukeNORO (commented):
Thank you. It looks good to me

mrmustard/math/backend_numpy.py
```diff
@@ -228,10 +229,14 @@ def set_diag(self, array: np.ndarray, diag: np.ndarray, k: int) -> np.ndarray:
...
-    def einsum(self, string: str, *tensors) -> np.ndarray | None:
-        if type(string) is str:
-            return np.einsum(string, *tensors)
-        return None  # provide same functionality as numpy.einsum or upgrade to opt_einsum
+    def einsum(self, string: str, *tensors, optimize: bool | str = False) -> np.ndarray:
+        has_sparse = any(isinstance(t, sparse.COO) for t in tensors)
+        has_dense = any(not isinstance(t, sparse.COO) for t in tensors)
+
+        if has_sparse and has_dense:
+            tensors = [t.todense() if isinstance(t, sparse.COO) else t for t in tensors]
```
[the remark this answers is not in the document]
ziofil: No, not all of them, only the ones that interact through an einsum call. Unless I misunderstood, `opt_einsum.contract` does not allow you to mix sparse and dense matrices, so this line avoids the problem by densifying the sparse ones. Is that not the right thing to do?

mrmustard/physics/mm_einsum.py
ziofil: thanks for pointing this out, it's fixed now.

mrmustard/math/backend_numpy.py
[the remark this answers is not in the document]
RyosukeNORO: Well, when I did `opt_einsum.contract` with a dense matrix and a sparse matrix, the result was a sparse matrix. I used opt_einsum 3.4.0.

code
```
import opt_einsum as oe
import sparse
dense_matrix = np.array([[1,2],[3,4]])
sparse_matrix = sparse.COO.from_numpy(np.array([[5,6],[7,8]]))
result = oe.contract(
    "ab, ab -> ab",
    dense_matrix,
    sparse_matrix,
)
print(type(result))
```
output
```
<class 'sparse._coo.core.COO'>
```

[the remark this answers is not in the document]
ziofil: You are right, that code works. The failure comes from opt_einsum using a cached version of einsum... I'm investigating.

mrmustard/physics/ansatz/array_ansatz.py
```diff
@@ -202,11 +204,9 @@ def contract(
 
         reduced_array1 = array1[tuple(slices1)]
         reduced_array2 = array2[tuple(slices2)]
-
-        result = math.einsum(einsum_str, reduced_array1, reduced_array2)
+        result = math.einsum(einsum_str, reduced_array1, reduced_array2, optimize=True)
```
apchytr, line 207: Do we want to keep `optimize=True` now? Originally I think it was included to help fix the sparse + dense issues but it looks like the solution now is globally setting SPARSE_AUTO_DENSIFY

mrmustard/physics/mm_einsum.py
```diff
@@ -0,0 +1,296 @@
...
+    Contracts a network of Ansatze according to a custom contraction order, supporting both Fock
+    and Bargmann representations, batch dimensions, and named indices.
+
+    This function generalizes the concept of Einstein summation (einsum) to quantum states and
+    operators, allowing for flexible contraction of complex tensor networks in quantum optics.
+
+    The path_type argument controls the style in which the contraction path is expressed.
+    - LA pops the contrated ansatze from the list and adds the result to the end.
+    If [a1,a2,a3] is the list of ansatze the contraction order is [(0,1), (0,2)] means a3 @ (a1 @ a2).
+    - SSA uses a dictionary to keep track of the contracted ansatze, so that there can be static single
```
apchytr, line 65: UA should also be documented here

mrmustard/physics/mm_einsum.py
```diff
@@ -0,0 +1,296 @@
...
+    }
+
+    return tuple(shape_a[i] for i in core_idx_a), tuple(shape_b[i] for i in core_idx_b)
+
+
+def prepare_idx_out(
+    indices: dict[int, list[int | str]], a: int, b: int, output: list[int | str]
+) -> list[int | str]:
+    r"""
+    Prepares the index of the output of the contraction of two ansatze.
```
apchytr, line 236: Can you document Args + Returns

mrmustard/physics/mm_einsum.py
```diff
@@ -0,0 +1,296 @@
...
+    if len(output_idx_str) > 1:
+        result = result.reorder_batch([final_idx_str.index(i) for i in output_idx_str])
+    if len(output_idx_int) > 1:
+        result = result.reorder([final_idx_int.index(i) for i in output_idx_int])
+    if final_idx_int and all(fock_dims.get(i, 1) == 0 for i in final_idx_int):
+        result = to_bargmann(result)
+    return result
+
+
+SHAPE = tuple[int, ...]
```
apchytr, line 160: Move to typing.py?

tests/test_physics/test_mm_einsum.py
```diff
@@ -0,0 +1,487 @@
...
+from mrmustard.physics.mm_einsum import mm_einsum
+from mrmustard.physics.ansatz import ArrayAnsatz, PolyExpAnsatz
+
+from ..conftest import skip_jax
+
+
+class TestMmEinsum:
+    """Unit tests for the mm_einsum function."""
+
+    settings.SEED = 42
```
apchytr, line 32: Gah looking through our tests we set the seed globally in many places. Probably best to just set it globally in the conftest.py

mrmustard/physics/mm_einsum.py
```diff
@@ -0,0 +1,296 @@
...
+) -> PolyExpAnsatz | ArrayAnsatz | ArrayLike:
+    r"""
+    Contracts a network of Ansatze according to a custom contraction order, supporting both Fock
+    and Bargmann representations, batch dimensions, and named indices.
+
+    This function generalizes the concept of Einstein summation (einsum) to quantum states and
+    operators, allowing for flexible contraction of complex tensor networks in quantum optics.
+
+    The path_type argument controls the style in which the contraction path is expressed.
+    - LA pops the contrated ansatze from the list and adds the result to the end.
```
apchytr, line 63:
```suggestion
    - LA pops the contracted ansatze from the list and adds the result to the end.
```

mrmustard/physics/mm_einsum.py
```diff
@@ -0,0 +1,296 @@
...
+
+def _ints(seq: list[int | str]) -> list[int]:
+    return [i for i in seq if isinstance(i, int)]
+
+
+def _strings(seq: list[int | str]) -> list[str]:
+    return [i for i in seq if isinstance(i, str)]
+
+
+def ua_to_linear(ua: list[tuple[int, int]]) -> list[tuple[int, int]]:
```
apchytr, line 35: Just a side comment: This might be nice to have in `utils.py` if we choose to support `optimal_path` making use of these different path types

tests/test_physics/test_mm_einsum.py
ziofil: agreed, but not in this same PR

mrmustard/physics/mm_einsum.py
ziofil: if other files or the user start needing we can move them, for now I think it makes it more readable to keep things in this file

mrmustard/physics/ansatz/array_ansatz.py
```diff
@@ -225,11 +224,11 @@ def reduce(self, shape: Sequence[int]) -> ArrayAnsatz:
             >>> assert fock1 == fock2
 
             >>> fock3 = fock1.reduce((2, 2, 2))
-            >>> array3 = [[[0, 1], [3, 4]], [[9, 10], [12, 13]]]
+            >>> array3 = np.array([[[0, 1], [3, 4]], [[9, 10], [12, 13]]])
```
apchytr, line 227:
```suggestion
            >>> array3 = math.astensor([[[0, 1], [3, 4]], [[9, 10], [12, 13]]])
```

apchytr (commented):
Great!!

apchytr (approved):

PR #576, opened 2025-04-10
Updated API of contract for Ansatze

ziofil:
**Context:**
The API for the `contract` method of `PolyExpAnsatz` and `ArrayAnsatz` did not allow for the efficient implementation of diagonal operators (e.g. Rgate, Attenuator kraus operators, etc...) because hilbert space indices could not be repeated at the output.

The API now requires `idx1`, `idx2`, `idx_out` to be fully specified, and it's the same API of `mm_einsum`, so no need to do conversions on the fly when setting up a call to mm_einsum.

**Benefits:**
- Allows for more general contractions of ArrayAnsatz (in particular it enables efficient diagonal operators)
- Minimal parsing needed for mm_einsum calls
- Easier to understand what the contraction does

**Possible Drawbacks:**
- callers of `contract` need to update the way they call the function.


mrmustard/lab_dev/circuit_components_utils/trace_out.py
```diff
@@ -81,17 +80,33 @@ def __custom_rrshift__(self, other: CircuitComponent | complex) -> CircuitCompon
...
+            wires = ket + bra
+            _, ovlp, core_out = (wires.adjoint @ wires)[0].contracted_labels(self.wires)
+            core1 = list(range(C))
+            core2 = list(range(C, 2 * C))
+            for i in ovlp[: len(self.modes)]:
+                core2[i] = core1[i]
+            batch = [chr(97 + i) for i in range(B)]
+            if not bra:
+                ansatz = other.ansatz.conj.contract(
+                    other.ansatz, batch + core1, batch + core2, core_out
```
apchytr, line 98: Do we need `batch + core_out`? I think we want to be zipping the batch dimensions here

mrmustard/lab_dev/circuit_components_utils/trace_out.py
```diff
@@ -81,17 +80,33 @@ def __custom_rrshift__(self, other: CircuitComponent | complex) -> CircuitCompon
...
+            B = other.ansatz.batch_dims
+            C = other.ansatz.core_dims
+            wires = ket + bra
+            _, ovlp, core_out = (wires.adjoint @ wires)[0].contracted_labels(self.wires)
+            core1 = list(range(C))
+            core2 = list(range(C, 2 * C))
+            for i in ovlp[: len(self.modes)]:
+                core2[i] = core1[i]
+            batch = [chr(97 + i) for i in range(B)]
+            if not bra:
```
apchytr, line 96: I think this can be simplified if we do something like 
```
ansatz1 = other.ansatz.conj if not bra else other.ansatz
ansatz2 = other.ansatz if not bra else other.ansatz.conj
ansatz = ansatz1.contract(ansatz2, batch + core1, batch + core2, core_out)
```

mrmustard/physics/ansatz/polyexp_ansatz.py
```diff
@@ -239,38 +246,89 @@ def from_function(cls, fn: Callable, **kwargs: Any) -> PolyExpAnsatz:
...
-        return PolyExpAnsatz(A, b, c)
+        # --- Parse and Validate Inputs ---
+        batch1 = [label for label in idx1 if isinstance(label, str)]
+        core1 = [label for label in idx1 if isinstance(label, int)]
+        batch2 = [label for label in idx2 if isinstance(label, str)]
+        core2 = [label for label in idx2 if isinstance(label, int)]
+        batch_out = [label for label in idx_out if isinstance(label, str)]
+        core_out = [label for label in idx_out if isinstance(label, int)]
+
+        # Check dimensions match expected counts
```
apchytr, line 290: Hmmm I dont think we should be validating the batch labels here. Mostly because `join_Abc` already validates batch_str. Part of what was nice about having the `batch_str` argument is that we didn't have to touch it until the lowest level `join_Abc` call (except to maybe generate it from `mode`)

mrmustard/physics/representations.py
```diff
@@ -122,14 +122,16 @@ def contract(self, other: Representation, mode: str = "kron"):
...
+            eins_str = zip_batch_strings(self.ansatz.batch_dims, other.ansatz.batch_dims)
         elif mode == "kron":
             eins_str = outer_product_batch_str(self.ansatz.batch_shape, other.ansatz.batch_shape)
-        ansatz = self.ansatz.contract(other.ansatz, batch_str=eins_str, idx1=idx_z, idx2=idx_zconj)
+        in_, batch_out = eins_str.split("->")
+        batch1, batch2 = in_.split(",")
+        ansatz = self.ansatz.contract(
+            other.ansatz, list(batch1) + idx1, list(batch2) + idx2, list(batch_out) + idx_out
+        )
         ansatz = ansatz.reorder(perm) if perm else ansatz
```
apchytr, line 135: Looking at `PolyExpAnsatz.contract` it looks like the reordering is now happening at the ansatz. Therefore do we need this here?

mrmustard/lab_dev/circuit_components_utils/trace_out.py
ziofil: good catch, I'll add a test too

mrmustard/physics/representations.py
ziofil: Yes because idx_out has no information about what wires represent, it's just a sorted collection of the leftover labels. Since the permutation is already computed in __matmul__ of wires, I left it there and called reorder.

ziofil: there might be a more efficient way to do it, but it's a very marginal gain. I'll add a TODO to revisit.

mrmustard/physics/ansatz/polyexp_ansatz.py
ziofil: You're right, it's just that this is very informative.

mrmustard/physics/ansatz/polyexp_ansatz.py
```diff
@@ -409,7 +477,24 @@ def reorder(self, order: Sequence[int]):
...
+
+        core_dims_indices_A = range(self.batch_dims, self.batch_dims + 2)
+        core_dims_indices_b = range(self.batch_dims, self.batch_dims + 1)
+        core_dims_indices_c = range(self.batch_dims, self.batch_dims + self.num_derived_vars)
+
+        new_A = math.transpose(self.A, list(order) + list(core_dims_indices_A))
+        new_b = math.transpose(self.b, list(order) + list(core_dims_indices_b))
+        new_c = math.transpose(self.c, list(order) + list(core_dims_indices_c))
+
+        return PolyExpAnsatz(new_A, new_b, new_c)
```
ziofil, line 494: add missing _lin_sup

arsalan-motamedi (approved):

PR #579, opened 2025-04-21
Bye bye lab!

ziofil:
This PR removes the old `lab` module and reorganizes the rest of the codebase (also, I fixed a sign issue in the two mode squeezed vacuum).

This PR comes with a soundtrack:
https://youtu.be/LRyrWN-fftE


mrmustard/lab_dev/states/ket.py
```diff
@@ -273,6 +273,14 @@ def expectation(self, operator: CircuitComponent):
 
         return result
 
+    def fidelity(self, other: State) -> float:
+        r"""
+        The fidelity between this ket and another state.
+        """
```
arsalan-motamedi, line 279: Seems like the fidelity is defined with a squared. I suggest clarifying it i.e., adding

```suggestion
        r"""
        The fidelity between this ket and another state.
        
        .. details::
        
        .. math::
            F(|\psi\rangle, \phi\rangle) = |\langle \psi, \phi \rangle|^2
        """
```

mrmustard/lab_dev/states/dm.py
```diff
@@ -272,6 +274,29 @@ def expectation(self, operator: CircuitComponent):
 
         return result
 
+    def fidelity(self, other: State) -> float:
+        r"""
+        The fidelity between this DM and another state.
+
+        Args:
```
arsalan-motamedi, line 281: Same as above, I believe it is worth noting that we include the `^2` in the definition of fidelity.

arsalan-motamedi (commented):
Other than the two minor comments above, it looks great to me!

mrmustard/lab_dev/states/dm.py
arsalan-motamedi: Perhaps, just:

```suggestion
        The fidelity between this DM and another state. Note that the fidelity is the squared overlap, consistent with the pure state's fidelity.

        Args:
```

arsalan-motamedi (approved):

PR #580, opened 2025-04-22
Binomial stable strategies with dictionaries

ziofil:
This PR adds the function `vanilla_step_dict_stable`, which is then used in `binomial_step_dict_stable` and `binomial_step_dict_stable_no_prob` for computing subsets of Fock indices using dictionaries rather than arrays. This allows for custom methods to be implemented more straightforwardly than messing around with strides and array indices at the expense of doing hashmap lookups rather than hopping through memory addresses.


ziofil:
not needed any longer

PR #585, opened 2025-05-05
Cleanup of settings and remove trainer.py

apchytr:
**Context:** Some more cleanup of legacy things. This time is cleaning up settings from the old lab and removing trainer.py. In addition to this `AUTOSHAPE_MAX` is three distinct settings `AUTOSHAPE_MAX`, `AUTOSHAPE_MIN` and `DEFAULT_FOCK_SIZE`. The first two are settings for clipping in `autoshape_numba` the last is the default shape when autoshape cannot be applied. 

**Description of the Change:** Removed old MM settings. Added `AUTOSHAPE_MIN` and `DEFAULT_FOCK_SIZE`. `autoshape_numba` now clips according to `AUTOSHAPE_MAX` and `AUTOSHAPE_MIN`. Removed trainer.py and ray dependency. Merged `autoshape` implementations on `DM` and `Ket` to `State`


mrmustard/math/lattice/autoshape.py
```diff
@@ -149,5 +150,7 @@ def autoshape_numba(A, b, c, max_prob, max_shape) -> int:  # pragma: no cover
             ) / SQRT[k + 2]
             norm += np.abs(buf3[(k + 1) % 2, 1])
             k += 1
-        shape[m] = k or 1
+
```
apchytr, line 153:
```suggestion
```

mrmustard/lab/states/base.py
```diff
@@ -339,6 +340,72 @@ def from_quadrature(
...
+
+        Example:
+        .. code-block::
+            >>> from mrmustard import math
+            >>> from mrmustard.lab import Vacuum
+
+            >>> assert math.allclose(Vacuum([0]).fock_array(), 1)
+        """
+        batch_shape = (
+            self.ansatz.batch_shape[:-1] if self.ansatz._lin_sup else self.ansatz.batch_shape
```
ziofil, line 376: this is not for now, but shouldn't `batch_shape` in the ansatz always return the batch shape? there shouldn't be the need to depend on `_lin_sup` externally from the ansatz.

mrmustard/lab/states/base.py
```diff
@@ -339,6 +340,72 @@ def from_quadrature(
...
+            try:  # fock
+                shape = self.ansatz.core_shape
+            except AttributeError:  # bargmann
+                if self.ansatz.num_derived_vars == 0:
+                    if not self.wires.ket or not self.wires.bra:
+                        ansatz = self.ansatz.conj & self.ansatz
+                    else:
+                        ansatz = self.ansatz
+                    A, b, c = ansatz.triple
+                    ansatz = ansatz / self.probability
```
ziofil, line 390: what's the point of normalizing now, after A,b,c have been taken out?

mrmustard/lab/states/base.py
```diff
@@ -339,6 +340,72 @@ def from_quadrature(
...
+        )
+        if batch_shape:
+            raise NotImplementedError("Batched auto_shape is not implemented.")
+        if not self.ansatz._lin_sup:
+            try:  # fock
+                shape = self.ansatz.core_shape
+            except AttributeError:  # bargmann
+                if self.ansatz.num_derived_vars == 0:
+                    if not self.wires.ket or not self.wires.bra:
+                        ansatz = self.ansatz.conj & self.ansatz
```
ziofil, line 386: what if self is a DM?

mrmustard/lab/transformations/dgate.py
```diff
@@ -103,11 +103,10 @@ def __init__(
     def fock_array(self, shape: int | Sequence[int] = None) -> ComplexTensor:
         r"""
         Returns the unitary representation of the Displacement gate using the Laguerre polynomials.
-        If the shape is not given, it defaults to the ``auto_shape`` of the component if it is
-        available, otherwise it defaults to the value of ``AUTOSHAPE_MAX`` in the settings.
+
         Args:
             shape: The shape of the returned representation. If ``shape`` is given as an ``int``,
-                it is broadcasted to all the dimensions. If not given, it is estimated.
+                it is broadcasted to all the dimensions. If not given, it is generated via ``auto_shape``.
```
ziofil, line 109: this is a unitary, we should probably say that it would use the default fock size

tests/test_lab/test_states/test_number.py
```diff
@@ -44,7 +44,7 @@ def test_init(self, modes, n, cutoffs):
     def test_auto_shape(self):
         # meant to cover the case where we have derived variables
         state = Number(0, 2).to_bargmann().dm()
-        assert state.auto_shape() == (settings.AUTOSHAPE_MAX, settings.AUTOSHAPE_MAX)
+        assert state.auto_shape() == (settings.DEFAULT_FOCK_SIZE, settings.DEFAULT_FOCK_SIZE)
```
ziofil, line 47: perhaps Number's shape should default to n+1?

mrmustard/lab/states/base.py
apchytr: Good catch 👍

mrmustard/lab/states/base.py
apchytr: In that case it should have both ket and bra wires and be covered in L387-L388

mrmustard/lab/states/base.py
apchytr: Yeah that's the functionality I wanted when I was doing exposing batches but I ran into issues implementing it. The solution I had was significantly uglier than what is currently present. I think it's worth revisiting with a clearer mindset however because I agree the ansatz should handle the `_lin_sup` logic

tests/test_lab/test_states/test_number.py
apchytr: I'm a little confused why this isnt already doing that? We set a manual shape on `Number` already and `auto_shape`'s `respect_manual_shape` is `True` by default... let me investigate

ziofil (approved):
Thanks Anthony!

PR #590, opened 2025-05-12
improved stellar decomposition

arsalan-motamedi:
**Context:**
There are flaky tests in our current implementations of stellar decomposition for `DM`s. Moreover, our current implementation of physical stellar decomposition on `Ket`s makes uses of `contract` which can be circumvented, and hence, improving the speed.

Addressing the issues above.

**Benefits:**
Having no flaky tests + faster decomposer

**Possible Drawbacks:**
None.

**Related GitHub Issues:**
None.


ziofil (approved):
thanks arsalan

PR #591, opened 2025-05-14
Stable Fock BSgate

ziofil:
This PR introduces a custom Fock implementation of the BSgate which is fast and stable.


mrmustard/lab/transformations/bsgate.py
```diff
@@ -104,3 +110,61 @@ def __init__(
...
+    ) -> ComplexTensor:
+        r"""
+        Returns the unitary representation of the Beam Splitter gate in the Fock basis.
+
+        Args:
+            shape: The shape of the returned representation. If ``shape`` is given as an ``int``,
+                it is broadcasted to all the dimensions. If not given, it defaults to
+                ``settings.DEFAULT_FOCK_SIZE``.
+            method: The method to use to compute the Fock array. Available methods are:
+                - ``"vanilla"``: The default method.
```
apchytr, line 125: Is there a better way to describe this? Probably don't want our "default method" to not be the default 😅

ziofil: ah, right

ziofil, line 125:
```suggestion
                - ``"vanilla"``: standard recurrence relation (not numerically stable, but slightly faster than the stable one).
```

apchytr (approved):

PR #594, opened 2025-05-23
JAX Optimizer

apchytr:
**Context:** Jax is an alternative framework to Tensorflow for running optimization on our quantum circuits. It brings along a number of benefits including JIT compilation that makes it preferable to the alternative. Here we introduce the `OptimizerJax` class that aims to replace our current `Optimizer`. Initial testing already indicates massive performance boosts especially at a significantly large number of steps. 

Note: This PR only implements support for euclidean optimizations with unitary, orthogonal and symplectic to be implemented in the future. 

**Description of the Change:** Introduces the `OptimizerJax` class with various vjps. Misc changes to make certain functions jit compatible. Updated a warning in ArrayAnsatz. Rem `add_at_modes`, `left_matmul_at_modes`, `right_matmul_at_modes`, `matvec_at_modes` in `BackendManager`. Rem `getitem` and `setitem` in numpy and tf backends. Rem `Autocast` in numpy and jax backends. Moved all the `math.custom_gradient` methods in `fock_utils` to the backend and removed `math.custom_gradient` 

**Benefits:** Massive performance boost for our optimizations.

**Possible Drawbacks:** As mentioned above we still do not currently support other non-euclidean update methods.


.github/workflows/tests_jax.yml
```diff
@@ -38,12 +38,28 @@ jobs:
         id: record_file
         run: echo "filename=durations_$(date +%s)_$(git rev-parse --short HEAD).txt" >> "$GITHUB_OUTPUT"
 
+      - name: Restore cached Primes
+        id: cache-primes-restore
+        uses: actions/cache/restore@v4
+        with:
+          path: |
+            /tmp/jax_cache
+          key: ${{ runner.os }}-primes
```
timmysilv, line 47: perhaps something more unique to this job
```suggestion
      - name: Restore JAX compilation cache
        id: jax-cache
        uses: actions/cache/restore@v4
        with:
          path: /tmp/jax_cache
          key: jax-cache-${{ matrix.python-version }}
```

.github/workflows/tests_jax.yml
```diff
@@ -38,12 +38,28 @@ jobs:
...
           uv run --all-extras pytest tests --backend=jax -p no:warnings --tb=native ${{ env.UPLOAD_TIMINGS == 'true' && '--durations=0 -vv ' || '' }}| tee durations.txt
           set +o pipefail
 
+      - name: Save Primes
+        id: cache-primes-save
+        uses: actions/cache/save@v4
+        with:
+          path: |
+            /tmp/jax_cache
+          key: ${{ steps.cache-primes-restore.outputs.cache-primary-key }}
```
timmysilv, line 61:
```suggestion
      - name: Save JAX cache
        uses: actions/cache/save@v4
        with:
          path: /tmp/jax_cache
          key: jax-cache-${{ matrix.python-version }}
```

timmysilv (commented):
just making sure we have different caches for different python versions but looks good

mrmustard/lab/states/base.py
```diff
@@ -387,14 +387,17 @@ def auto_shape(
...
+                    try:
+                        shape = autoshape_numba(
+                            math.asnumpy(A),
+                            math.asnumpy(b),
+                            math.asnumpy(c),
+                            max_prob or settings.AUTOSHAPE_PROBABILITY,
+                            max_shape or settings.AUTOSHAPE_MAX,
+                            min_shape or settings.AUTOSHAPE_MIN,
+                        )
+                    except TracerArrayConversionError:  # pragma: no cover
```
ziofil, line 399: Can we create a BackendError so that we don't use anything backend-specific in lab?

mrmustard/math/backend_jax.py
```diff
@@ -547,27 +535,7 @@ def hermite_renormalized_unbatched(
...
-                lambda A, b, c: strategies.vanilla_numba(
-                    shape, np.array(A), np.array(b), np.array(c)
-                ),
-                jax.ShapeDtypeStruct(shape, jnp.complex128),
-                A,
-                b,
-                c,
-            )
-        return G
+        return hermite_renormalized_unbatched(A, b, c, shape, stable)
```
ziofil, line 538: Why do we need a function calling an identically named function?

mrmustard/physics/fock_utils.py
```diff
@@ -414,63 +415,88 @@ def sample_homodyne(state: Tensor, quadrature_angle: float = 0.0) -> tuple[float
...
+        y: The displacement angle.
+        shape: The shape of the displacement matrix.
+        tol: The tolerance to determine if the displacement is small enough to be approximated by the identity.
 
-    if np.sqrt(x * x + y * y) > tol:
-        gate = strategies.displacement(tuple(shape), alpha)
+    Returns:
+        The matrix representing the displacement gate.
+    """
+    if math.backend_name == "jax":  # pragma: no cover
```
ziofil, line 431: I'm not a fan with polluting lab with things that depend on the backend. What would be the alternative to doing this?

mrmustard/physics/fock_utils.py
```diff
@@ -414,63 +415,88 @@ def sample_homodyne(state: Tensor, quadrature_angle: float = 0.0) -> tuple[float
...
+        return displacement_jax(x, y, shape, tol)
     else:
-        gate = math.eye(max(shape), dtype="complex128")[: shape[0], : shape[1]]
+        alpha = math.asnumpy(x) + 1j * math.asnumpy(y)
 
-    ret = math.astensor(gate, dtype=gate.dtype.name)
-    if math.backend_name in ["numpy", "jax"]:
```
ziofil, line 427: hmm we had it before too

mrmustard/math/backend_jax.py
hitarth64: Actually this is indeed a bit confusing for someone else not familiar with mrmustard taking a look in future. If it cannot be renamed, can we instead call it using `jax_vjps.hermite_renormalized_unbatched`?

mrmustard/physics/fock_utils.py
hitarth64: Can we have a routine in math backend called `displacement`? That would solve this issue.

mrmustard/physics/fock_utils.py
```diff
@@ -414,63 +415,88 @@ def sample_homodyne(state: Tensor, quadrature_angle: float = 0.0) -> tuple[float
...
-            math.asnumpy(math.conj(dLdGc)),
-            math.asnumpy(theta),
-            math.asnumpy(phi),
-        )
-        return math.astensor(dtheta, dtype=theta.dtype), math.astensor(dphi, dtype=phi.dtype)
+    if math.backend_name == "jax":  # pragma: no cover
+        return beamsplitter_jax(theta, phi, shape, method)
+    else:
+        t, s = math.asnumpy(theta), math.asnumpy(phi)
+        if method == "vanilla":
```
hitarth64, line 479: Similar comment as above: we can potentially replace these lines with `math.beamsplitter`. This can be done by moving `@math.custom_gradient` to the backend routine of tf.

mrmustard/training/optimizer_jax.py
```diff
@@ -0,0 +1,200 @@
...
+
+__all__ = ["OptimizerJax"]
+
+
+class Objective(eqx.Module):
+    r"""
+    A dataclass that contains the static and dynamic parameters of the model.
+    """
+    static: list[str]
+    dynamic: list[jax.Array]
```
hitarth64, line 46: Can I request to add docstring about what is stored as `static` and what is stored as `dynamic` parameters?

mrmustard/training/optimizer_jax.py
```diff
@@ -0,0 +1,200 @@
...
+
+    def __init__(self, trainable_params: dict[str, Variable]):
+        self.static = list(trainable_params.keys())
+        self.dynamic = [array.value for array in trainable_params.values()]
+
+    def __call__(self, cost_fn: Callable, by_optimizing: Sequence[CircuitComponent]):
+        trainable_params = OptimizerJax._get_trainable_params(by_optimizing)
```
hitarth64, line 53: A little bit of logic here could help others: something like "it is updating the values of each of the individual gate parameters since jax does not propagate local updates globally."

mrmustard/training/optimizer_jax.py
```diff
@@ -0,0 +1,200 @@
...
+    @eqx.filter_jit
+    def make_step(
+        self,
+        optim,
+        loss,
+        model,
+        opt_state,
+    ):
+        r"""
+        Make a step of the optimization.
```
hitarth64, line 121: Some docstrings and if possible typing hints would be really helpful for future debugging- thanks anthony!

mrmustard/training/optimizer_jax.py
```diff
@@ -0,0 +1,200 @@
+# Copyright 2025 Xanadu Quantum Technologies Inc.
```
hitarth64, line 1: Pretty sure that even though this jaxoptimizer looks innocent, it probably took days. Great work!

hitarth64 (commented):
Massive amount of work, anthony! Left a few comments - please let me know and I re-review it quickly. Thanks!

timmysilv (commented):
no major concerns from me!

mrmustard/lab/states/number.py
```diff
@@ -73,9 +73,7 @@ def __init__(
         )
         self._representation = self.from_ansatz(
             modes=(mode,),
-            ansatz=ArrayAnsatz.from_function(
-                fock_state, n=self.parameters.n, cutoffs=self.parameters.cutoff
-            ),
+            ansatz=ArrayAnsatz.from_function(fock_state, n=n, cutoffs=cutoff),
```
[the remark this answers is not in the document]
apchytr: `cutoffs` here refers to the cutoffs argument in `fock_state` which does support an array of cutoffs so I think it's fine as is

mrmustard/math/backend_jax.py
```diff
@@ -520,7 +509,7 @@ def sqrtm(self, tensor: jnp.ndarray, dtype, rtol=1e-05, atol=1e-08) -> jnp.ndarr
 
     # Special functions for optimization
     def DefaultEuclideanOptimizer(self):
-        return jax.experimental.optimizers.adam(learning_rate=0.001)
+        return optax.inject_hyperparams(optax.adamw)
```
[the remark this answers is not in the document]
apchytr: What do you mean by more freedom?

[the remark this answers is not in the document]
hitarth64: Oh I meant that for tf, learning rate is fixed. However, here since inject_hyperparams is used, a user can in principle later on provide extra arguments to the optimizer. Not a problem - just jax is providing extra layer of flexibility.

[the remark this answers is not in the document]
hitarth64: I see. I was not aware of it - thanks!

mrmustard/math/backend_numpy.py
```diff
@@ -416,16 +408,12 @@ def trace(self, array: np.ndarray, dtype=None) -> np.ndarray:
...
         return np.transpose(a, axes=perm)
 
-    @Autocast()
     def update_tensor(
         self, tensor: np.ndarray, indices: np.ndarray, values: np.ndarray
     ) -> np.ndarray:
-        indices = self.atleast_nd(indices, 2)
-        for i, v in zip(indices, values):
-            tensor[tuple(i)] = v
+        tensor[indices] = values
```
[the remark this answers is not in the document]
apchytr: I don't think it should just based off the docstring's "Updates a tensor in place with the given values." description. But also, the tf backend can also update in place depending on if the memory in the existing tensor can be re used. I guess that would be equivalent to calling np.asarray here. Otherwise the Jax backend is the only "odd man out" as it doesn't support in place updates at all.

mrmustard/training/optimizer_jax.py
```diff
@@ -0,0 +1,200 @@
...
+        orthogonal_lr: float = 0.1,
+        euclidean_lr: float = 0.001,
+        stable_threshold=1e-6,
+    ):
+        self.learning_rate = {
+            update_euclidean: euclidean_lr,
+            update_symplectic: symplectic_lr,
+            update_unitary: unitary_lr,
+            update_orthogonal: orthogonal_lr,
+        }
```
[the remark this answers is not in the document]
apchytr: Ah wasn't supposed to be there in the first place since we dont make use of any of these. Removed 👍

mrmustard/training/optimizer_jax.py
```diff
@@ -0,0 +1,200 @@
...
+        Minimizes the given cost function by optimizing ``Variable``s either on their own or within a ``CircuitComponent`` / ``Circuit``.
+
+        Args:
+            cost_fn: A function that will be executed in a differentiable context in
+                order to compute gradients as needed.
+            by_optimizing: A list of elements that contain the parameters to optimize.
+            max_steps: The minimization keeps going until the loss is stable or max_steps are
+                reached (if ``max_steps=0`` it will only stop when the loss is stable).
+        """
+        if settings.PROGRESSBAR:
```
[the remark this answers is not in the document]
timmysilv: I thought about this pattern, and I'd actually prefer `Bar` to just be a function that returns a `YesBar` or `NoBar` based on inputs, and have both returned objects follow an "API" (no need to be so official imo, but we could be). Said API entails the implementation of the enter and exit dunders, and maybe an `advance` method (meaning `NoBar` shouldn't return None, but just a dummy) ?

mrmustard/physics/fock_utils.py
apchytr: Moved all the math.custom_gradient methods to the backend 👍 . As a result we no longer make use of math.custom_gradient anywhere so also removed that

hitarth64 (approved):
Looks great! 🚀

mrmustard/training/optimizer_jax.py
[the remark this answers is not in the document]
ziofil: why not just `with YesBar() if settings.PROGRESSBAR else NoBar() as bar:`?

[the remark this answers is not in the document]
apchytr: I think that works but `with ProgressBar(settings.PROGRESSBAR) as prog_bar:` is a bit cleaner imo. It's nice to not have to think about YesBar and NoBar outside of the ProgressBar itself

mrmustard/lab/states/base.py
```diff
@@ -387,14 +386,18 @@ def auto_shape(
...
                     A, b, c = ansatz.triple
-                    shape = autoshape_numba(
-                        math.asnumpy(A),
-                        math.asnumpy(b),
-                        math.asnumpy(c),
-                        max_prob or settings.AUTOSHAPE_PROBABILITY,
-                        max_shape or settings.AUTOSHAPE_MAX,
-                        min_shape or settings.AUTOSHAPE_MIN,
-                    )
+                    try:
```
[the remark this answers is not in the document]
apchytr: Will be following this up with a PR to cleanup `auto_shape` so I'll update documentation there 👍

PR #595, opened 2025-05-23
Add `out` keyword to hermite_renormalized

ziofil:
In the numpy backend we can now pass `out=my_array` for it to fill a memory-mapped array rather than allocating new memory.


tests/test_math/test_vanilla.py
```diff
@@ -165,35 +165,59 @@ def test_full_batch_vjp(self):
...
         """
+        if out:
+            skip_jax()
+            skip_tf()
         A, b, c = random_triple(2, (), seed=673)
         shape = (3, 3)
-        G = math.hermite_renormalized(A, b, c, shape, stable=stable)
+        if out:
+            out_arr = math.zeros(shape, dtype=math.complex128)
+            G = math.hermite_renormalized(A, b, c, shape, stable=stable, out=out_arr)
```
apchytr, line 180: I think there needs to be a check to see if out_arr and G are the same array

mrmustard/physics/representations.py
```diff
@@ -140,13 +140,17 @@ def contract(self, other: Representation, mode: str = "kron"):
...
+    def fock_array(
+        self, shape: int | Sequence[int], out: ComplexTensor | None = None
+    ) -> ComplexTensor:
         r"""
         Returns an array of this representation in the Fock basis with the given shape.
 
         Args:
             shape: The shape of the returned array, not including batch dimensions. If ``shape`` is
             given as an ``int``, it is broadcast to all the dimensions.
+            out: if provided, the result will be stored in this tensor.
```
apchytr, line 152: Hmm this is a little misleading imo. It's specifically the result of hermite renormalized that is written to `out` but for example the Fock path doesn't make use of `out`. Also, I believe the operations after the hermite renormalized call mean that no matter what the resulting array is not the same as the one provided in `out`

mrmustard/math/backend_manager.py
```diff
@@ -717,40 +723,67 @@ def hermite_renormalized(
             A: The A matrix. Can be unbatched (shape D×D) or batched (shape B×D×D).
             b: The b vector. Can be unbatched (shape D) or batched (shape B×D).
             c: The c scalar. Can be scalar or batched (shape B).
-            shape: The shape of the final tensor.
+            shape: The shape of the final tensor (excluding batch dimensions)
             stable: Whether to use the numerically stable version of the algorithm (also slower).
+            out: if provided, the result will be stored in this tensor.
```
apchytr, line 728:
```suggestion
            out: If provided, the result will be stored in this tensor.
```

PR #598, opened 2025-05-29
Adding support for `optimize` in einsum 

apchytr:
**Context:** The `optimize` parameter in `einsum` can be quite useful but our custom backend does not current support it's use.

**Description of the Change:** Took out the `optimize` changes from [mm_einsum #571](https://github.com/XanaduAI/MrMustard/pull/571). Added an `EINSUM_OPTIMIZE` setting defaulting to greedy. Fixed a bug in DM when calling fock_array with standard ordering with a linear superposition state

### DIFF
--- a/mrmustard/lab/states/dm.py
+++ b/mrmustard/lab/states/dm.py
@@ -392,7 +392,9 @@ class DM(State):
         array = super().fock_array(shape or self.auto_shape())
         if standard_order:
             m = self.n_modes
-            batch_dims = self.ansatz.batch_dims
+            batch_dims = (
+                self.ansatz.batch_dims - 1 if self.ansatz._lin_sup else self.ansatz.batch_dims
+            )
             axes = (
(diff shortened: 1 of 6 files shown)


mrmustard/math/backend_jax.py
```diff
@@ -27,6 +27,7 @@
 import equinox as eqx  # pragma: no cover
 from jax import tree_util  # pragma: no cover
 
+
```
apchytr, line 30:
```suggestion
```

mrmustard/lab/states/dm.py
```diff
@@ -392,7 +392,9 @@ def fock_array(
         array = super().fock_array(shape or self.auto_shape())
         if standard_order:
             m = self.n_modes
-            batch_dims = self.ansatz.batch_dims
+            batch_dims = (
```
ziofil, line 395: should the batch_dims property of the ansatz do this directly?

PR #599, opened 2025-05-30
Adding persistent compilation cache to BackendJax

apchytr:
**Context:** Here we enable a persistent compilation cache for jitted functions in `BackendJax`. This avoids the need to recompile functions everytime we do something like run pytest.

Split out from [JAX Optimizer](https://github.com/XanaduAI/MrMustard/pull/594)

**Description of the Change:** Enabled a persistent cache located in `mrmustard/math/tmp/jax_cache`. Updated the github workflow to also make use of this cache (thanks  to @timmysilv). Reordered dependencies alphabetically. Cleanup backend pragma no covers. Added `interactive` group for jupyter notebooks.

**Benefits:** Avoids instances of needing to recompile the jitted functions.


timmysilv:
this looks great, just one concern: I'm a little uneasy about this relative path within mrmustard. can you instead use [platformdirs.user_cache_dir](https://github.com/tox-dev/platformdirs) to get a folder that works on all platforms, eg. `f"{user_cache_dir('mrmustard')}/jax_cache"`?

mrmustard/math/backend_jax.py
```diff
@@ -16,31 +16,36 @@
...
-import jax.scipy as jsp  # pragma: no cover
-import numpy as np  # pragma: no cover
-import equinox as eqx  # pragma: no cover
-from jax import tree_util  # pragma: no cover
-
-from .autocast import Autocast  # pragma: no cover
-from .backend_base import BackendBase  # pragma: no cover
-from .lattice import strategies  # pragma: no cover
-from .lattice.strategies.compactFock.inputValidation import (  # pragma: no cover
+from __future__ import annotations
```
hitarth64, line 19: Thank you @apchytr for fixing this. How were you able to remove "# pragma no cover"?

hitarth64 (approved):
Looks great! Do we time savings even on github actions? If so, how much? Thanks!

PR #601, opened 2025-06-03
Cleaning up how we skip tests

apchytr:
**Context:** Noticed we could improve the way we skip tests by making use of `pytest.mark`. Now instead of having `skip_np`, `skip_tf` and `skip_jax` we have a single `pytest.mark.requires_backend`.

Taken from [JAX Optimizer #594](https://github.com/XanaduAI/MrMustard/pull/594)

**Description of the Change:** Rem `skip_np`, `skip_tf` and `skip_jax` replaced with `pytest.mark.requires_backend`. Fixed opt_lab test 

**Benefits:** Cleaner implementation of skipping with more flexibility.

### DIFF
--- a/tests/test_math/test_compactFock.py
+++ b/tests/test_math/test_compactFock.py
@@ -1,7 +1,7 @@
 """
 Unit tests for mrmustard.math.compactFock.compactFock~
 """
-
+import pytest
 import numpy as np
 
 from mrmustard import math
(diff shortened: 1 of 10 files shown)


tests/test_math/test_compactFock.py
```diff
@@ -83,15 +79,12 @@ def cost_fn():
         assert opt.opt_history[i - 1] >= opt.opt_history[i]
 
 
+@pytest.mark.requires_backend()  # TODO: implement gradient of hermite_renormalized_1leftoverMode
```
timmysilv, line 82: nice, maybe we can document this somewhere (no args = skip due to interface incompatibility)?

timmysilv (approved):
nice! if we decorate a test with two different `requires_backend` decorators, will it work for both backends?

timmysilv (approved):

PR #606, opened 2025-06-13
Batched from XY

arsalan-motamedi:
**Context:**
We have not currently implemented `from_XY` with batches. Also, the computations in `XY_to_channel_Abc` are not compatible with our batching conventions. The shape of `b` also has an extra dimension.

We now allow batched initialization via `.from_XY`.

**Benefits:**
Consistency

**Possible Drawbacks:**
None

**Related GitHub Issues:**
None

### DIFF
--- a/mrmustard/lab/transformations/base.py
+++ b/mrmustard/lab/transformations/base.py
@@ -513,9 +513,8 @@ class Channel(Map):
             This channel has a Bargmann triple that is computed in https://arxiv.org/pdf/2209.06069. We borrow
             the formulas from the paper to implement the corresponding channel.
         """
-        if X.shape[:-2] != ():
-            raise NotImplementedError("Batching is not implemented.")
-        if X.shape != (2 * len(modes_out), 2 * len(modes_in)):
+
+        if X.shape[-2:] != (2 * len(modes_out), 2 * len(modes_in)):
             raise ValueError(
                 f"The dimension of X matrix ({X.shape}) and number of modes ({len(modes_in), len(modes_out)}) don't match."
             )
--- a/mrmustard/physics/triples.py
+++ b/mrmustard/physics/triples.py
@@ -928,11 +928,26 @@ def XY_to_channel_Abc(
             "The dimension of X and Y matrices are not the same."
             f"X.shape = {X.shape}, Y.shape = {Y.shape}"
         )
-
-    xi = 1 / 2 * math.eye(2 * m, dtype=math.complex128) + 1 / 2 * X @ X.T + Y / settings.HBAR
+    batch_shape = X.shape[:-2]
+    if batch_shape != ():
+        Im = math.stack(
+            [math.eye(2 * m, dtype=math.complex128)] * int(math.prod(batch_shape))
+        ).reshape(batch_shape + (2 * m,) * 2)
+        im = math.stack([math.eye(m, dtype=math.complex128)] * int(math.prod(batch_shape))).reshape(
+            batch_shape + (m,) * 2
+        )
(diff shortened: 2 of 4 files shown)


mrmustard/lab/transformations/base.py
```diff
@@ -513,9 +513,8 @@ def from_XY(
             This channel has a Bargmann triple that is computed in https://arxiv.org/pdf/2209.06069. We borrow
             the formulas from the paper to implement the corresponding channel.
         """
-        if X.shape[:-2] != ():
-            raise NotImplementedError("Batching is not implemented.")
-        if X.shape != (2 * len(modes_out), 2 * len(modes_in)):
+
+        if X.shape[-2:] != (2 * len(modes_out), 2 * len(modes_in)):
             raise ValueError(
```
apchytr, line 518: I noticed we don't have this error documented. Can we include that here?

mrmustard/physics/triples.py
```diff
@@ -928,44 +928,62 @@ def XY_to_channel_Abc(
             "The dimension of X and Y matrices are not the same."
             f"X.shape = {X.shape}, Y.shape = {Y.shape}"
         )
-
-    xi = 1 / 2 * math.eye(2 * m, dtype=math.complex128) + 1 / 2 * X @ X.T + Y / settings.HBAR
+    batch_shape = X.shape[:-2]
+    if batch_shape != ():
+        Im = math.stack(
```
apchytr, line 933: Can you use math.broadcast_to instead? I think ditto for `im` and `Xm`

mrmustard/physics/triples.py
```diff
@@ -928,44 +928,62 @@ def XY_to_channel_Abc(
...
-
-    A = math.Xmat(2 * m) @ R @ xi_inv_in_blocks @ math.conj(R).T
-    temp = math.block([[(xi_inv @ d).reshape(2 * m, 1)], [(-X.T @ xi_inv @ d).reshape((2 * m, 1))]])
-    b = 1 / math.sqrt(complex(settings.HBAR)) * math.conj(R) @ temp
-    c = math.exp(-0.5 / settings.HBAR * d @ xi_inv @ d) / math.sqrt(math.det(xi))
+    R_transpose = math.einsum("...ij->...ji", R)
+    A = Xm @ R @ xi_inv_in_blocks @ math.conj(R_transpose)
+    temp_1 = math.einsum("...ij,...j->...i", xi_inv, d)
+    temp_2 = math.einsum("...ij,...jk,...k->...i", X_transpose, xi_inv, d)
+    temp = math.concat([temp_1, temp_2], -1)
```
apchytr, line 984: Are there better names we can give to `temp_1`, `temp_2` and `temp`?

mrmustard/lab/transformations/base.py
```diff
@@ -506,18 +506,23 @@ def from_XY(
...
             This channel has a Bargmann triple that is computed in https://arxiv.org/pdf/2209.06069. We borrow
             the formulas from the paper to implement the corresponding channel.
         """
-        if X.shape[:-2] != ():
-            raise NotImplementedError("Batching is not implemented.")
-        if X.shape != (2 * len(modes_out), 2 * len(modes_in)):
+
+        if X.shape[-2:] != (2 * len(modes_out), 2 * len(modes_in)) or Y.shape[-2:] != (
+            2 * len(modes_out),
+            2 * len(modes_out),
```
heltluke, line 522: Should one of these be `modes_in`?

arsalan-motamedi: Thanks for the careful review. The ``Y`` matrix is a square matrix of output size (not necessarily the same as ``X``). So, it's fine this way.

PR #608, opened 2025-06-18
Removing remnant legacy methods

apchytr:
**Context:** As an exercise, this PR aims to remove many of the methods that are no longer used in the core functionality of Mr Mustard.

**Description of the Change:** Updated `DM.physical_stellar_decomposition_mixed` to make use of `math.pinv` and updated the tensorflow `pinv` to account for the complex case. Updated several tests that globally mutated settings to instead make use of the context manager. Removed the following:

backend_manager.py
- angle, atleast_1d, atleast_2d, atleast_3d, block_diag, boolean_mask, constraint_func, convolution, lgamma, maximum, minimum, moveaxis, repeat, round, set_diag, squeeze, cholesky, Categorical, MultivariateNormalTriL, single_mode_to_multimode_vec, single_mode_to_multimode_mat, poisson, binomial_conditional_prob, convolve_probs_1d, convolve_probs

bargmann_utils.py
- norm_ket, trace_dm

fock_utils.py
- ket_to_dm, ket_to_probs, dm_to_probs, U_to_choi, number_means, number_variances, validate_contraction_indices, sample_homodyne

gaussian.py
- number_cov, log_negativity

gaussian_integrals.py
- reorder_abc

**Benefits:** A more lightweight Mr Mustard experience.

**Possible Drawbacks:** Many of these methods may still be useful for user to have access to outside of the core MM functionality. In this case, this PR can always serve as reference for these methods and we can reintroduce them on a "as needed" basis.

### DIFF
--- a/mrmustard/math/backend_manager.py
+++ b/mrmustard/math/backend_manager.py
@@ -23,7 +23,6 @@ from typing import Any, Callable, Sequence
 
 from jax.errors import TracerArrayConversionError
 import numpy as np
-from scipy.special import binom
 from scipy.stats import ortho_group, unitary_group
 
 from ..utils.settings import settings
@@ -90,7 +89,7 @@ all_modules = {
 }
 
 
-class BackendManager:  # pylint: disable=too-many-public-methods, fixme
+class BackendManager:
     r"""
     A class to manage the different backends supported by Mr Mustard.
     """
@@ -255,19 +254,6 @@ class BackendManager:  # pylint: disable=too-many-public-methods, fixme
         array2 = self.astensor(array2)
         return self._apply("allclose", (array1, array2, atol, rtol))
 
-    def angle(self, array: Tensor) -> Tensor:
-        r"""
-        The complex phase of ``array``.
-
-        Args:
-            array: The array to take the complex phase of.
-
-        Returns:
-            The complex phase of ``array``.
-        """
-
-        return self._apply("angle", (array,))
-
     def any(self, array: Tensor) -> bool:
         r"""Returns ``True`` if any element of array is ``True``, ``False`` otherwise.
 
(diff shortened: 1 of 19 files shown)


mrmustard/math/backend_manager.py
```diff
@@ -1689,45 +1433,26 @@ def random_symplectic(self, num_modes: int, max_r: float = 1.0) -> Tensor:
...
         return ortho_group.rvs(dim=N, random_state=settings.rng)
 
     def random_unitary(self, N: int) -> Tensor:
-        """a random unitary matrix in :math:`U(N)`"""
+        r"""
+        A random unitary matrix in :math:`U(N)`.
+        """
         if N == 1:
-            return self.exp(1j * settings.rng.uniform(size=(1, 1)))
+            return np.exp(1j * settings.rng.uniform(size=(1, 1)))
```
apchytr, line 1448:
```suggestion
            return self.exp(1j * settings.rng.uniform(size=(1, 1)))
```

heltluke (approved):
Love the use of the contextmanager to temporarily change settings in tests, and also the move from `numpy` to `math`. I'm a bit less sure about the new `scipy` imports though. I might have guessed that, where we can, we'd want to avoid `scipy` to make things jit-able?

Overall, it looks good to me, but likely worth gating on @rachelchadwick or @aplund regarding the deletions.

apchytr:
> I'm a bit less sure about the new `scipy` imports though. I might have guessed that, where we can, we'd want to avoid `scipy` to make things jit-able?

Is this in reference to replacing `math.lgamma` with `scipy.special.loggamma`? If so, this shouldn't affect our ability to jit as it's isolated to the numpy backend where we expect to only jit compile with the Jax backend (excluding numba functions). The jax backed makes use of `jax.lax.lgamma` instead which is already jitted 👍

aplund (approved):

PR #622, opened 2025-07-09
Add numba compilation cache on all njit functions

aplund:
**Context:** When tracking down a massive slowdown when using the IPython debugger (which seems to be an issue with that implementation), the slowdown occurred within the numba compilation.  It became clear that there was no caching of numba jit compilation and hence recompiled every instance.

**Description of the Change:** Add the cache=True kwarg to all the numba jitted functions.  The one exception is the binomial_subspace_basis functions which segfaults for me when loaded from the cache.  It looks like that function needs something extra as it must violate some assumed invariant between processes.

**Benefits:** The runtime for the tests goes from a couple of minutes to 10s of seconds on the second run.  There will also be similar speed-ups in the mrkite tests on the second run of tests.

**Possible Drawbacks:** This will not speed up the github CI tests as they start with nothing pre-compiled.  The usual issues with caches, of course, apply.  See:

https://numba.readthedocs.io/en/stable/developer/caching.html


apchytr (approved):
This is awesome Austin thanks! As for the CI tests there might be a way to set up a persistent cache similarly to how we do it for Jax. @timmysilv would it be possible to investigate this? Happy to include it either in this PR or in a subsequent.

heltluke (dismissed):

apchytr (approved):

timmysilv:
I just pushed the cache step changes for all three! I'll re-trigger CI after it finishes to make sure it's used correctly

apchytr (approved):

timmysilv (approved):

apchytr (dismissed):

apchytr (approved):

PR #632, opened 2025-07-22
feat: [sc-96057] Better README bagdes

heltluke:
The GitHub Workflow badges were for all branches instead of just `main` and the PyPI Python version badge only works for full releases and not pre-releases.

This PR:
- updates workflow badges to only change state if workflows fail/succeed on `main`
- updates Python version badge to rely on `pyproject.toml` and not PyPI
- updates logo images with new syntax (see https://github.com/orgs/community/discussions/16910)

Story details: https://app.shortcut.com/xanaduai/story/96057

### DIFF
--- a/.github/workflows/upload.yml
+++ b/.github/workflows/upload.yml
@@ -55,4 +55,4 @@ jobs:
           uv pip install --pre mrmustard
 
       - name: Print about
-        run: uv run python -c "import mrmustard; print(mrmustard.about())"
+        run: python -c "import mrmustard; print(mrmustard.about())"
--- a/.github/workflows/upload.yml
+++ b/.github/workflows/upload.yml
@@ -52,7 +52,7 @@ jobs:
       - name: Uninstall MrMustard and reinstall from PyPI
         run: |
           uv pip uninstall mrmustard
-          uv pip install --pre mrmustard
+          uv pip install --no-cache-dir --prerelease=allow mrmustard
 
       - name: Print about
         run: python -c "import mrmustard; print(mrmustard.about())"
--- a/.github/workflows/upload.yml
+++ b/.github/workflows/upload.yml
@@ -1,14 +1,7 @@
 name: Upload
 on:
-  push:
-    branches:
-      - develop
-      - main
-  pull_request:
-  workflow_call:
-
-  # release:
-  #   types: [published]
+  release:
+    types: [published]
 
 jobs:
   upload:
@@ -43,11 +36,11 @@ jobs:
       - name: Run tests
         run: uv run --no-project pytest tests -p no:warnings --tb=native --backend=tensorflow
 
-      # - name: Publish
-      #   uses: pypa/gh-action-pypi-publish@release/v1
-      #   with:
-      #     user: __token__
-      #     password: ${{ secrets.PIPY_TOKEN }}
+      - name: Publish
+        uses: pypa/gh-action-pypi-publish@release/v1
+        with:
+          user: __token__
+          password: ${{ secrets.PIPY_TOKEN }}
 
       - name: Install MrMustard from pypi
         run: uv pip install --pre --upgrade mrmustard
--- a/.github/workflows/upload.yml
+++ b/.github/workflows/upload.yml
@@ -21,6 +21,8 @@ jobs:
 
       - name: Build and install Mr Mustard
         run: |
+          uv venv
+          source .venv/bin/activate
           uv build
           uv pip install dist/mrmustard*.whl
           # Move to 'src' to properly test only installed package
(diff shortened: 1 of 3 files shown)


apchytr (dismissed):

pyproject.toml
```diff
@@ -9,13 +9,12 @@ description = "Differentiable quantum Gaussian circuits"
 authors = [
     {name= "Xanadu", email= "filippo@xanadu.ai"},
 ]
-license = {text = "Apache License 2.0"}
+license = "Apache-2.0"
```
heltluke, line 12: Changed on account of the following build log:
>/home/runner/work/_temp/setup-uv-cache/builds-v0/.tmpVrDmGW/lib/python3.12/site-packages/setuptools/config/_apply_pyprojecttoml.py:82: SetuptoolsDeprecationWarning: `project.license` as a TOML table is deprecated
!!

        ********************************************************************************
        Please use a simple string containing a SPDX expression for `project.license`. You can also use `project.license-files`. (Both options available on setuptools>=77.0.0).

        By 2026-Feb-18, you need to update your project and remove deprecated calls
        or your builds will no longer be supported.

        See https://packaging.python.org/en/latest/guides/writing-pyproject-toml/#license for details.
        ********************************************************************************

.github/workflows/upload.yml
```diff
@@ -29,20 +38,19 @@ jobs:
...
-          password: ${{ secrets.PIPY_TOKEN }}
+      # - name: Publish
+      #   uses: pypa/gh-action-pypi-publish@release/v1
+      #   with:
+      #     user: __token__
+      #     password: ${{ secrets.PIPY_TOKEN }}
 
       - name: Install MrMustard from pypi
-        run: uv pip install mrmustard
+        run: uv pip install --pre --upgrade mrmustard
```
heltluke: The `--pre` flag will ensure we can get the pre-release version (see https://pip.pypa.io/en/stable/cli/pip_install/#pre-release-versions).

.github/workflows/upload.yml
```diff
@@ -29,20 +38,19 @@ jobs:
           mv mrmustard src
 
       - name: Install only test dependencies
-        run: uv sync --inexact --no-install-project --all-extras
+        run: uv sync --only-dev --inexact --no-install-project
```
heltluke, line 34: We don't actually need to install all extras here, just those for testing:
```
dev = [
    "hypothesis==6.31.6",
    "pre-commit==3.7.1",
    "pytest-cov==3.0.0",
    "pytest==8.0",
    "ruff>=0.12.0",
]
```

heltluke:
There was a successful run of a modified "upload.yml" that only built, ran tests, installed from PyPI, and printed the "about" here: https://github.com/XanaduAI/MrMustard/actions/runs/16473307388/job/46568016618

pyproject.toml
timmysilv: I remember something complaining and asking me to format the license with this dict-like style once in the past. hope it's not an issue...

.github/workflows/upload.yml
```diff
@@ -29,7 +31,7 @@ jobs:
...
 
       - name: Run tests
         run: uv run --no-project pytest tests -p no:warnings --tb=native --backend=tensorflow
```
apchytr: Can we change the backend to JAX here (or even numpy)? Tensorflow is pretty slow and I dont want it slowing down releases

.github/workflows/upload.yml
timmysilv: although `uv pip` is similar to `pip`, it seems different here. According to its helptext, this should do it:
```suggestion
        run: uv pip install --prerelease=allow --upgrade mrmustard
```

apchytr (dismissed):
This is great thanks for this!

.github/workflows/upload.yml
timmysilv: I think this `uv run` should still be avoided?
```suggestion
        run: pytest tests -p no:warnings --tb=native --backend=tensorflow
```

.github/workflows/upload.yml
timmysilv: on another note, maybe instead of avoiding `uv run`, we just run `uv pip uninstall mrmustard` before installing to make extra-sure? basically I want to know that it's not just using the existing installation, because that's what happened last time.

pyproject.toml
timmysilv: nvm I see the updated docs: https://packaging.python.org/en/latest/guides/writing-pyproject-toml/#license

timmysilv: nvm i read the docs it should be cool 😎

.github/workflows/upload.yml
heltluke: I see you're about to remove `tensorflow` anyway, but yeah, sure, this can be `jax`.

heltluke:
```suggestion
        run: uv run --no-project pytest tests -p no:warnings --tb=native --backend=jax
```

.github/workflows/upload.yml
heltluke: Won't `--prerelease=allow` install pre-release versions for all dependencies? Note that I did test this both locally and at https://github.com/XanaduAI/MrMustard/actions/runs/16473307388/job/46568016618

heltluke: And sure, we could uninstall first as well.

heltluke: Better yet, I could clear the virtual environment?

heltluke: Hmm... `--no-cache-dir` is maybe the best call, actually.

apchytr (approved):

timmysilv (approved):

PR #634, opened 2025-07-24
OptimizerJax bugfix: reset opt_history 

apchytr:
**Context:** An `OptimizerJax` instance should be reusable but currently is not due to the population of `opt_history` which tricks the optimizer into thinking that it has finished optimizing on subsequent runs. This PR introduces a bugfix that allows users to reuse a single `OptimizerJax` instance 

**Description of the Change:** `OptimizerJax.minimize` now clears `opt_history`. Added `test_reuse_optimizer`. Reorganized the tests in `test_opt_lab_jax.py` in alphabetical order.

**Benefits:** Can reuse the same Jax instance which potentially allows us to avoid some recompilation.

### DIFF
--- a/tests/test_training/test_opt_lab_jax.py
+++ b/tests/test_training/test_opt_lab_jax.py
@@ -49,23 +49,89 @@ class TestOptimizerJax:
     Tests for the ``OptimizerJax`` class.
     """
 
-    @given(n=st.integers(0, 3))
-    def test_S2gate_coincidence_prob(self, n):
-        """Testing the optimal probability of obtaining |n,n> from a two mode squeezed vacuum"""
-        S = TwoModeSqueezedVacuum(
-            (0, 1),
-            r=abs(settings.rng.normal(loc=1.0, scale=0.1)),
-            r_trainable=True,
-        )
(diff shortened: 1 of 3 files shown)


tests/test_training/test_opt_lab_jax.py
```diff
@@ -403,23 +421,29 @@ def cost_fn(G):
...
+            return math.abs(circ.contract().fock_array((2, 2, 2, 2))[1, 1, 1, 1]) ** 2
 
-    def test_sgate_optimization(self):
-        """Test that Sgate is optimized correctly."""
+        opt = OptimizerJax(euclidean_lr=0.001)
+        (circ,) = opt.minimize(cost_fn, by_optimizing=[circ], max_steps=300)
+        r_var = circ.components[3].parameters.r
+        assert math.allclose(math.sinh(r_var.value) ** 2, 1, atol=1e-2)
+
+    def test_reuse_optimizer(self):
```
apchytr, line 445: This is the only new test added. Everything else is a reorder

heltluke (approved):

PR #635, opened 2025-07-24
[sc-96259] Bugfix: `State.fock_distribution` 

apchytr:
**Context:** Currently a bug exist in `State.fock_distribution` where batch dimensions and multi-mode states are not handled correctly. 

MWE:
```
state = SqueezedVacuum(0, 1.0) >> SqueezedVacuum(1, 1.0) 
print(np.sum(state.fock_distribution(32))) # Correct 
print(np.sum(state.dm().fock_distribution(32))) # Wrong
```

**Description of the Change:** Fixed `State.fock_distribution` to properly handle batches and multi-mode. Added a test. Added `math.diagonal`. Added `State.is_separable`. Cleaned up `math.matmul` and made use of `multi_dot`


mrmustard/lab/states/base.py
```diff
@@ -405,10 +405,14 @@ def fock_distribution(self, cutoff: int) -> ComplexTensor:
...
             The Fock distribution.
         """
+        batch_shape = self.ansatz.batch_shape
         fock_array = self.fock_array(cutoff)
         if not self.wires.ket or not self.wires.bra:
-            return math.reshape(math.abs(fock_array) ** 2, (-1,))
-        return math.reshape(math.abs(math.diag_part(fock_array)), (-1,))
+            return math.reshape(math.abs(fock_array) ** 2, (*batch_shape, -1))
+        axis1 = -self.n_modes - 1
+        for i in range(self.n_modes):
```
heltluke: Is the idea that complexity that was living in `math.diag` has now been moved here? I like keeping the `math` calls simple where we can (i.e. just calling existing `numpy` and `jax` functions), but I can't say that I find the changes here immediately obvious.

Did you work it out with pen and paper? With the help of the update to `test_fock_distribution`? If it's trivial, a comment here on GitHub for my education is enough, but if it's a little more involved some text in the `fock_distribution` docstring or a comment here in the code might save our future selves some effort.

apchytr: Do you mean `diag_part` as that was the method called here? To answer your question no. I would argue that the added complexity in `diag` was bad practice as it deviates away from the intended numpy implementation as users would expect. Also, numpy already provided a way to do the diagonal on `ndim>2` through the more general `diagonal` (included as part of this PR) and I would prefer people to be using that. As for `diag_part` it's essentially the same with an unnecessary `ret.flags.writeable = True` removed (and cleaned up to use the equivalent `diag`).

Not necessary to write out on pen and paper. It was just a matter of passing in the correct axes based on the wires of the mode you're grabbing the diagonal of.

heltluke (dismissed):
Might want to gate on @ziofil also reviewing, but looks good to me!

tests/test_lab/test_states/test_dm.py
```diff
@@ -469,12 +469,18 @@ def test_expectation_error(self):
...
-        x = math.broadcast_to(1, batch_shape)
-        y = math.broadcast_to(2, batch_shape)
-        state = Coherent(0, x + 1j * y)
-        assert math.allclose(state.fock_distribution(10), state.dm().fock_distribution(10))
+    def test_fock_distribution(self, n_modes, batch_shape):
+        alpha = math.broadcast_to(1 + 0.1j, batch_shape)
+        cutoff = 10
+        state = Coherent(0, alpha=alpha)
+        for i in range(n_modes - 1):
+            state >>= Coherent(i + 1, alpha=alpha)
```
ziofil: why not using `Ket.random`? if we build state out of coherent states it's a special state (separable)

apchytr: Good catch 👍 I've updated things so that we have logic for handling the separable case as well as a more general implementation. Testing wise the separable case is a little more efficient (nothing major really just a few ms). Screenshot below is with a 4 mode Coherent state no batches and a cutoff of 10.
Screenshot 2025-08-05 at 10 31 10 AM

mrmustard/lab/states/base.py
```diff
@@ -144,6 +172,27 @@ def purity(self) -> float:
...
+
+            >>> state = Ket.random([0])
+            >>> x = np.linspace(-5, 5, 100)
+
+            >>> assert np.all(state.wigner(x,0).real >= 0)
+        """
+        if isinstance(self.ansatz, PolyExpAnsatz):
+            return (self >> BtoPS(self.modes, s=0)).ansatz.PS
+        raise ValueError(
+            "Wigner ansatz not implemented for Fock states. Consider calling ``.to_bargmann()`` first.",
```
heltluke, line 193: We could also do this here on a user's behalf, right (e.g. `return self.to_bargmann() >> BtoPS(self.modes, s=0).ansatz.PS`)? Would that be taking away too much control from a user, or making their lives better/easier?

heltluke: That said, I see all you did here was move this `@property` from below.

heltluke (approved):

mrmustard/lab/states/base.py
apchytr: This is somewhat related to an offline conversation. In general, I think we should start moving away from having implicit behaviour (i.e. calling `to_bargmann` or `to_fock` under the hood) and make things more explicit (as it is here). However, I could see this method being an exception as it's a convenience method rather than core functionality

heltluke: Yeah, maybe a bit tough to draw the line as to what's a convenience method vs what's not. I agree that the less that is implicit the better.

PR #636, opened 2025-07-26
[Sc 96349] Port Missing vjps to JAXBackend

apchytr:
**Context:** Follow up to the remove tensorflow PR. In this PR we port the remaining custom gradients to the JAX backend.

TODOs added: 
- sc-95226: add vjp for fast_diagonal and make use of it in `hermite_renormalized_1leftoverMode_jax`

- sc-96431: figure out tracer type error in `test_compactFock_1leftover_gradients`

**Description of the Change:** Ported all hermite renormalized and fock utils vjps from the tensorflow backend. Brought back `test_compactFock_diagonal_gradients` for the JAX backend tests. Significantly cleaned up the hermite renormalized methods in the backend. Updated `SqueezedVacuum` and `Sgate` to make use of their respective fock lattice strategies.

**Benefits:** Can do optimizations with the other hermite renorm and fock utils methods. Lots of cleanup

PR #637, opened 2025-07-28
[sc-94954] Make JAX an optional dependency

apchytr:
**Context:** We can have a more lightweight Mr Mustard experience if we make JAX an optional dependency as it is not always used. In this PR we introduce the `jax_backend` optional dependency group.

**Description of the Change:** `jax`, `equinox`, `optax` are optional. `opt_einsum` is included as a direct dependency. Necessary changes to make backend JAX optional.

**Benefits:** Can now install Mr Mustard without having to install JAX

### DIFF
--- a/.github/CHANGELOG.md
+++ b/.github/CHANGELOG.md
@@ -10,7 +10,7 @@
 * Added a ``rich`` based repr to ``ParameterSet``.
 [(#616)](https://github.com/XanaduAI/MrMustard/pull/616)
 
-* Made `JAX` and optional dependency.
+* Made `JAX` an optional dependency.
 [(#637)](https://github.com/XanaduAI/MrMustard/pull/637)
 
 ### Bug fixes
--- a/.github/CHANGELOG.md
+++ b/.github/CHANGELOG.md
@@ -10,6 +10,9 @@
 * Added a ``rich`` based repr to ``ParameterSet``.
 [(#616)](https://github.com/XanaduAI/MrMustard/pull/616)
 
+* Made `JAX` and optional dependency.
+[(#637)](https://github.com/XanaduAI/MrMustard/pull/637)
+
 ### Bug fixes
 
 * Fixed a bug in `OptimizerJax` that would make the optimizer think it reached a stable optimum upon repeat calls to `minimize`.
--- a/.github/CHANGELOG.md
+++ b/.github/CHANGELOG.md
@@ -10,7 +10,7 @@
 * Added a ``rich`` based repr to ``ParameterSet``.
 [(#616)](https://github.com/XanaduAI/MrMustard/pull/616)
 
-* Made `JAX` an optional dependency.
+* Made `JAX` an optional `jax_backend` dependency.
 [(#637)](https://github.com/XanaduAI/MrMustard/pull/637)
 
 ### Bug fixes
--- a/.github/workflows/tests_numpy.yml
+++ b/.github/workflows/tests_numpy.yml
@@ -56,7 +56,7 @@ jobs:
           NUMBA_CPU_FEATURES: ""
         run: |
           set -o pipefail
-          uv run --all-extras pytest tests --backend=numpy -p no:warnings --tb=native ${{ env.UPLOAD_TIMINGS == 'true' && '--durations=0 -vv ' || '' }}| tee durations.txt
+          uv run --no-group jax_backend pytest tests --backend=numpy -p no:warnings --tb=native ${{ env.UPLOAD_TIMINGS == 'true' && '--durations=0 -vv ' || '' }}| tee durations.txt
           set +o pipefail
 
       - name: configure aws credentials
--- a/mrmustard/math/backend_numpy.py
+++ b/mrmustard/math/backend_numpy.py
@@ -88,6 +88,7 @@ class BackendNumpy(BackendBase):
         return np.array(array, ndmin=n, dtype=dtype)
 
     def BackendError(self):
+        # no numpy backend specific errors
         raise NotImplementedError
 
     def broadcast_to(self, array: np.ndarray, shape: tuple[int]) -> np.ndarray:
--- a/mrmustard/math/backend_numpy.py
+++ b/mrmustard/math/backend_numpy.py
@@ -87,6 +87,9 @@ class BackendNumpy(BackendBase):
     def atleast_nd(self, array: np.ndarray, n: int, dtype=None) -> np.ndarray:
         return np.array(array, ndmin=n, dtype=dtype)
 
+    def BackendError(self):
+        raise NotImplementedError
+
     def broadcast_to(self, array: np.ndarray, shape: tuple[int]) -> np.ndarray:
         return np.broadcast_to(array, shape)
 
--- a/mrmustard/physics/fock_utils.py
+++ b/mrmustard/physics/fock_utils.py
@@ -29,9 +29,9 @@ from mrmustard.math.caching import tensor_int_cache
 from mrmustard.utils.typing import Batch, Scalar, Tensor, Vector
 
 try:
-    from equinox import EquinoxRuntimeError
+    import jax
 except ImportError:
-    EquinoxRuntimeError = None
+    jax = None
 
 # ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 # ~~~~~~~~~~~~~~ static functions ~~~~~~~~~~~~~~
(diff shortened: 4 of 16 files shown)


.github/CHANGELOG.md
```diff
@@ -10,6 +10,9 @@
 * Added a ``rich`` based repr to ``ParameterSet``.
 [(#616)](https://github.com/XanaduAI/MrMustard/pull/616)
 
+* Made `JAX` and optional dependency.
```
apchytr, line 13:
```suggestion
* Made `JAX` an optional dependency.
```

pyproject.toml
```diff
@@ -49,6 +48,12 @@ dependencies = [
     "thewalrus >=0.19.0,<1",
 ]
 
+[project.optional-dependencies]
+jax = [
+    "equinox>=0.11.10",
+    "optax>=0.2.4,<0.3",
+]
```
timmysilv: `jax` itself should be part of this group. I know it's a downstream dependency but we should be explicit in this file.

Also, to not cause any burden on MM devs, we should make a `dependency-group` called "jax" that includes these deps (unpinned) and add it to the default groups

tests/test_math/test_backend_manager.py
```diff
@@ -42,18 +42,22 @@ class TestBackendManager:
...
         assert math.BackendError is TracerArrayConversionError
 
     def test_get_backend(self):
         r"""
         Tests the ``get_backend`` method.
         """
         assert math.get_backend("numpy").name == "numpy"
-        assert math.get_backend("jax").name == "jax"
+        with contextlib.suppress(ModuleNotFoundError):
+            assert math.get_backend("jax").name == "jax"
```
timmysilv: so even if you don't have jax installed, this assertion works? or `contextlib.supress` just moves on if the provided exception occurs?

tests/test_math/test_jitting.py
timmysilv, line 1: assuming this whole file requires jax, you can just replace this:
```python
import jax	
import jax.numpy as jnp	
import pytest
```
with this:
```python
import pytest
jax = pytest.importorskip("jax")
jnp = pytest.importorskip("jax.numpy")
```
Then, if jax isn't installed, the whole file is just skipped

tests/test_math/test_backend_manager.py
apchytr: https://docs.astral.sh/ruff/rules/suppressible-exception/

Just moves on! It's equivalent to a try except but I guess recommended as a ruff rule. There is a performance hit but I think that's fine for these tests

pyproject.toml
apchytr: Updated and added to the `dev` group. Or do we prefer to have a separate one?

timmysilv: I think `dev` is fine because our devs will want it installed : )
I would also put `equinox` and `optax` in dev though

apchytr: Gah was hoping `uv` would be clever enough to pick it up in a way that I didn't need to respecify 😅

mrmustard/training/optimizer.py
```diff
@@ -20,197 +20,206 @@
...
-    def make_step(
-        self,
-        optim: GradientTransformation,
-        cost_fn: Callable,
-        by_optimizing: Sequence[Variable | CircuitComponent | Circuit],
-        opt_state: OptState,
-    ) -> tuple[Sequence[Variable | CircuitComponent | Circuit], OptState, float]:
-        r"""
-        Make a step of the optimization.
+try:
```
apchytr, line 29: @timmysilv is there a better way of handling this besides wrapping it in a try except? For context, I want people to be able to import the `Optimizer` even if they don't have JAX installed. It would just be a dummy one that tells the user to install jax on init

.github/workflows/tests_numpy.yml
```diff
@@ -56,7 +56,7 @@ jobs:
           NUMBA_CPU_FEATURES: ""
         run: |
           set -o pipefail
-          uv run --all-extras pytest tests --backend=numpy -p no:warnings --tb=native ${{ env.UPLOAD_TIMINGS == 'true' && '--durations=0 -vv ' || '' }}| tee durations.txt
+          uv run --no-group jax_backend pytest tests --backend=numpy -p no:warnings --tb=native ${{ env.UPLOAD_TIMINGS == 'true' && '--durations=0 -vv ' || '' }}| tee durations.txt
```
apchytr, line 59: This is so that we're sure if you do not install JAX tests will still pass

mrmustard/training/optimizer.py
timmysilv: afaik Python doesn't love lazy imports. There was a PEP for it and it was rejected.

As an example, [PennyLane made a hack](https://github.com/PennyLaneAI/pennylane/blob/master/pennylane/data/base/_lazy_modules.py) to support it, but as you can see it's a bit involved.

If I were you, I'd make optimizer fail to import without jax and call it a day, or move the real optimizer into another file, and only import that file if jax succeeds its import.

apchytr: updated to raise on import

mrmustard/math/backend_numpy.py
```diff
@@ -87,6 +87,9 @@ def astensor(self, array: np.ndarray, dtype=None) -> np.ndarray:
     def atleast_nd(self, array: np.ndarray, n: int, dtype=None) -> np.ndarray:
         return np.array(array, ndmin=n, dtype=dtype)
 
+    def BackendError(self):
+        raise NotImplementedError
```
heltluke, line 92: Worth adding a comment, or even to the error itself, that we should never end up here---it's just so the backend manager has something to `_apply`?

mrmustard/physics/fock_utils.py
```diff
@@ -21,14 +21,18 @@
...
 from scipy.special import comb, factorial
 
 from mrmustard import math, settings
 from mrmustard.math.caching import tensor_int_cache
 from mrmustard.utils.typing import Batch, Scalar, Tensor, Vector
 
+try:
+    import jax
+except ImportError:  # pragma: no cover
+    jax = None
```
heltluke, line 34: Do we fail immediately if someone tries to set the backend to "jax"? If not, I think we might also want to add a message in `fock_state` following `if math.backend_name == "jax":`.

tests/test_math/test_backend_manager.py
```diff
@@ -53,7 +61,8 @@ def test_get_backend(self):
         Tests the ``get_backend`` method.
         """
         assert math.get_backend("numpy").name == "numpy"
-        assert math.get_backend("jax").name == "jax"
+        with contextlib.suppress(ImportError):
+            assert math.get_backend("jax").name == "jax"
```
heltluke: Why is there an `ImportError` here?

tests/test_math/test_backend_manager.py
```diff
@@ -63,7 +72,8 @@ def test_einsum(self):
         res = math.astensor([[7, 10], [15, 22]])
 
         assert math.allclose(math.einsum("ij,jk->ik", ar, ar), res)
-        assert math.allclose(math.einsum("ij,jk->ik", ar, ar, backend="jax"), res)
+        with contextlib.suppress(ModuleNotFoundError):
+            assert math.allclose(math.einsum("ij,jk->ik", ar, ar, backend="jax"), res)
```
heltluke: Why is there a `ModuleNotFoundError` here? Should this also be split and/or decorated with `@pytest.mark.requires_backend("jax")`?

tests/test_math/test_compactFock.py
```diff
@@ -9,7 +9,11 @@
 from mrmustard.lab import DM, Ggate, SqueezedVacuum, Vacuum
 from mrmustard.lab.transformations.attenuator import Attenuator
 from mrmustard.physics import gaussian
-from mrmustard.training import Optimizer
+
+try:
+    from mrmustard.training import Optimizer
+except ImportError:
+    Optimizer = None
```
heltluke, line 16: I might have thought that `test_compactFock_1leftover_gradients` should have the `@pytest.mark.requires_backend("jax")` decorator, just like `test_compactFock_diagonal_gradients`. But it can't yet? What ends up getting tested?

tests/test_math/test_jitting.py
```diff
@@ -16,132 +16,136 @@
...
-        >> BS_12
-        >> att
-        >> att.on(1)
-        >> att.on(2)
-    )
-    output_fock_state = state_out.fock_array(shape=(20, 5, 5, 20, 5, 5))
-    marginal = output_fock_state[:, 4, 4, :, 4, 4]
-    return math.real(math.trace(marginal))
+jax = pytest.importorskip("jax")
+jnp = pytest.importorskip("jax.numpy")
```
heltluke, line 25: TIL!

tests/test_math/test_backend_manager.py
heltluke: Should maybe split into `test_get_backend_numpy` and `test_get_backend_jax`, decorating the second with `@pytest.mark.requires_backend("jax")`?

mrmustard/physics/fock_utils.py
apchytr: Yeah we get a nice error message that tells users to install the `jax_backend` group

tests/test_math/test_compactFock.py
apchytr: If nothing is specified it always gets skipped. Updated [here](https://github.com/XanaduAI/MrMustard/pull/636/files#diff-e768fecde2483177fd28c19adbf8366f829fc52e592281f14bdab77400291d73R96) but basically it's a test that has several issues. In this PR the vjp isn't defined but in the linked PR there are tracer errors that need to be investigated as part of another story.

tests/test_math/test_backend_manager.py
apchytr: `ModuleNotFoundError` is because in that case opt einsum is the one throwing an error. But updated to split the tests out

mrmustard/math/backend_numpy.py
apchytr: Yeah mostly there for the sake of symmetry between the backends

heltluke (approved):

